PR #742, opened 2015-01-06
Controls

cbrightsci:
This is an edit of the CmakeLists.txt to access the branch I made on the renderer side. So, it will require a new build on the machines.


dcwhite:
@cbrightsci I tried it out and noticed a difference, but it will need more testing during the lunch tomorrow, with more large meshes.

PR #760, opened 2015-01-14
SetuptDCS update

moritzdannhauer:
GUI-TODO:
1) refnode is always set to zero -> Dan found the problem in the module *.cc code
2) table header does not appear


dcwhite:
@moritzdannhauer I'll take a look at this today and try to find the GUI problem.

dcwhite:
path

@moritzdannhauer Fix this path in the network file so others can run it.

dcwhite:
@moritzdannhauer See commit 606af3c12122f9ee8533aebc7e7dbe9cb6d3ee59.
- [ ] Don't call clear() on the tableWidget, it removes the headers. Bad Qt function there.
- [ ] The real problem was in the .ui file: "horizontalHeaderVisible" was not checked. In addition to that I cleaned up the layouts. So just merge in that commit to your PR branch, and I will accept it after you fix the network file.

aylakhan:
TMS_TOY_EXAMPLE.srn5 will also need to be updated in SCIInstitute/BrainStimulator, since we're moving away from including toolkit networks in this repo. All file paths must be relative to %SCIRUNDATADIR%.

PR #834, opened 2015-01-31
Object selection

cbrightsci:
Just for you @dcwhite :)


dcwhite:
Excited to test this!

aylakhan:
Looks good in tikhonov-inverse network!

src/Interface/Modules/Render/ViewScene.h
```diff
@@ -90,6 +92,7 @@ namespace SCIRun {
       virtual void hideEvent(QHideEvent* evt) override;
       virtual void contextMenuEvent(QContextMenuEvent* evt) override {}
     private:
+      bool isObjectUnselected(std::string name);
```
dcwhite: Change to const std::string&

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -47,7 +47,7 @@ using namespace SCIRun::Render;
 ViewSceneDialog::ViewSceneDialog(const std::string& name, ModuleStateHandle state,
 	QWidget* parent /* = 0 */)
 	: ModuleDialogGeneric(state, parent), mConfigurationDock(0), shown_(false),
-	itemManager_(new ViewSceneItemManager)
+	itemManager_(new ViewSceneItemManager(state, this))
```
dcwhite: Passing this in initialization list of constructor is risky. I'd pull that out and use a setter inside the constructor body

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -390,6 +399,33 @@ void ViewSceneDialog::assignBackgroundColor()
...
+void ViewSceneDialog::handleSelectedItem(const QString& name)
+{
+  unselectedObjectNames_.erase(std::remove(unselectedObjectNames_.begin(), unselectedObjectNames_.end(), name.toStdString()), unselectedObjectNames_.end());
+  newGeometryValue();
+}
+
+//------------------------------------------------------------------------------
+bool ViewSceneDialog::isObjectUnselected(std::string name)
+{
+  for (auto it = unselectedObjectNames_.begin(); it != unselectedObjectNames_.end(); ++it)
```
dcwhite: This manual loop can be replaced with an std::find call
```
return std::find(unselectedObjectNames_.begin(); unselectedObjectNames_.end(), name) != unselectedObjectNames_.end();
```

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -390,6 +399,33 @@ void ViewSceneDialog::assignBackgroundColor()
 }
 
 //------------------------------------------------------------------------------
+void ViewSceneDialog::handleUnselectedItem(const QString& name)
+{
+  unselectedObjectNames_.push_back(name.toStdString());
+  newGeometryValue();
```
dcwhite: This may be slow when user clicks a checkbox many times in a row. I'll test that out.

dcwhite:
Works OK in networks with a single ViewScene. Having problem with multiple ViewScenes, report soon.

dcwhite:
False alarm transparency was set to 0 :P

src/Interface/Modules/Render/ViewScene.cc
dcwhite: Heh, the combo box we're using doesn't allow a ton of clicks in a row, preventing this goofy behavior. We can examine the performance later

PR #850, opened 2015-02-05
Brig work

brigb123:
Updates to bounding box and show color map


src/Modules/Fields/EditMeshBoundingBox.cc
```diff
@@ -267,7 +267,7 @@ bool EditMeshBoundingBox::isBoxEmpty() const
 Core::Datatypes::GeometryHandle EditMeshBoundingBox::buildGeometryObject() {
 
     Core::Datatypes::GeometryHandle geom(new Core::Datatypes::GeometryObject(NULL));
-
+    geom->objectName = "Bounding Box";
```
dcwhite: @brigb123 Need a "more unique" name. It should be different for every module instance/reexecution. I used the stringized pointer value of the geom object as a cheap way to do this.

dcwhite:
```
std::ostringstream ostr;
ostr << "Bounding Box " << geom.get();
geom->objectName = ostr.str();
```

PR #852, opened 2015-02-05
Convert matrix to string module

mhansen1:
Everything seems to be working ok. Added in a test network that uses the PrintDatatype module. 

Also, added a converters namespace that has the ConvertMatrixToString module/algorithm files. If this isn't wanted/needed I can easily take it out.


src/Modules/Factory/ModuleFactoryImpl1.cc
```diff
@@ -193,6 +195,7 @@ void ModuleDescriptionLookup::addEssentialModules()
   addModuleDesc<ConvertIndicesToFieldData>("Real ported module", "..."); 
   addModuleDesc<SolveInverseProblemWithTikhonov>("...", "...");
   addModuleDesc<ShowColorMapModule>("ShowColorMap", "Visualization", "SCIRun", "Real ported module", "..."); 
+  addModuleDesc<ConvertMatrixToString>("ConvertMatrixToString","Converters","SCIRun","...","...");
```
dcwhite: @mhansen1 This is an acceptable factory diff. Although you'll need to merge again, with your own change from the other PR, and with master, so be careful.

MatrixToString_test_network/Converter_test.srn5
```diff
@@ -0,0 +1,341 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
```
dcwhite: @mhansen1 Please move this test network to SCIRun/src/ExampleNets/regression

src/Core/Algorithms/Factory/HardCodedAlgorithmFactory.cc
```diff
@@ -95,6 +96,7 @@ using namespace SCIRun::Core::Algorithms::DataIO;
 using namespace SCIRun::Core::Algorithms::Math;
 using namespace SCIRun::Core::Algorithms::BrainStimulator;
 using namespace SCIRun::Core::Algorithms::Math;
+using namespace SCIRun::Core::Algorithms::Converters;
```
dcwhite: @mhansen1 You're missing the line to actually connect the module and algorithm. Running your test network results in an error leading to this file: 

ERROR: Null algorithm object, make sure AlgorithmFactory knows about this module's algorithm types.

dcwhite:
Maybe you're missing a commit, assuming your network works...

mhansen1:
Ya, I'm still missing a commit. Trying to get the Laplacian matrix module
to get a good merge so that I can merge the two (making sure that it builds
fine still). You wanted me to merge the Laplacian matrix module with the
ConvertMatrixToString, right?

dcwhite:
Let's keep them separate; if you get this one good, I will merge to master, then you can merge master into the Laplacian branch.

Issue #574, opened 2014-07-31
Linking SCIRun executable with WinMain entry point fails


aylakhan:
Applies to winInstaller branch. Linking SCIRun executable with WinMain entry point (created by adding WIN32 parameter to CMake command ADD_EXECUTABLE) fails with error LNK2019: unresolved external symbol WinMain referenced in function _tmainCRTStartup.

dcwhite:
#856 broke command line arguments being forwarded. Told @mhansen1 to build RUN_TESTS to verify fix.

PR #856, opened 2015-02-09
Windows entry point (WinMain)

mhansen1:
Changed the program entry point so that if the target system is Windows a different entry point is used and made it so that the console will still be available if needed.


dcwhite:
@mhansen1 Sending this one back. All command-line arguments were broken with this change. As an initial test of command-line argument forwarding, build the RUN_TESTS project in Visual Studio--it runs regression tests that rely on command line arguments working.

dcwhite:
Linked issue is #574

dcwhite:
Any updates on this one?

mhansen1:
Should be just about done. Got it working with the RUN_TESTS project, but
haven't run it for all the tests yet. Assuming that it goes fine through
all of the tests I'll commit the changes that I have today/tomorrow morning.

Also, if a test fails, is that because one of the modules isn't working
correctly, or should they all pass?

dcwhite:
There are some failing tests currently. A few unit tests have random errors (rare but annoying), and 2 of the set of regression networks are broken currently. You should get an output from RUN_TESTS similar to this:
```
 98% tests passed, 3 tests failed out of 129
1>  
1>  Total Test time (real) = 212.77 sec
1>  
1>  The following tests FAILED:
1>  	 79 - .Test.Samples.ElectrodeCoilSetup_srn (Failed)
1>  	111 - .Test.Samples.showColorMap_srn (Failed)
1>  	128 - .Test.Samples.writeMatlabMatrix_srn (Failed)
```
79 and 128 are the broken ones; 111 is one of the random failures. The rest should pass.

src/Main/CMakeLists.txt
```diff
@@ -51,9 +51,10 @@ IF(APPLE)
 
   ADD_EXECUTABLE(${APPLICATION_NAME} MACOSX_BUNDLE ${SCIRunMain_SOURCES})
 
-ELSEIF(WIN32)
+ELSEIF(WIN32)  
   IF(SCIRUN_SHOW_CONSOLE)
-    ADD_EXECUTABLE(${APPLICATION_NAME} ${SCIRunMain_SOURCES})
+    ADD_DEFINITIONS(-DSCIRUN_SHOW_CONSOLE=1)
+    ADD_EXECUTABLE(${APPLICATION_NAME} WIN32 ${SCIRunMain_SOURCES})
```
aylakhan: Line 57 and 59 are the same.

mhansen1:
Seems to be fine now. It's running through all of the tests, any fails a few of them. The only possible issue is at the end there is an error. I've attached a screen shot of the error. It seems to be related to administrative access, and still shows up outside of this branch. Also seems to be related to CMake. Is this a part that I should look into, or not worry about it?
run_tests_error

dcwhite:
What does the VS output window report, as opposed to the error list? You could also try running the tests in release mode, it will be faster.

mhansen1:
The output window shows pretty much the same, but I have been looking around a bit and it seems that this is standard. There is a setting (or line of code) in the CMake files for the test that sets/unsets this. If the build has failed tests or build errors, then it fails and returns a non-zero code. I can keep looking into this and see if that's right or not in this case.
Also, here's a screen shot of the VS output window, but it says the same thing as the error window.
run_tests_error2

aylakhan:
If you run ctest in the command shell, you'll get a list of failing tests and it would report to the dashboard with more information, which would be helpful. Try running: C:\Program Files (x86)\CMake 2.8\bin\ctest.exe -A CMakeCache.txt -D Experimental in the SCIRun bin directory.

dcwhite:
Look right above the MSB error. That is where the list of failing tests appears. Ignore the MSB error.

aylakhan:
Results at http://my.cdash.org/viewTest.php?onlyfailed&buildid=736960.

dcwhite:
Looks like a git fetch is needed on the test data repo, the newest tests are failing. But otherwise normal.

mhansen1:
Ya, just needed the test data update. Now the only tests that are failing are the ones that you mentioned. So if I don't need to worry about the MSB errors, is there anything else that I should be worried about with this PR, or does it all look good?

dcwhite:
I'll test it again within the next few days and let you know here.

PR #877, opened 2015-02-13
functional spheres and cylinders.

brigb123:
The renderer doesn't update without deleting/re-adding createstandardcolormap and/or viewscene. This includes the shader needed. Code should be refactored for compactness.


dcwhite:
Testing on Windows. Multiple problems:
* Switching to cylinder edges didn't work. Got this message:
```
ren::RenderSimpleGeom: Unable to satisfy shader! Not enough attributes.
```
* In the 2x2x2 LatVol case, edges as lines disappear completely. 
* In 3x3x3 case, I turned edges off and then nodes on, and the edges reappeared.
* Changing the nodes from points to spheres and back toggles the edge transparency, so I was not even able to test spheres.

dcwhite:
After deleting/readding ViewScene, nodes button does toggle the nodes, but I get the same shader error as cylinders.

dcwhite:
@cbrightsci May want to check this one for merge issues--with your RenderState enum transparency fix. See above regressions.

brigb123:
@dcwhite  i just tested every latvol possibility with edges and cylinders. i was able to get each situation to render properly by deleting and re-adding the view scene

dcwhite:
OK I will test on Mac to verify that, but I can't merge until Windows works too. Merging this should wait until you don't have to delete/re-add ViewScene--since that is stupid.

brigb123:
That is fine. I'll test again when that issue is resolved. I don't know what specific code is needed to push those updates down the render pipe. If that is outlined properly, I can add that logic. Not sure why Windows wouldn't follow suit with Mac on this, but it will probably resolve with the view scene issue.

dcwhite:
Merging this is waiting on #891 by me. Probably next week.

PR #899, opened 2015-03-05
Fonts are now using textured, anti-aliased, better sizes.

brigb123:


brigb123:
@dcwhite I still can't build on windows, might need your help for that, so IDK if this fix will work on windows. Let me know about that. I'm using triangles and textures (easier than i thought), which appears to work with the scale bar, so should work for the font.

src/Modules/Visualization/ShowField.cc
```diff
@@ -100,6 +100,21 @@ void ShowFieldModule::execute()
 {
   boost::shared_ptr<SCIRun::Field> field = getRequiredInput(Field);
   boost::optional<boost::shared_ptr<SCIRun::Core::Datatypes::ColorMap>> colorMap = getOptionalInput(ColorMapObject);
+  
+  //set the default colormap scaling to be full range if no rescaling was provided.
+  if (colorMap &&
```
dcwhite: @aylakhan @brigb123 @cbrightsci I'm not sure this code should be in ShowField. It kind of defeats the purpose of the RescaleColorMap module, and just moves the old ShowField color scaling to a different place. Plus it will introduce a bug with tensor field visualization.

aylakhan:
No scaling in ShowField! All colormap scaling code must be removed.

src/Modules/Visualization/ShowField.cc
brigb123: I was thinking the same thing. But it does confine the min/max to the colormap. field values are not altered in showfield. I'm not sure how else to get the field min/maxes down to the shader, where it is ultimately needed.

src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -1205,7 +1214,52 @@ namespace SCIRun {
...
+                    std::cerr << "Cannot find font \"Assets/times_new_roman.font\"" << std::endl;
+                    success = false;
+                    in.close();
+                }
+            }
+            if (success) {
+                size_t w,h;
+                in >> w >> h;
+                uint16_t *font_data = new uint16_t[w*h];
+                in.read((char*)font_data,sizeof(uint16_t)*w*h);
```
aylakhan: C style cast

dcwhite:
@brigb123 Closing this one too since it has rescale merged into it. Make a new PR with only the font changes.

PR #903, opened 2015-03-05
Fonts are rendered as textures on triangles.

brigb123:
Fonts are anti-aliased, sizes are reasonable for viewing, and rendered as textures on triangles instead of as a set of points.


dcwhite:
Thanks

src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -1099,13 +1108,13 @@ namespace SCIRun {
...
 		// Create default colormaps.
-		void SRInterface::generateColormaps()
+		void SRInterface::generateTextures()
 		{
 			size_t resolution = 1000;
             float step = 1.f / static_cast<float>(resolution);
             ColorMap cm("Rainbow");
 			std::vector<uint8_t> rainbow;
-			rainbow.reserve(resolution * 3);
+			rainbow.reserve(resolution * 4);
```
aylakhan: Magic numbers - can the 4s be changed to a const with a descriptive name?

brigb123: This part of code will actually leave with the rescale refactor PR #906 .

dcwhite:
@brigb123 Discussed this one today. We'd like to merge the PR and just disable font numbers for now--we can open an issue to continue the work when there's time. Otherwise we get branch divergence. Can you add the necessary commits to disable fonts? You don't need to hide any widgets; just disable them.

PR #923, opened 2015-03-25
update on setuptdcs

moritzdannhauer:
Sorry for 2 pull requests in a row on the same module.
I just added 2 more lines of code to make it more stable.


src/Core/Algorithms/BrainStimulator/SetupRHSforTDCSandTMSAlgorithm.cc
```diff
@@ -3,7 +3,7 @@
 
    The MIT License
 
-   Copyright (c) 2015 Scientific Computing and Imaging Institute,
+   Copyright (c) 2012 Scientific Computing and Imaging Institute,
```
dcwhite: Bad merge, hehe...

moritzdannhauer:
how about now?

PR #939, opened 2015-04-20
Added the RescaleColorMap modules to FwdInvToolkit example networks.

mhansen1:
Added in the RescaleColorMap module to the FwdInvToolkit examples.


dcwhite:
Looks great! Thanks

PR #949, opened 2015-05-04
RegisterWithCorrespondences Module

mhansen1:
Should be good to merge (updated the branch right before the pull request). The module is building fine and has the UI and everything. The only part that may need to be tweaked is the algorithm. I don't know if something I did or did not do, but I can't find a .fld (or .nrrd) file that is compatible with the correspondence ports. Also should be building without any errors (only reason it would is if something got changed when fixing merge conflicts).


src/Modules/Legacy/Fields/RegisterWithCorrespondences.cc
```diff
@@ -28,85 +28,42 @@
...
-  get_input_handle("Correspondences2",cors2,true);
+	auto input1 = getRequiredInput(InputField);
+	auto input2 = getRequiredInput(Correspondences1);
+	auto input3 = getRequiredInput(Correspondences2);
   
-  if (inputs_changed_ || !oport_cached("Output") || method_.changed() ||
-      edmethod_.changed())
-  {
-    update_state(Executing);
+    auto output = algo().run_generic(withInputData((InputField,input1),(Correspondences1,input2),(Correspondences2,input3)));
```
dcwhite: @mhansen1 Small mistake using this macro, that fails compilation on clang (see Travis error link): to fix, remove the commas between the arguments to withInputData.

dcwhite: Windows builds, so it must be silently removing the commas.

dcwhite: Actually, Windows will just ignore what's after the commas. So the module was not passing any field object for Correspondences1 or 2 to the algo, which explains why you couldn't get a field to work there.

src/Core/Algorithms/Legacy/Fields/RegisterWithCorrespondences.cc
```diff
@@ -662,25 +741,22 @@ RegisterWithCorrespondencesAlgo::runN(FieldHandle input, FieldHandle Cors1, Fiel
...
     
     Cors->size(num_cors);
     points->size(num_pts);
     
     double xcomp=0.0,ycomp=0.0,zcomp=0.0,mag=0.0;
-    //MatrixHandle SMat;
-    Smat=new DenseMatrix(num_pts,num_cors);
-    DenseMatrix *Sigma = dynamic_cast<DenseMatrix*>(Smat->dense());
+    //Smat = matrix_convert::to_dense(Smat);
+    DenseMatrixHandle Sigma = matrix_convert::to_dense(Smat);
```
dcwhite: @mhansen1 Conversion error: original code allocated a dense matrix here, please add it back. Code crashes with null pointer.
```
//original
Smat=new DenseMatrix(num_pts,num_cors);
DenseMatrix *Sigma = dynamic_cast<DenseMatrix*>(Smat->dense());
```

dcwhite:
@mhansen1 Use this network to test the module:
https://github.com/SCIInstitute/SCIRun/blob/master/src/ExampleNets/IBBM2015/Registration_Network.srn5

dcwhite:
After fixing the matrix allocation above, the module runs but produces an invalid field. I can investigate that after merging. If you fix the two easy issues I will merge it.
- [ ] Remove commas in withInputData call
- [ ] Allocate dense matrix inside radial_basis_func:
```
Smat.reset(new DenseMatrix(num_pts, num_cors));
auto Sigma = Smat;
```

mhansen1:
Forgot to add in the code in your last comment for the Sigma matrix. Has something else in right now, but the test network is still running and not crashing SCIRun. For some reason when I hit build with the new code it said that the CMakeLists file was out of date, so I have to wait for it to rebuild before I can make sure that the code piece that you gave me works with the algorithm. Once I get home from work tonight I can fix it and do another commit.

dcwhite:
OK

mhansen1:
Should be good to go now.

aylakhan:
Thank you Max!

PR #1071, opened 2015-06-30
Extend electrode coil setup

moritzdannhauer:


moritzdannhauer:
@dcwhite : its ready for review and merging. i had some problems with the algo test routines since
i need to fill the table of the GUI with values. maybe you have an idea how to do that. this module is needed for IBBM and the testing lunch.

PR #1094, opened 2015-07-08
Viewscene adjustments

cbrightsci:


dcwhite:
@cbrightsci An old glitch reemerged: when you add a ViewScene (on Mac atleast), the config window pops up very quickly before being hidden. It should start hidden.

dcwhite:
@cbrightsci I connected one CLV->SF to a ViewScene. Opened new object picker, deselected the checkbox for ShowField:0, and then two unchecked boxes appeared:
screenshot 2015-07-08 15 48 46

dcwhite:
Looks like that just happens with one connected. With two SFs, the checkboxes behave correctly.

cbrightsci:
@dcwhite It should be good now.

dcwhite:
Second issue is fixed. Still seeing the first issue.

Suggestions:
- [ ] Move "Objects" tab to first position in Configure screen so it show up first. 
- [ ] Then add hotkey to ViewScene to bring up the list. It's heavily used so we need quick access.

dcwhite:
Cool I will test early tomorrow.

PR #1213, opened 2015-10-07
new ported module: PrintStringIntoString

jessdtate:
No UI or output yet, but should build.


dcwhite:
@jessdtate No problems given what we discussed, but I'll wait on merging this for two reasons:
* Wait until Travis build problem is fixed, so this can be properly auto-checked;
* I'm going to look at the dynamic port-related code, to see if that is why it's not outputting anything. 
When you tackle the UI, you can just add on commits to this open PR.

jessdtate:
this module should be ready to go.  It doesn't have an algorithm, but we are going to get to that later.

src/Modules/Legacy/String/PrintStringIntoString.h
```diff
@@ -0,0 +1,62 @@
...
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ DEALINGS IN THE SOFTWARE.
+ */
+
+#ifndef MODULES_STRING_PRINTSTRINGINTOSTRING_H
+#define MODULES_STRING_PRINTSTRINGINTOSTRING_H
+
+#include <Dataflow/Network/Module.h>
+#include <Modules/Math/share.h>
```
dcwhite: @jessdtate Windows build problem (that I mentioned in the lunch): this needs to be the new share.h, i.e. Module/Legacy/String/share.h

dcwhite:
@jessdtate Build is good and a quick check showed it's working. Could you include a simple test network in src/ExampleNets/regression? Once that's in I will merge it.

PR #1225, opened 2015-10-26
Distance field

jessdtate:
better test for edges and precision.  Fixed bug in closest_point_on_tri in CompGeom.cc because it was choosing the wrong edge.  This fixes it in more cases.  There may still be cases where it will fail.


dcwhite:
@jessdtate How hard would it be to write a unit test for this change?

jessdtate:
You could compare it to a known solution, or compare it to some of the other algorithms.  It is a bit tricky to test for the cases it may have problems, though I guess you could do a neighborhood check and see if the value is vastly different than any of the neighbors.  I'm not completely sure of the structure and abilities of the unit testing side of things.

dcwhite:
Let me see if I can start a unit test. I'll merge this PR then make a new issue for that.

PR #1232, opened 2015-11-03
Splitfilename module conversion

jessdtate:
Converted module SplitFileName.  this module is done and is working as intended.


src/Core/Application/Version.cc
```diff
@@ -0,0 +1,32 @@
...
+   THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+   DEALINGS IN THE SOFTWARE.
+*/
+
+#include <Core/Application/Version.h>
+
+const std::string VersionInfo::GIT_VERSION_TAG = "v5.0-alpha.X.2lunch-92-g43b8731";
+const std::string VersionInfo::GIT_COMMIT_SHA = "43b873137697ae6b441c4a028940b787784cd448";
```
dcwhite: @jessdtate This file should not be checked in--it's auto-generated. I deleted it in master. Can you delete it from your PR branch? It should be ignored--not sure why that's not happening. For now just remember not to stage it, since it will change every commit.

jessdtate:
i removed it from my branch.  That file was annoying but I didn't know what to do with it.

PR #1245, opened 2015-11-24
Jcf tikhonov refactor

jcollfont:
Here are the changes to the Tikhonov implementation.

It does not compile, the errors seem to be when calling the ,method .inverse().

This method call appears 5 times in lines:  268, 281, 307, 320 and 668


src/Modules/Legacy/Inverse/SolveInverseProblemWithTikhonovImpl.cc
```diff
@@ -188,103 +188,275 @@ namespace LinearAlgebra
...
+        {
+            
+            // check that the matrix is of appropriate size (equal number of rows as columns in fwd matrix)
+            if ( N != sourceWeighting_->nrows() )
+            {
+                BOOST_THROW_EXCEPTION(DimensionMismatch() << DimensionMismatchInfo("Solution Regularization Matrix must have the same number of rows as columns in the Forward Matrix !"));
+            }
+          
+            RRtr = *sourceWeighting_ * sourceWeighting_->transpose();
+            iRRtr = RRtr.inverse();   // todo: @JCOLLFONT need to compute inverse of RRtr and set try/catch
```
dcwhite: @jcollfont The fix for this and similar compile errors is to force evaluation. Eigen likes to optimize assignments and in this case doesn't want to compute the inverse unless you tell it explicitly:
```
iRRtr = RRtr.inverse().eval();
```
That's it. Same for other inverse calls.

Issue #956, opened 2015-05-07
SolveInverseProblemWithTikhonov regularization matrices seem not to do anything


jcollfont:
While testing the SolveInverseProblemWithTikhonov, its output seems not to change when I am inputing a regularization matrix in either the solution or measurement constraing input

dcwhite:
:+1:

dcwhite:
@jcollfont I believe those ports are not hooked up yet at all. Some test networks would be helpful--the toolkit network does not have any inputs attached.

dcwhite:
The ports are indeed hooked up and the values are passed into the algorithm code. See lines 230-280 of https://github.com/SCIInstitute/SCIRun/blob/master/src/Modules/Legacy/Inverse/SolveInverseProblemWithTikhonovImpl.cc. If you send me some test networks with data that show the expected results when these ports are used (no existing networks use them), I can have something to debug.

dcwhite:
Any updates on this issue? Do the latest test cases exercise the additional ports? If so, please add comments to that PR referencing this issue.

jcollfont:
The module is currently accepting regularization matrices for both source and measurement.

In both cases the matrices are being checked that they have the appropriate size and the errors are being thrown.
The tests for these issues are implemented.

PR #1314, opened 2016-02-11
JCF: new Tikhonov implementation done

jcollfont:
Finished the implementation of the Tikhonov module.

All the tests for input types and their configurations work. Could not figure out how make the functional tests to check the results of the inverses work. Testing on SCIRun, everything seems to be fine.

Will start separating the module into parent and child classes, so that new implementations of similar inverses can make use of this


dcwhite:
Does this PR obviate #1245?

src/Modules/Legacy/Inverse/Tests/TikhonovFunctionalTest.cc
```diff
@@ -185,3 +194,585 @@ TEST_F(TikhonovFunctionalTest, DISABLED_loadIDNonSquareFwdMatrixANDRandDataDiffS
...
+
+    
+    // check result
+    ASSERT_NEAR( residual.norm(), 0,  abs_error );
+    
+}
+
+
+// ID forward matrix with residual regularization matrix and single lambda
+TEST_F(TikhonovFunctionalTest, functionTestIDFwdMatrixANDRandDataInputIDResidualReguWithSingleLambda)
```
dcwhite: @jcollfont Rather than leave commented-out test code, if you can't get a test working, just add `DISABLED_` to the start of the test name and it won't be run as part of the suite. This will keep the code from getting stale.

src/Modules/Legacy/Inverse/SolveInverseProblemWithTikhonovImpl.cc
```diff
@@ -300,453 +561,210 @@ void TikhonovAlgorithmImpl::run(const TikhonovAlgorithmImpl::Input& input)
...
-      }
-      else
-      {
-        inverseMatrix_.reset(new DenseMatrix(forward_transpose * regForMatrix.inverse()));
-      }
-      inverseSolution_.reset(new DenseMatrix(*inverseMatrix_ * *measuredData_));
+        LinearAlgebra::solve_lapack(G,  y, b);
     }
-    else
+    catch (LinearAlgebra::LapackError&)
```
aylakhan: Lapack exception?

aylakhan:
Can solve_lapack be renamed? Also, what is the purpose of the LapackError exception?

aylakhan:
Also, can you delete the macro'd out tcl_comand code? We won't be needing that.

dcwhite:
@jcollfont Please respond to all the above questions.

jcollfont:
Answering questions in order:

solve_lapack erased as well as the error exception. 
Now we are directly using lu().solve(); 
The only issue with this method is that it assumes invertibility and throws no exception in case it fails.


tcl_command eliminated

jcollfont:
The problems with issue #956 regarding the input of regularization matrices should be solved at this moment.

aylakhan:
If you're really concerned about error checking, there are instructions for overriding the standard eigen assert macro here: http://eigen.tuxfamily.org/dox/TopicAssertions.html.

aylakhan:
Could you please disable the commented out tests in TikhonovFunctionalTest.cc as Dan described? I'll merge the pull request once that's done.

jcollfont:
@aylakhan I tried setting them as disabled, but the test functions do not even compile. I will just erase them from the branch I am pushing and keep a local copy that I'll work with

PR #1327, opened 2016-02-20
Text/ Widget Movement/ Clipping plane

cbrightsci:


dcwhite:
By default, the point probe is gigantic. In SCIRun 4, the default size is pretty small. Can we find a happy medium?

dcwhite:
Otherwise, the select-drag-update mechanism is working great!

dcwhite:
ShowColorMap text looks great too

PR #1339, opened 2016-02-26
Unit test setvalue function

moritzdannhauer:
The function set_value() breaks on tet meshes - it seems like a bug.


dcwhite:
I'll look at this one today.

dcwhite:
@moritzdannhauer I ended up merging the first one--can you pull SCIInstitute/master into this PR so I can test this too?

moritzdannhauer:
I just did. Does it work ?

dcwhite:
Looks like the merge commit is not pushed here yet.

dcwhite:
Now it's there, and travis has it. Thanks.

src/Core/Datatypes/Legacy/Field/Tests/VFieldTests.cc
```diff
@@ -135,4 +134,44 @@ TEST(VFieldTest, TetVolMeshAddValuesLinearBasis)
...
+  FieldInformation fieldinfo("TetVolMesh", 0, "double");
+  FieldHandle field = CreateField(fieldinfo);
+  
+  ASSERT_TRUE(field.get() != nullptr);
+  
+  VMesh *vmesh = field->vmesh();
+  VField *vfield = field->vfield();
+  vfield->resize_values();
+  std::cout << "An error will appear in the next line. Why is that? 2/2" << std::endl;
+  vfield->set_value(0, 1.0);
```
dcwhite: @moritzdannhauer 
Two problems with this.
* The order of arguments is reversed. The first parameter is the value, the second is the index. So the code should say
```
vfield->set_value(1.0, 0);
vfield->set_value(2.0, 1);
etc
```
* The code is supposed to fail. You have created an empty field; its data vectors are hence empty. You're calling an indexing-like operator on the vectors using `set_value`. It's the equivalent of doing the following with a normal vector:
```
std::vector<double> empty;
empty[0] = 1.0;
```
And generating the same crash, a range error. A valid test will call some other function to allocate 4 data values, or make sure the field is created with all 0 values. 

If the test is supposed to indicate that the field creation code you use *should* be allocating the internal data vectors already, this is a feature request, not a bug. The rest of SCIRun Field code is working fine with respect to data buffers.

PR #1364, opened 2016-03-24
createGeometricTransform non-Widget solution

cbrightsci:


dcwhite:
@cbrightsci The UI is a bit cramped on OSX. Up the dialog minimum size by a few pixels.

dcwhite:
There are many different module parameters to test.
* Can you make notes in your sample network with the expected transformed mesh? It's hard to tell from the UI settings, and 3 of the outputs are the same (gray/green/red all the same, blue different, is this correct?)
* This would be a good test case for a parameterized module unit test. Let's make an issue for that.

PR #1389, opened 2016-03-30
Create custom name in showfield for viewscene object list

cbrightsci:


dcwhite:
The name is passed through. If one changes it and reexecutes, both the current and previous name appear in the object selection menu, until you click it. This will be confusing for users.

dcwhite:
I also verified the name is saved in the network file.

PR #1417, opened 2016-05-07
Convert real to complex matrix

moritzdannhauer:
As requested ...


moritzdannhauer:
There was no algorithm class for the converter modules... so I did not make one.
Why does Travis fail?

dcwhite:
Ran out of time, as usual. I'll run it again.

PR #1418, opened 2016-05-09
Fixed_refinetetmeshlocallytests

moritzdannhauer:
@dcwhite: Is that what you had in mind?


dcwhite:
I'll compare the build times and see how much effect this had.

dcwhite:
Compile time went from 50 seconds to 14 seconds. Pretty good.

moritzdannhauer:
Glad to hear that!

Issue #1152, opened 2015-07-23
SetFieldData doesn't produce error when matrix/field doesn't match up


dcwhite:
It just output a null field instead.

dcwhite:
When the matrix has too few matrix values (rows), an error is produced but it's the wrong message: ` Internal error (data not tensor)`. When there are too many rows, the data is all set to zero. Both cases should be tested and fixed.

dcwhite:
Found out more about what's going on: algo is falling into a case where it's trying to match a single tensor in a matrix (i.e. 1x9 or 1x6) to all node values, but the field isn't set to accept tensors. These cases should be cleared up somehow.

SeyhmusGuler:
It doesn't give an error when data is a sparse matrix. ReportFieldInfo produces an error of 'Input data required on port InputField' afterwards though. I thought this error is also related to SetFieldData not properly creating the new field.

dcwhite:
@jessdtate Could you enumerate the various combinations of valid inputs and sizes, and test each one to see if they behave as expected? This task would be good to delegate to another student. Just @ them here and I'll assign it.

jessdtate:
What a poorly written issue.  I will go back and test this again.  SCIrun has a thing where if there is an error in the module, it will disappear if there is an execute command, but no execution (no change in inputs).  It could have been a normal error that I didn't see.

jessdtate:
So the possible combinations are:
# nodes x 1 -- scalar on nodes
# elem x 1 -- scalar on elems
# nodes x 3 -- vectors on nodes
# elem x 3 -- vectors on elems
# nodes x 6 -- tensors on nodes
# elem x 6 -- tensors on elems
# nodes x 9 -- tensors on nodes
# elem x 9 -- tensors on elems

jessdtate:
I have tried all those combos on a few fields, and I cannot replicate the error.  However, I noticed that when trying to set tensors, the data doesn't set properly (tensor of zeros). should I make a new issue for that or post the example here.

jessdtate:
Set field data was producing error messages when the fields didn't match.  in every case that I tried.

dcwhite:
There are also cases in the code where if a single value (scalar/vector/tensor) is readable, it's applied to all nodes/elems. We could remove that case. Also, what happens if num nodes == num elems?

dcwhite:
And yes, just post the example here.

jessdtate:
[...]

PR #1491, opened 2016-07-05
fix for SetFieldData tensor mapping

jessdtate:
this is for the tensor problem mentioned in issue #1152.  setfielddata should now properly set tensor data to a field.  also included is a fixe to a problem with a single vector input, which would also provide an zero vector.  

Note @cbrightsci : the null tensors were visible to showfieldglyphs module, but this fix will not allow the tensor tab to pop up.


Mazze:
Hi 
is it possible, that the change to fix the module SetDatafield brook the setting of a 9 valued tensor? In the commit 'fix the tensor data setting for one case' (9c75cb0) in the new line 394 the tensor values are extracted from the data. The assumption seems to be a symmetric tensor. Therefore, I would expect reading the values of the array at 0,1,2**,4,5**,8; However it is 0,1,2**,3,4**,8. The problem still exists in the current commit (f721f54).
The result is, that SCIRun 4.7 and 5.0.beta.d simulation give different results.

dcwhite:
@Mazze Could you open a new issue, and attach two networks (4.7 and 5.0) that demonstrate the problem? Once that's done I can have a fix in quickly.

PR #1552, opened 2016-09-16
Improved progress bar

dcwhite:
This is a first pass at a more "colorful" progress bar that summarizes the state of the network while it's executing.


dcwhite:
@aylakhan The PR review process is new, too. Want to try it out?

PR #1554, opened 2016-09-22
Fix line-ending problem for ascii fields

dcwhite:
This will fix the problem going forward. For old versions of Windows-saved ASCII field files, first run a line-ending conversion tool on them to read them in Linux.


aylakhan:
Does this write out ASCII fields as binary?

dcwhite:
No.

dcwhite:
The only difference is that line endings are preserved as LF only.

aylakhan:
If that's the case, can you put the same fix in ObjToFieldReader::write too?

dcwhite:
Sure

PR #1571, opened 2016-11-15
Closes #1568

### LINKED ISSUE #1568: over replacing %SCIRUNDATADIR% variable
### Description

When typing a string that matches the SCIRUNDATADIR string in any module, it replaces it with %SCIRUNDATADIR%.  This seems to happens in every text field in every module, but is probably the most problematic in iterfacewithpython.  This should replacement should be limit to the read and import modules.  The export write modules should still interpret the %SCIRUNDATADIR% as the actual path. 

### Steps to Reproduce

1. open a module with text like CreateString
2. type the string that matches the SCIRUNDATADIR, ie, /Users/jess/software/FwdInvToolkit_mine, in the text field

**Expected behavior:** 

The string should stay the same unless it is one of the read or import module

**Actual behavior:** 

changes to %SCIRUNDATADIR%.  In IWP this changes a string that can be used as a path to one the the python interface doesn't recognize.

### Versions
scirun v5.0-beta.E-96-g17438de
os x 10.10 and 10.8

dcwhite:
Fixes over-replacing of SCIRUNDATADIR

### DIFF
--- a/src/Core/Algorithms/Base/AlgorithmBase.cc
+++ b/src/Core/Algorithms/Base/AlgorithmBase.cc
@@ -80,7 +80,7 @@ void Variable::setValue(const Value& val)
   value_ = val;
 
   {
-    if (boost::get<std::string>(&val))
+    if ("Filename" == name_.name() && boost::get<std::string>(&val))
     {
       auto stringPath = toString();
       {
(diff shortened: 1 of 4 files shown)


dcwhite:
After this is merged I'll make the release.

PR #1598, opened 2017-03-03
Better complex matrix construction implementation

dcwhite:
This implementation uses Eigen operators. It can construct a giant (Moritz-scale) sparse complex matrix; the previous implementation took forever/ran out of memory.


dcwhite:
Yes, in @moritzdannhauer's previous PR.

aylakhan:
Found them - thanks! Test names are interesting.

PR #1601, opened 2017-03-16
Implement module ShowString

dcwhite:
Improved UI compared to 4. Possibly some stability issues on the rendering side; more extensive testing is required.


dcwhite:
@jessdtate Test this today?

jessdtate:
there was a problem with the default font that was causing some other problems.  Other than that, it was working fine.

dcwhite:
The auto-execute widgets were still a problem, so I'll just remove that feature from those widgets for now.

dcwhite:
@jessdtate Please test again.

jessdtate:
The font is working now, and I can change all the options.  Changing the coordinates does not seem to move the text.  Maybe it is not hooked up?

dcwhite:
Oops, got a bit excited with the double-slashes there. Coordinates should work now.

jessdtate:
everything seems to work now.

PR #1603, opened 2017-03-22
Convert_complex_to_real_matrix

shayestehfard:
This Module converts complex matrixes to real ones. It works in both sparse and dense matrix cases and shows the real part and imaginary parts of the matrix separately as the output. Also it shows the minimum and maximum values for each of these output matrixes.


src/ExampleNets/regression/ConvertComplexToRealMatrix.srn5
```diff
@@ -0,0 +1,218 @@
...
+		<item_version>0</item_version>
+	</disabledConnections>
+	<moduleTagLabels class_id="22" tracking_level="0" version="0">
+		<count>0</count>
+		<item_version>0</item_version>
+	</moduleTagLabels>
+	<loadTagGroups>0</loadTagGroups>
+</networkFile>
+</boost_serialization>
+
```
dcwhite: No errors here, but the test network should have something connected to the module you're testing, at least to assert it creates some output. In this case, ReportComplexMatrixInfo would suffice.

src/Modules/Math/ConvertComplexToRealMatrix.cc
```diff
@@ -0,0 +1,102 @@
...
+  if (needToExecute())
+  {
+    update_state(Executing);
+    
+    auto input_dense = boost::dynamic_pointer_cast<ComplexDenseMatrix>(input_complex_matrix);
+    auto input_sparse = boost::dynamic_pointer_cast<ComplexSparseRowMatrix>(input_complex_matrix);
+
+    if (!input_dense && !input_sparse)
+    {
+     error("Number of Rows or Columns are zero");
```
dcwhite: Copy/Paste error message--please correct.

PR #1604, opened 2017-03-24
Travis/Clang with new timeout limit can handle full build

dcwhite:
I'll figure out how many unit tests can be run on the clang side later.


dcwhite:
@aylakhan Could you check this one out, should only take you 2 minutes.

PR #1630, opened 2017-06-02
Improved subnet GUI

dcwhite:
Still a lot to do, and some unfinished pieces with connections. It's worth a testing lunch to find any obvious bugs.


dcwhite:
While in dreamland, I figured out a way to get connections working. Should be added here soon.

PR #1643, opened 2017-06-14
Closes #1642

### LINKED ISSUE #1642: Interactive Execution with GetSliceFromStructuredFieldByIndices
### Description

Parsing through an image stack with GetSliceFromStructuredFieldByIndices and a showfield is a little tedious right now as I need to right click->execute in the module UI every time I move the slider. I would like the option to have this module execute interactively on release of the mouse when I move the slider around.

dcwhite:
All widgets in GSFSFBI trigger execution now.

### DIFF
--- a/src/ExampleNets/regression/probeMovesWithSliceSnappingToNode.srn5
+++ b/src/ExampleNets/regression/probeMovesWithSliceSnappingToNode.srn5
@@ -246,7 +246,7 @@
 									<name>ZLocation</name>
 									<value>
 										<which>1</which>
-										<value>4.28571428571428381e-01</value>
+										<value>1.42857142857142794e-01</value>
 									</value>
 								</second>
 							</item>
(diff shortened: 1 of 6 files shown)


dcwhite:
So turn off spinbox update? Yeah I can add some more checkboxes

jessdtate:
Yeah, It would be nice to have all three options

dcwhite:
OK

dcwhite:
Check out the latest version of this branch to make sure the checkboxes work OK.

dcwhite:
This will also need a fresh review from somebody

jessdtate:
This doesn't quite behave as I would expect.  I think it should be normal interactive with the interactive mode selected without the other checkmarks, ie, execute as it is moved.  then check the execute on release   (which  I think should be default) to enable that option.  Unchecking the interactive mode would mean you have to hit the execute button.  This would mean you don't need the execute on click option for the axes.  

If this is harder to implement, then the interactive mode could be execute on release.

dcwhite:
It was tricky due to there being two widgets for a single state value. I should have removed the global checkbox, to make it less confusing, which I can do now.

PR #1645, opened 2017-06-19
new Boolean comparison module

jessdtate:
This module allows users to compare two matrices and return different outputs based on the comparison. Also a quit option.  

Some fixes to the Module ui code to enable pulling list of options from module code.

Some fixes to Module creation document.


src/Modules/Math/BooleanCompare.cc
```diff
@@ -0,0 +1,110 @@
...
+    std::string then_statement = state->getValue(Parameters::Then_Option).toString();
+    std::string else_statement = state->getValue(Parameters::Else_Option).toString();
+    
+    auto out_mat=output.get<DenseMatrix>(Variables::Solution);
+    double *data=out_mat->data();
+    
+    if ((data[0]==1 && then_statement == "quit") || (data[0]==0 && else_statement == "quit"))
+    {
+
+      std::cout<<"cond matrix :"<<*out_mat<<std::endl;
```
aylakhan: Log instead of print?

dcwhite: Use either `remark()` or `LOG_DEBUG`

dcwhite:
@jessdtate Please add a test network

src/Modules/Math/BooleanCompare.cc
```diff
@@ -0,0 +1,110 @@
...
+      std::cout<<"ordered to quit.  Trying to quit."<<std::endl;
+//      quit nicely
+      std::string q_message=NetworkEditorPythonAPI::quit(false);
```
dcwhite: This part forces BUILD_WITH_PYTHON to be enabled. To fix the gcc build (which is run without python enabled), you can make your entire module require python, or just this section. 

If the former, toggle including the file in the CMakeLists and surround the whole file in `#ifdef BUILD_WITH_PYTHON`. If the latter, just surround the quit code with the same macro.

src/Modules/Math/BooleanCompare.cc
```diff
@@ -93,12 +94,20 @@ BooleanCompare::execute()
...
 //      quit nicely
+      LOG_DEBUG("cond matrix :"<<*out_mat<<std::endl);
+      remark("ordered to quit.  Trying to quit.");
+      std::ostringstream ostr;
       std::string q_message=NetworkEditorPythonAPI::quit(false);
-      std::cout<<"Quit message = "<<q_message<<std::endl;
+      ostr << "Quit message = "<<q_message;
+      remark(ostr.str());
+#else
+      remark(" cannot quite without building with python");
```
dcwhite: s/quite/quit

dcwhite:
I hooked up a bot that should have fixed that automatically, but I guess it's not working yet.

dcwhite:
Still need that test network!

dcwhite:
thanks

src/Interface/Modules/Math/BooleanCompareDialog.cc
```diff
@@ -0,0 +1,103 @@
...
+                condition_.insert(StringPair("A is less than B (<)","lessop"));
+                condition_.insert(StringPair("A is less or equal to B (<=)","lesseqop"));
+                condition_.insert(StringPair("A is equal to B (==)","eqop"));
+                condition_.insert(StringPair("A is greater than B (>)","greatop"));
+                condition_.insert(StringPair("A is greater or equal to B (>=)","greateqop"));
+                
+                then_.insert(StringPair("Return: null","null"));
+                then_.insert(StringPair("Return: first input","first"));
+                then_.insert(StringPair("Return: second input","second"));
+                then_.insert(StringPair("Return: thrid input","third"));
```
dcwhite: `s/thrid/third/`

PR #1661, opened 2017-06-29
Package fix

aylakhan:
Add script to fix Mac OS X installer packages generated by CPack. Run the repair_package.sh script after calling 'make package' or 'ninja package' etc. Fix executable paths for libraries in app bundle dependent on Python.


dcwhite:
Cool, I will test this tonight

jessdtate:
I get an error trying to build:
CMake Error: File /Users/jess/software/SCIRun_mine/src/CMake/Modules/CPack.Distribution.dist.in does not exist.
CMake Error at CMakeLists.txt:976 (CONFIGURE_FILE):
configure_file Problem configuring file

PR #1678, opened 2017-07-27
Bundle fix

dcwhite:
I had "fixed" the BUILD_BUNDLE flag, but that broke the actual bundle making. This should fix all OSX builds.


dcwhite:
I forgot to test the headless build. Let me work on that now

PR #1682, opened 2017-08-10
AppendMatrix_reimplementation

moritzdannhauer:
This is an extension of the module AppendMatrix to get the full and expected functionality known from SCIRun4, i.e., +support for sparse/densecolumn matrices and use of dynamic input ports.
Please review code and test, thanks!


jessdtate:
mostly works fine. I found a problem with testing, I can't append dense single column matrices  row wise.  column wise works.  The transform is also a problem, ie, row matrices combined column wise.  this is not a problem with sparse matrices (either row or column matrix).

src/Core/Algorithms/Math/AppendMatrix.cc
```diff
@@ -24,68 +24,120 @@
...
 
-  if (params == ROWS)
+AppendMatrixAlgorithm::Outputs AppendMatrixAlgorithm::run(const AppendMatrixAlgorithm::Inputs& input, const AppendMatrixAlgorithm::Parameters& params) const
+{ 
+  auto lhsPtr = input.get<0>();
+  auto rhsPtr = input.get<1>();
+  if (!lhsPtr || !rhsPtr)
+   error(" At least two matrices are needed to run this module. ");
+
+  if (!((matrixIs::sparse(lhsPtr) && matrixIs::sparse(rhsPtr)) || (matrixIs::dense(lhsPtr) && matrixIs::dense(rhsPtr)) || (matrixIs::column(lhsPtr) && matrixIs::column(rhsPtr))))
```
jessdtate: This seems to be where the issues of the dense column or row matrices are happening

jessdtate: line 88

moritzdannhauer:
Thanks for testing. I assume your comments regard DenseMatrix inputs rather than ColumnDenseMatrix module inputs. I fixed that issue and attached my testing networks for this.
Let me know if there any other issues.

PR #1690, opened 2017-09-06
Parallelize_GenerateStreamLines_module

moritzdannhauer:
This is a pull-request (for Dan) to look more closely to small solution discrepancies that I found.


dcwhite:
@moritzdannhauer How about we merge this change as is, then add a checkbox to the UI for toggling single-multithreaded. Then comparison testing will be much easier.

PR #1701, opened 2017-09-29
Complex calculations et al

dcwhite:
See commit list for all closed issues. New complex number ones, and some old random ones.


dcwhite:
Need to fix non-mesh geometry objects

PR #1724, opened 2017-10-27
Cleanup_tet_mesh

moritzdannhauer:
Done!


dcwhite:
@moritzdannhauer Please use the new module factory config file approach! Never edit `HardCoded...Factory` any more.

moritzdannhauer:
How about now?

dcwhite:
This one needs a module count unit test fix

PR #1725, opened 2017-10-27
Remove unused nodes

moritzdannhauer:
Done!


dcwhite:
@moritzdannhauer Please use the new module factory config file approach!
BTW I'm not @Dan, but @dcwhite.

moritzdannhauer:
How about now?

src/Modules/Factory/Config/RemoveUnusedNodes.module
```diff
@@ -0,0 +1,19 @@
+{
+  "module": {
+    "name": "RemoveUnusedNodes",
+    "namespace": "Fields",
+    "status": "In progress: needs more testing",
+    "description": "Removes meshnodes that are not connected in any mesh
```
dcwhite: @moritzdannhauer This is breaking the build. Newlines in these descriptions are not supported.

dcwhite:
Can you check that your code is building? Travis gets a linking error.

dcwhite:
@moritzdannhauer Similar fix needed at the algo level--count needs increasing:
```
[0;32m[ RUN      ] [mAlgorithmLayerTests.ListAllAlgorithms
algorithm factory size: 82
/home/travis/build/SCIInstitute/SCIRun/src/Core/Algorithms/Factory/Tests/AlgorithmLayerTests.cc:45: Failure
```

Issue #1692, opened 2017-09-11
matlab format cannot be read in python


jessdtate:
### Description

the matlab files saved out of SCIRun cannot be read by  python using the scipy.io.loadmat function.  The files load into Matlab fine, and once saved from matlab will open in python.  

### Steps to Reproduce

1. save a matrix or a field in matlab format
2. in python, run 
`import scipy.io
scipy.io.loadmat('matlab file')`
3.  in matlab load file and save it. 
try same python code as above

**Expected behavior:** 

File should load in to python properly

**Actual behavior:** 

I get the following error:
 File "<stdin>", line 1, in <module>
  File "/Users/jess/software/anaconda3/lib/python3.6/site-packages/scipy/io/matlab/mio.py", line 136, in loadmat
    matfile_dict = MR.get_variables(variable_names)
  File "/Users/jess/software/anaconda3/lib/python3.6/site-packages/scipy/io/matlab/mio5.py", line 272, in get_variables
    hdr, next_position = self.read_var_header()
  File "/Users/jess/software/anaconda3/lib/python3.6/site-packages/scipy/io/matlab/mio5.py", line 232, in read_var_header
    header = self._matrix_reader.read_header(check_stream_limit)
  File "scipy/io/matlab/mio5_utils.pyx", line 563, in scipy.io.matlab.mio5_utils.VarReader5.read_header (scipy/io/matlab/mio5_utils.c:5658)
  File "scipy/io/matlab/mio5_utils.pyx", line 615, in scipy.io.matlab.mio5_utils.VarReader5.read_header (scipy/io/matlab/mio5_utils.c:5577)
  File "scipy/io/matlab/mio5_utils.pyx", line 492, in scipy.io.matlab.mio5_utils.VarReader5.read_int8_string (scipy/io/matlab/mio5_utils.c:4690)
TypeError: Expecting miINT8 as data type

### Versions
scirun v5.0-beta.N-101-g6b70c91ac
os x 10.12, 10.10

jessdtate:
here are some test files:
from scirun
[test_Matrix.mat.zip](https://github.com/SCIInstitute/SCIRun/files/1294079/test_Matrix.mat.zip)
from matlab
[test_Matrix_matlab.mat.zip](https://github.com/SCIInstitute/SCIRun/files/1294081/test_Matrix_matlab.mat.zip)

dcwhite:
Related discussion:
https://github.com/scipy/scipy/issues/3288

[...]

PR #1727, opened 2017-10-27
Matlab strings--potential fix for scipy read error

dcwhite:
@jessdtate can you verify this fixes the scipy error referenced in #1692? Also check round-trip save/load for basically every other matlab file type and program that reads/writes such types...


jessdtate:
There seems to be a problem reading integer matrices into scirun now.  Before, they could be read from matlab.  Here is the error from a saved matrix (same from Python and matlab):

Message: Import failed. 
ERROR: /Users/jess/software/SCIRun/src/Modules/DataIO/GenericReader.h(190): Throw in function virtual void SCIRun::Modules::DataIO::GenericReader >, SCIRun::Modules::MatrixPortTag>::execute() [HType = boost::shared_ptr >, PortTag = SCIRun::Modules::MatrixPortTag] Dynamic exception type: boost::exception_detail::clone_impl std::exception::what: Import failed. [SCIRun::Core::tag_error_message*] = Import failed. 

[Matlab_save_int_matrix.mat.zip](https://github.com/SCIInstitute/SCIRun/files/1432242/Matlab_save_int_matrix.mat.zip)
[python_save_int_matrix.mat.zip](https://github.com/SCIInstitute/SCIRun/files/1432243/python_save_int_matrix.mat.zip)

jessdtate:
looks like that ints on matrix doesn't work with ports at all.

dcwhite:
OK I'll take a look

jessdtate:
a slight correction,  ints from matlab didn't work either, but the error was just: import failed.

jessdtate:
I can get it to work with any fields.  I still get the error: TypeError: Expecting miINT8 as data type

dcwhite:
OK I'll close this and come up with a different fix.

jessdtate:
Maybe we should merge this into master?   It works with matrix files, so it's half fix at least.  I've been including it in local builds for a while to deal with matrix files.

dcwhite:
I can see if I can merge it in with a feature flag--enabled at run-time, or a setting which you can save as turned on.

jessdtate (approved):

PR #1732, opened 2017-11-08
TMS_modules

moritzdannhauer:
Converted 4 scirun4 modules and support classes for quasi-static TMS modeling
Implementation: Petar Petrov


jessdtate:
I don't think the units should be hard coded to the module UI.  If anything, this should be an editable text box.

jessdtate:
Since the modules all call the same algorithms class, it seems like it would be better if the coil modules were combined to one module.

jessdtate:
The link to the documentation goes to the brianstimulator page.  This is ok temporarily, but they should really have module documentation.

jessdtate:
@dcwhite you can go ahead and merge this

PR #1739, opened 2017-12-15
Qwt for Tikhonov L-curve and a couple other small issues

dcwhite:


dcwhite:
This will require a rebuild of Superbuild to get the newest external.

jessdtate:
@dcwhite Do you want to check the commit I did?  it works for me now.

PR #1741, opened 2018-01-09
Cleaver 2

dcwhite:
Work in progress, not ready to merge.
Remaining tasks:
- [x] Hook up sizing field port
- [x] Hook up background field port
- [x] Fix re-execution problem (applies to Cleaver1 too)
- [x] Expose mesh type choice
- [ ] Fix ... whatever problem users find.

@jessdtate Please provide a network that uses sizing and background field ports, so I have something to test.


dcwhite:
Right, Cleaver2's generated sizing field is cell-centered. I can map it over to node-centered but the user can also do that with a module--I'd prefer the latter to keep this module simpler.

dcwhite:
Right, Cleaver2's generated sizing field is cell-centered. I can map it over to node-centered but the user can also do that with a module--I'd prefer the latter to keep this module simpler.

dcwhite:
Discussed with @jessdtate, we changed the output latvol sizing field to node-centered and it looks fine now.

PR #1751, opened 2018-03-02
LinePlotter module

dcwhite:


dcwhite:
@jessdtate Command to update saved networks:
`find . -type f -exec sed -i.bak s/AdvancedPlotter/LinePlotter/g {} \;`

PR #1783, opened 2018-06-06
Closes #1781

### LINKED ISSUE #1781: Change boost version to 1.66
Boost 1.67 introduced changes in the build system and boost-python library names, which SCIRun will need to be fixed for. Rather than try to figure that out, let's use 1.66 instead, which has the new library Stacktrace and most of the pertinent updates.

1.67 seems to work fine on Windows and Linux, but the build is failing on Mac--which didn't show up unless you built from scratch. Build failure found by @chhabragarima (thanks!)

dcwhite:


dcwhite:
This one isn't quite ready--I need to redo the boost repo branch.

dcwhite:
OK this is ready.

PR #1814, opened 2018-07-12
Calc mesh center

chhabragarima:
CalculateMeshCenter Converted


dcwhite:
@chhabragarima The preference is to convert any legacy algorithm code "in place." Can you move the converted algo files back to Core/Algorithms/Legacy/Fields/MeshDerivatives/? Just overwrite the old ones.

chhabragarima:
Do I need to change the name of the files as the original ones?

dcwhite:
Yes, they should match exactly, we don't need the old unconverted code laying around.

src/Core/Algorithms/Legacy/Fields/MeshDerivatives/CalculateMeshCenterAlgo.cc
```diff
@@ -0,0 +1,165 @@
...
+  FieldHandle output;
+  
+  //pull parameter from UI
+  std::string method=getOption(Parameters::Method);
+  
+  
+  //Safety Check
+  if(!inputField)
+  {
+    error("No input field");
```
jessdtate: you can use `THROW_ALGORITHM_INPUT_ERROR("No input field")`  instead of the everything in this if statement

src/Core/Algorithms/Legacy/Fields/MeshDerivatives/CalculateMeshCenterAlgo.cc
```diff
@@ -0,0 +1,165 @@
...
+  fo.make_nodata();
+  
+  /*if(inputField->vmesh()->num_nodes()==0)
+  {
+    warning("Input field does contain any nodes, output will be an empty field");
+    output=CreateField(fo);
+    AlgorithmOutput result;
+    result[Variables::OutputField] = output ;
+    return result;
+  }*/
```
jessdtate: get rid of commented out code.  especially since you seem to have something that replaces it below

PR #1843, opened 2018-08-28
Gui improvements

dcwhite:
Improvements to code editor modules and ShowField UI


jessdtate:
The parser help pops up right behind the modules (calculatefielddata, etc), making it difficult to see or completely obscured.  This is a bigger problem with the new vertical format.  could we make it open on top of the module UI?  it may be problem for some new users.

dcwhite:
Sure I can fix that

Issue #1468, opened 2016-06-21
diagonal glyphs not visualizing properly


jessdtate:
### Description

when vectors align on the diagonal (1,1,1), (-1,-1,-1), or a scaled version of one of the two the vectors do not display properly.  the lines glyph is the only glyph that will display properly, the others will mostly not display, but may show part of the glyph, for instance the shaft of the arrow, depending on location of the origin of the vector.  vectors display fine if the values are changed slightly, ie, (1,1,.99).  This error is likely do to some numerical error.

[point_vectors.srn5.zip](https://github.com/SCIInstitute/SCIRun/files/326624/point_vectors.srn5.zip)
### Steps to Reproduce
1. create a vector with direction +/- (1,1,1)
2. visualize with showFieldGlyphs.
3. try different glyphs.

**Expected behavior:**
The all glyphs should be displayed.

**Actual behavior:** 
Glyphs do not appear, or partially appear depending on vector location.  
### Versions

scirun v5.0-alpha.AF-6-g8816a3f
os x 10.8 and 10.10

dcwhite:
@SCIInstitute/cibc-users From now until IBBM, please mark your new issues with "NEEDED FOR IBBM 2016" if it is that high priority. Otherwise, we will push the issue until after IBBM.

jessdtate:
This one is needed for IBBM, so:

NEEDED FOR IBBM 2016

PR #1850, opened 2019-01-09
resolved issue #1468

Haydelj:
Fixed glyph error at (1,1,1). Added a test for this case and a fix.


src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -276,9 +276,12 @@ void GlyphGeom::generateCylinder(const Point& p1, const Point& p2, double radius
   if (num_strips < 0) num_strips = 20.0;
   double r1 = radius1 < 0 ? 1.0 : radius1;
   double r2 = radius2 < 0 ? 1.0 : radius2;
-
+  
   //generate triangles for the cylinders.
-  Vector n((p1 - p2).normal()), u = (10 * n + Vector(10, 10, 10)).normal();
+  Vector n((p1 - p2).normal());
+  Vector u = (10 * n + Vector(10, 10, 10)).normal();
+  if(n.x() == n.y() && n.x() == n.z())u = (10 * n + Vector(10, 0, 0)).normal();
```
```suggestion
  if (n.x() == n.y() && n.x() == n.z())
    u = (10 * n + Vector(10, 0, 0)).normal();
```

dcwhite: The code fixes @jessdtate's use case. First, let's improve the spacing.

dcwhite: However, I'm concerned about the magic number `10` appearing twice, and the check comparing doubles exactly. We can discuss at today's meeting.

src/Core/GeometryPrimitives/Vector.h
```diff
@@ -384,6 +385,17 @@ inline Vector Cross(const Vector& v1, const Vector& v2)
 		v1.d_[0]*v2.d_[1]-v1.d_[1]*v2.d_[0]);
 }
 
+//returns a unit vector
```
dcwhite: `const double tangent_threshold = 1 / sqrt(2.0);`

dcwhite:
@Haydelj Master is merged, need above small updates.

PR #1855, opened 2019-01-15
Python choice

dcwhite:
Makes newer Python versions available, and updates default Python to 3.5.

A complete "version matrix" is coming soon to the docs, but here is a summary:
* Mac ought to support 3.5 from an installer, and 3.6 from a user build.
* Linux can support at least 3.5 and 3.6. 
* As of right now (still building), I think Windows will support 3.7.

The user can choose which version to build from the Superbuild config stage, using variable USER_PYTHON_VERSION, which provides a dropdown menu of choices.

This also includes a checkbox preference to disable the holiday backgrounds.


dcwhite:
Hooray, a clean build. @jessdtate let's get this in today.

Issue #1083, opened 2015-07-02
memory increase when objects turned off in ViewScene


jessdtate:
When objects are turned off in the view scene, the memory increases, depending on the size of the geometry.

jessdtate:
quite important

dcwhite:
@jessdtate IBBM. Based on the labels, I'd say this one is pretty high priority.

dcwhite:
Status: waiting on SRInterface refactoring

PR #1885, opened 2019-03-08
Proposed Fix For Memory Duplication

Haydelj:
A proposed fix for issue #1083


dcwhite:
We should test this with both large objects in ViewScene and with many small/medium objects, both watching the memory usage and VS framerate.

dcwhite:
This one is a redo

Issue #1846, opened 2018-10-02
Sumientra issues


dcwhite:
> I’ve been trying to convert my SCIRun 4 networks to SCIRun 5. Generally it works. Most modules can be easily converted. What’s weird is that none of the module settings are converted, even though the importLog does show an entry for each setting. This is manageable though.

- [x] We're have experimental code written to convert a lot of the module settings. It's waiting on the compiler upgrade that's going on now.

> The main problem is that running anything in 5 is very slow. All modules seem to take much longer than I’m used to. I timed it for a simple network, ran identical nets in 4 and 5, and it was 12 vs 24 seconds. I also get a lot of spinning wheel, just for trying to open a module GUI. 

- [x] Would you be able to share the slow networks and data? That would help with debugging.

> And visualizing a full head mesh (4M elements) is almost impossible; it takes several minutes of spinning wheel for ViewScene to show the mesh, and then the wheel pops back up every time I even touch the ViewScene window (or anything else really). Visualizing a surface or a very small volume mesh does work. Is there anything I can change to improve this? I installed this: SCIRun-5.0.beta.CinC2018-macOS.pkg.

- [x] In general we've seen faster rendering in 5. What is your graphics hardware?

> The Matlab import modules work well. Being able to search for a file through the standard Mac interface is bliss. Only issue is that after opening a file, the module does not show the available fields, so you first have to run the module (causing the entire network to run and crash) in order to see the fields and select which field you want to load on each port. It would be great if this could be fixed.

- [ ] There are workarounds for this, I will check out that module to see if there is a slicker way.

> I have some more questions:
> 
> It seems that many of the ViewScene settings cannot be saved. Location of the window and orientation of the scene, but also some of the settings that can be un/checked in the Settings popup. It all resets when loading the network, and I cannot find any of these settings in the html file. Is this correct?

- [x] That's also a known problem, and I will raise the importance of the corresponding issue.

> Since I need to be able to run SCIRun from command line, I also downloaded the source. It seemed to build OK, though it gives a string of ‘warning: this version of Mac OS X is unsupported’, while I’m on 10.10.5 and the website says it’s tested for 10.8-10.11. 

- [x] We need to update the website doc for supported OS versions, since it's been a while. I might be dropping 10.10 at some point soon, probably after 10.14 comes out.

> The result (/bin/SCIRun/SCIRun_test) does open normally, but I can’t access the menu, so can’t insert any modules. 

[...]

Issue #1463, opened 2016-06-20
ViewScene "AutoView" issues with GeneratePointProbe widget or single node rendered as sphere


ajanson:
### Description

Clicking autoview with a single GenerateSinglePointProbeFrom field widget or a single node passed into showfield rendered as a sphere zooms in really far and seems to lock the viewscene until a different object is added and autoview is pressed again.
### Steps to Reproduce
1. Connect first port of GenerateSinglePointProbeFromField to viewscene
2. Click autoview in the viewscene
3. It zooms in really far and another object needs to be added to get back.

**Expected behavior:** It should zoom in only until the whole sphere is visible in the window

**Actual behavior:** Zooms in way too far
### Versions

v5.0-alpha.AF-1-g483d529

dcwhite:
@ajanson This is still an issue, but could you attach a network that show the problem? I might be able to work on this one.

ajanson:
[pointProbe_ViewScene.zip](https://github.com/SCIInstitute/SCIRun/files/1411243/pointProbe_ViewScene.zip)

dcwhite:
I'll make a build with #1888 in it for @ajanson to test.

Issue #952, opened 2015-05-06
Investigate pixelation of surface faces


dcwhite:
Mesh description and v4 comparison coming soon.

![viewscene_2015 05 05 152438 635_1](https://cloud.githubusercontent.com/assets/1399383/7500554/8d589492-f3eb-11e4-9d53-623fe89a9ec5.png)
![viewscene_2015 05 05 152604 578_2](https://cloud.githubusercontent.com/assets/1399383/7500553/8d5857d4-f3eb-11e4-8d07-7fc5a3f3d19e.png)

PR #1888, opened 2019-03-13
Draw Order Fix

Haydelj:
I used a bounding sphere around the scene bounding box to generate near and far clipping planes in order to reduce Z-Fighting. Right now the bounding sphere generation is tied to auto view. Before auto view is pressed we use the default behavior with the near clipping plane set to 1.0. I also fixed SRInterface so that it actually passes the projection matrix from SRCamera to spire.


dcwhite:
@Haydelj Is this tied to a specific issue #? If so please indicate. Also, it would be nice if you provided some example "before/after" screenshots attached here. Thanks.

jessdtate:
This was reported in Sumientra's issue #1846.

jessdtate:
also #952

Haydelj:
Now also fixes auto view scene for glyphs resolving issue #1463

dcwhite:
This one will be after #1896

dcwhite:
@Haydelj TODO: add toString(), move font files back

jessdtate:
@Haydelj found some bugs with the math he was using for this fix.  Stay tuned for an update

Haydelj:
Before:
Screen Shot 2019-04-04 at 12 38 47 PM
Screen Shot 2019-04-04 at 12 40 04 PM

After:
Screen Shot 2019-04-04 at 12 46 01 PM
Screen Shot 2019-04-04 at 12 47 17 PM

Issue #881, opened 2015-02-19
ViewScene doesn't initialize properly about 10-20% of the time


dcwhite:
Related to the promise system for retrieving assets, we believe. @iauns really needs to take a look.

dcwhite:
I took a tour down the rabbit hole and didn't really get anywhere.

dcwhite:
@aylakhan @cbrightsci Adding some capture code for this error which will stop the rendering loop if the axes don't get initialized after 50 frames. The slot turns the ViewScene module widget red. Testing it by adding multiple ViewScenes to an empty network reveals the error percentage is around 5%, but more often when it's the first one.

dcwhite:
See https://github.com/SCIInstitute/SCIRun/commit/b855797ac55212602be00f756012189fd822eb8e

dcwhite:
@jessdtate IBBM

aylakhan:
This is on @cbrightsci's renderer technical objectives list.

dcwhite:
I'm going to take another crack at this one since it was quite prevalent at this year's IBBM.

iauns:
I vaguely recall this from a while back. I remember thinking that we needed to reduce the complexity of the start up process and reduce the logic to something straight-forward. It also makes it easier if your rendering logic blocks for assets instead of using intermediates or waiting for them to load (SCIRun probably doesn't have a large number of assets to load or differing LODs for the individual objects). It also allows you to debug startup problems easier when looking at the command buffers generated by the user-mode OpenGL driver since you don't have initialization code spread over time. Feel free to ping me. I'll be in Utah this coming weekend if you want another pair of eyes on it.

dcwhite:
Thanks! But alas, I'll be in New York this weekend.

iauns:
Regardless, if you have any questions ping me via email or messenger and we can work through it.

dcwhite:
I burrowed a little bit deeper this time, and found a race condition occurring in the initialization phase, related to `ComponentContainer` and its `renormalize` function. I can write up a detailed report if you want.

aylakhan:
That would be a good idea.

dcwhite:
If we fix this, #738 will probably be resolved. So I'll close that one and just reference it here for testing.

dcwhite:
Will pair-debug with @jessdtate later this year.

dcwhite:
[...]

PR #1896, opened 2019-03-22
Red View Scene Fix

Haydelj:
A proposed fix for issue #881


src/Interface/Modules/Render/ES/Core.cc
```diff
@@ -47,6 +47,8 @@ namespace Render {
   ren::registerAll(*this);
   rendererRegisterAll(*this);
 
+  mCurrentTime = 0.0;
```
dcwhite:
```suggestion
  
```

dcwhite: Initialize the four uninitialized members in the parameter list above, like so:
`ESCore::ESCore() : mCoreSequence(0), mCurrentTime(0), mFPS(0), mLastRealTime(0), r_(0.0f), g_(0.0f), b_(0.0f), a_(0.0f)`

dcwhite:
This fixes the WindowsDebugBuildViewScene problem, so in all likelihood the red ViewScene problem is no more. Of course, now that debug build is working, it will expose a lot of other long-standing problems, which will be new issues. 

I'll have another review comment after thinking about the exception-throwing code. I think we should keep it and just increase the limit a bit, since there might be other bugs waiting to be unearthed that lead to the same failure.

PR #1910, opened 2019-04-03
Memory Duplication Fix

Haydelj:
Makes the transparency rendering system handle it's own sorted IBO's. Though this system does not have any garbage collection it allows us to enable garbage collection for everything else reducing memory usage dramatically. This addresses issue #1083 

Note: When we get a chance we should either completely rework the transparent object rendering system or implement some sort of garbage collection system for the sorted IBO's


dcwhite:
@Haydelj now this one needs a manual merge by you.

dcwhite:
Looks to be pretty easy though, just one extra field in Core.h

dcwhite:
I'll test this some more--Windows might be behaving better than Mac with the first GC fix.

jessdtate:
GC fix in master still crashes on OS X.  The performance seems the same with new fix, but doesn't crash.

PR #1918, opened 2019-04-19
Color Map Fix

Haydelj:
Fixes color maps not rendering after draw order fix and fixes UI elements clipping with objects.


dcwhite:
ShowColorMap is back and visible, but there's a possibly-related issue with it not showing up until ViewScene is resized (MacOS 10.10). I need to test this on other platforms and later Macs.

Issue #1897, opened 2019-03-22
Refactor SRInterface


dcwhite:
And related subobjects. This will follow closing of current PRs, so there are no bad merges. We'll aim to start at next Friday's meeting (3/29).

dcwhite:
TODO list created via whiteboard picture...needs uploading

Haydelj:
![IMG_6150](https://user-images.githubusercontent.com/28606560/55344608-072af380-546b-11e9-954f-344d153482b3.jpg)

PR #1924, opened 2019-05-02
Refactor for SRInterface ViewScene and SRCamera 

Haydelj:
Resolves Issue #1897 minus splitting SRInterface.cc and ViewScene.cc into multiple source files.


dcwhite:
MERGING TODAY no. 3

dcwhite:
@Haydelj Inlining one of the ESCore calls into the SRInterface header broke a unit test (link error) on Windows. It's a result of more code in the header == wider explicit link dependency (by exporting more symbols). The fix is to put a new SCISHARE on the ESCore class, and include `share.h`. I can fix this on my branch after the merge, but just be aware of this sort of thing in the future.

PR #1927, opened 2019-05-08
Update cleaver UI

allywarner:
InterfacewithCleaver2 names updated in module


dcwhite:
If you change the parameter names in the C++ code, it will break old networks. The safer way to do this is just to change the display names in the .ui file like you did, as well as the regular variable names in algorithm code. Could you redo this?

src/Core/Algorithms/Field/InterfaceWithCleaver2Algorithm.cc
```diff
@@ -60,8 +60,8 @@ using namespace SCIRun::Core::Algorithms::Fields;
 using namespace SCIRun::Core::Geometry;
 
 ALGORITHM_PARAMETER_DEF(Fields, Verbose);
-ALGORITHM_PARAMETER_DEF(Fields, VolumeScaling);
-ALGORITHM_PARAMETER_DEF(Fields, VolumeMultiplier);
+ALGORITHM_PARAMETER_DEF(Fields, VolumeSamplingRate);
```
dcwhite: So don't change this one...

src/Core/Algorithms/Field/InterfaceWithCleaver2Algorithm.cc
```diff
@@ -84,9 +84,9 @@ namespace detail
   static const double kDefaultAlpha = 0.4;
   static const double kDefaultAlphaLong = 0.357;
   static const double kDefaultAlphaShort = 0.203;
-  static const double kDefaultScale = 1.0;
+  static const double kDefaultSamplingRate = 1.0;
```
dcwhite: ...but this one is OK to change

src/Core/Algorithms/Field/InterfaceWithCleaver2Algorithm.cc
```diff
@@ -96,9 +96,9 @@ namespace detail
     cleaver2::MeshType mesh_mode;
     double alphaLong;
     double alphaShort;
-    double scaling;
+    double samplingRate;
```
dcwhite: this one OK

src/Core/Algorithms/Field/InterfaceWithCleaver2Algorithm.cc
```diff
@@ -508,8 +508,8 @@ InterfaceWithCleaver2Algorithm::InterfaceWithCleaver2Algorithm()
   addParameter(Parameters::Verbose, true);
   addParameter(Parameters::SimpleMode, false);
   addParameter(Parameters::ReverseJacobians, true);
-  addParameter(Parameters::VolumeScaling, detail::kDefaultScale);
-  addParameter(Parameters::VolumeMultiplier, detail::kDefaultMultiplier);
+  addParameter(Parameters::VolumeSamplingRate, detail::kDefaultSamplingRate);
```
dcwhite: this one (left of the comma) not OK (right of the comma OK)

src/Core/Algorithms/Field/InterfaceWithCleaver2Algorithm.h
```diff
@@ -39,8 +39,8 @@ namespace Algorithms {
 namespace Fields {
 
   ALGORITHM_PARAMETER_DECL(Verbose);
-  ALGORITHM_PARAMETER_DECL(VolumeScaling);
-  ALGORITHM_PARAMETER_DECL(VolumeMultiplier);
+  ALGORITHM_PARAMETER_DECL(VolumeSamplingRate);
```
dcwhite: not OK--you could add a comment that scaling means "sampling rate"

allywarner:
Will this be an issue when SCIRun pulls the Cleaver repo instead of having its own copy?

dcwhite:
SCIRun's algorithm file is its own. If you change parameter names on the Cleaver side, that won't matter, only positions and types in a function signature. The names in the ALGORITHM-style macros are used in SCIRun network files though, so those need to stay the same lest you break users' networks.

dcwhite:
MERGING TODAY no. 4

PR #1930, opened 2019-05-13
Widget movement

RubioJr9:
Moving the widget calculates the projection of the object per frame instead of just execution.
Zooming is disable while the moving widget.

Contributors:
@Haydelj 

Reviewers:
@dcwhite @jessdtate


jessdtate:
when using the trackpad, I can get the widget state to be frozen in the moving state without any buttons pressed.  This happens when I start to scroll while clicking the widget.  The widget stays red after releasing the trackpad and will remain that way until a new mouse command is sent (eg, moving the cursor).  I haven't been able to replicate this with a mouse.

jessdtate:
If there is a quick fix for the freezing, get it in, otherwise let's just merge it.

dcwhite:
I couldn't replicate the freezing with an iMac and external trackpad, so let's just merge it now

RubioJr9:
I couldn't replicate it on my macbook. I think Jess has a newer trackpad than me and Dan. It might only be an issue on new macbooks.

Issue #1928, opened 2019-05-10
(Not exactly) Backface culling


dcwhite:
From discussion today (5/10). You guys can fill this one in.

Haydelj:
Implement proper double sided rendering using OpenGL to detect back faces and flip the normals.

Issue #1474, opened 2016-06-24
viewscene: get lights to move correctly from GUI input


cbrightsci:
The lights controlled from the lights tab in the config menu of the viewscene need to be fixed so the movement occurs with the movement of the dot.

dcwhite:
The angle sliders work. Still need to figure out those circle/dot widgets--do we keep them?

dcwhite:
@jessdtate Beta or post-beta?

jessdtate:
we implemented some sliders to do the same thing, but the dot widget still moves around. it is a bit confusing.  I'm fine going to the sliders only, but it is unfinished in its current state.

PR #1933, opened 2019-05-14
Double Sided Lighting Fix

Haydelj:
This resolves issue #1928 allowing us to move on to implement moveable lights. It also switches the orientation glyph to use uniform shading and fixes the vertex ordering for the sphere glyph.

Double sided lighting before:
image

Double sided lighting after:
image


Haydelj:
Added controls for the lights making issue #1474 obsolete.

PR #1937, opened 2019-05-17
Matlab python

dcwhite:
First version of basic matlab code embedding into InterfaceWithPython. Single function call with single input/output. Helper script code must be set up as a triggered event or top-level objects.
[helperPython.txt](https://github.com/SCIInstitute/SCIRun/files/3193134/helperPython.txt)

Also adds a checkbox to the IWP UI that enables force-reexecution, when you want to call matlab code multiple times without changing any SCIRun inputs.


src/Modules/Python/InterfaceWithPython.cc
```diff
@@ -130,22 +135,25 @@ void InterfaceWithPython::execute()
...
-      auto intermediate = parser.extractSpecialBlocks(code);
-      auto readyToConvert = parser.concatenateNormalBlocks(intermediate);
-      auto convertedCode = parser.convertStandardCodeBlock(readyToConvert);
+      auto convertedCode = translator_->translate(code);
       NetworkEditorPythonAPI::PythonModuleContextApiDisabler disabler;
-      PythonInterpreter::Instance().run_script(convertedCode);
+      if (convertedCode.isMatlab && !matlabInitialized_)
+      {
+        PythonInterpreter::Instance().run_string("import matlab.engine");
+        PythonInterpreter::Instance().run_string("__eng = matlab.engine.start_matlab()");
```
jessdtate: we also need a line that imports the converters:
```
sys.path.append(path_to_file)
from MatlabConversion import *
```

Issue #1394, opened 2016-04-07
RegisterWithCorrespondences crashes with volumetric data


gduffley:
Module only works with trisurfs, crashes with volume data. The module should output a transform matrix for linear registrations as well.

dcwhite:
Please provide a test network and data to replicate the crash.

dcwhite:
Is the second sentence a new feature, or a missing part of the original module?

aylakhan:
@gduffley: reminder that we need a test network and data

dcwhite:
@gduffley Any updates on this one?

gduffley:
@dcwhite It still crashes and it doesn't output a matrix. This is not a module I use for research, but they ability to use volume data and output a matrix seems like obvious features the module should be able to handle

jessdtate:
@gduffley is the volume data that you are testing a structured mesh?  any example data.

gduffley:
iirc it was any latvol. I'm not sure about other volumetric meshes

jessdtate:
because of the potential to morph the mesh, structured meshes, such as latvol, have been precluded from the module.  The way that it is implemented, it wouldn't allow implementing structured data for the linear transformations without a major rewrite.  I have added an error check that is hopefully more helpful.  The workaround is to use ConvertMeshToUnstructuredMesh.

jessdtate:
As alternative to using ConvertMeshToUnstructuredMesh, the new matrix port with the transform will also help.  The transform can be applied to a latvol if another mesh is used as the input.

jessdtate:
It may be possible to refactor the module code to be more similar to TransformMeshWithTransform for the linear transform options.  Perhaps a new issue for that?

gduffley:
The module could do the conversion automatically and output an unstructured mesh. I think any of these solutions are reasonable, and now that there is an error mesh it's fine. I reported it because you get a crash when there is no a priori reason to think the module wouldn't be compatible

PR #1947, opened 2019-05-25
Registration

### LINKED ISSUE #1909: Exposing the transformation matrix in register with correspondence module
**Is your feature request related to a problem? Please describe.**
I would like to be able visualize, or export, the transformation matrix that is created when you register the two point sets. This should be a 4 by 4 matrix. 

**Describe the solution you'd like**
There could be multiple solutions. I think the ideal solution is having another port coming out of the module that is a "matrix" type, and exports the 4 by 4 matrix. That way the matrix can be automatically exported. Another option is to expose the matrix calculated in the module and a person can either write it down or copy and paste it. 

**Describe alternatives you've considered**
Currently, I calculate the matrix in matlab and use that calculated matrix in other systems and for other operations. However, this is annoying and I am worried the transformation matrices are not entirely identical. 

**Additional context**
See attached files to use for the module (that actually do things!) I sent them to you on Slack.

jessdtate:
closes #1909 and maybe #1394.  

matrix added to registerwithcorrespondence.  added a check for structured mesh.


src/Core/Algorithms/Legacy/Fields/RegisterWithCorrespondences.cc
```diff
@@ -864,27 +946,27 @@ bool RegisterWithCorrespondencesAlgo::runP(FieldHandle input, FieldHandle Cors1,
         }
     }
 #endif
-    
-    return (true);
+    std::cout<<"transformed points"<<std::endl;
```
dcwhite: Remove or change to LOG_DEBUG

src/Core/Algorithms/Legacy/Fields/RegisterWithCorrespondences.cc
```diff
@@ -894,15 +976,18 @@ bool RegisterWithCorrespondencesAlgo::runN(FieldHandle input, FieldHandle Cors1,
...
+    return nullptr;
   }
 
   VMesh* imesh = input->vmesh();
   VMesh* omesh = output->vmesh();
 
   omesh = imesh;
+  
+  DenseMatrix transArray(Eigen::MatrixXd::Identity(4,4));
+  DenseMatrixHandle transform(new DenseMatrix(transArray.matrix()));
```
dcwhite: This can be done on one line without copying:
`return boost::make_shared<DenseMatrix>(std::move(Eigen::MatrixXd::Identity(4,4)));`
If that doesnt work, this is one line that will (maybe) copy (but psha, 16 elements)
`return boost::make_shared<DenseMatrix>(Eigen::MatrixXd::Identity(4,4));`

src/Core/Algorithms/Legacy/Fields/RegisterWithCorrespondences.cc
```diff
@@ -847,8 +887,50 @@ bool RegisterWithCorrespondencesAlgo::runP(FieldHandle input, FieldHandle Cors1,
...
+  
+  printMatrix(trans1,"trans1");
+  printMatrix(trans2,"trans2");
+  printMatrix(scale1,"scale1");
+  printMatrix(scale2,"scale2");
+  
+  
+  DenseMatrix transform;
+  transform = trans2*rotation*scale1*trans1;
+  DenseMatrixHandle trans_out(new DenseMatrix(transform.matrix()));
```
dcwhite: Should be able to create and do the math on the same line without copying:
`auto trans_out = boost::make_shared<DenseMatrix>(trans2*rotation*scale1*trans1);`

src/Core/Algorithms/Legacy/Fields/RegisterWithCorrespondences.cc
```diff
@@ -847,8 +887,50 @@ bool RegisterWithCorrespondencesAlgo::runP(FieldHandle input, FieldHandle Cors1,
...
+  DenseMatrix trans1(Eigen::MatrixXd::Identity(4,4));
+  DenseMatrix trans2(Eigen::MatrixXd::Identity(4,4));
+  DenseMatrix scale1(Eigen::MatrixXd::Identity(4,4));
+  DenseMatrix scale2(Eigen::MatrixXd::Identity(4,4));
+  double *r_data = rotation.data();
+  
+  for (int p = 0; p < coefs.size(); p++)
+  {
+    r_data[p] = coefs[p];
+  }
```
dcwhite: If you must use the data pointer, at least use a copy function to replace the loop:
`std::copy(coefs.begin(), coefs.end(), r_data);`

src/Core/Algorithms/Legacy/Fields/RegisterWithCorrespondences.cc
```diff
@@ -580,15 +587,48 @@ bool RegisterWithCorrespondencesAlgo::runA(FieldHandle input, FieldHandle Cors1,
...
+  
+  DenseMatrix rotation(Eigen::MatrixXd::Identity(4,4));
+  DenseMatrix trans1(Eigen::MatrixXd::Identity(4,4));
+  DenseMatrix trans2(Eigen::MatrixXd::Identity(4,4));
+  double *r_data = rotation.data();
+  
+  for (int p = 0; p < coefs.size(); p++)
+  {
+    r_data[p] = coefs[p];
+  }
```
dcwhite: Same as below code with `std::copy`

src/Core/Algorithms/Legacy/Fields/RegisterWithCorrespondences.cc
```diff
@@ -580,15 +587,48 @@ bool RegisterWithCorrespondencesAlgo::runA(FieldHandle input, FieldHandle Cors1,
...
+  
+  trans2(0,3) = sumx;
+  trans2(1,3) = sumy;
+  trans2(2,3) = sumz;
+  
+
+  
+  DenseMatrix transform;
+  transform = trans2*rotation*trans1;
+  DenseMatrixHandle trans_out(new DenseMatrix(transform.matrix()));
```
dcwhite: Same as code below with math and creation in one statement

Issue #1368, opened 2016-03-25
new ViewScene features & clipping plane performance


moritzdannhauer:
Scale Bar
- [x] I was wondering if you change Length, Height and Num Ticks, should it change the scaling of the visualization or is it supposed to be independent of it 
- [x] GUI items: Unit, Multiplier and Line Width not functioning

Clipping
- [ ] pulling the UI sliders (X: ,Y: ,Z:, D:) is really slow and my machine is paging since SCIRun is taking up 30GB of RAM visualizing a 18 MB mesh

moritzdannhauer:
The issue of using a clipping plane (huge memory allocation) is not resolved and unfortunately requested while I was demo'ing SCIRun5. Is it possible to give this a higher priority of being fixed?

dcwhite:
It will probably be next year, if we get more graphics developer time.

dcwhite:
@jessdtate @Haydelj I'll start this by trying to fix the bugs showing up in Debug mode related to clipping planes. That will point me to the memory usage. But it will also be waiting on the SRInterface refactoring.

dcwhite:
After #.... is merged, I'll debug this again.

PR #1948, opened 2019-06-05
Clipping planes

Haydelj:
This fixes the clipping plane and scale bar performance issues by optimizing newGeomentryValue for those cases resolving issue #1368. It also adds a special orientation glyph shader, adds fog and multiple light support for clipping plane visualization, and fixes the clipping plane size issue.


texput.log
```diff
@@ -0,0 +1,18 @@
+This is XeTeX, Version 3.14159265-2.6-0.99999 (TeX Live 2018) (preloaded format=xetex 2018.4.16)  3 JUN 2019 12:36
```
dcwhite: Probably want to remove this file...

src/Interface/Modules/Render/ViewScene.h
```diff
@@ -74,7 +74,7 @@ namespace SCIRun {
 
     protected Q_SLOTS:
       //---------------- New Geometry --------------------------------------------------------------
-      void newGeometryValue();
+      void newGeometryValue(bool force = false);
```
dcwhite: This is a first review that will cover style issues. I haven't run the code yet. 

For `newGeometryValue` changes, we already know it's an overly complicated function. Adding a parameter with default value just makes it more complicated. The trick to smoothing this out is as follows:
- [ ] Remove the default value `= false`
- [ ] Rename the parameter from `force` to be as explicit as possible--in this case something like `forceImmediateGarbageCollection` 
- [ ] Add an overloaded `newGeometryValue` with no parameters that just calls the parameter one, passing false. Then you don't have to change exisiting usages like before, and the implementation stays in the .cc file

dcwhite: Then, perhaps, you can have a newly-named function that calls `newGeometryValue` passing true, like `newGeometryValueWithCleanup` or something. See if there is a name that makes sense there. Reading the client code will be a lot easier than having to remember what the magic boolean does.

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -1788,31 +1802,31 @@ void ViewSceneDialog::setAmbientValue(double value)
 {
   state_->setValue(Modules::Render::ViewScene::Ambient, value);
   setMaterialFactor(SRInterface::MAT_AMBIENT, value);
-  newGeometryValue();
+  newGeometryValue(true);
```
dcwhite: Thus this type of call would be `forceNewGeometry` or `newGeometryValueWithCleanup` or whatever

src/Interface/Modules/Render/ES/SRInterface.h
```diff
@@ -157,6 +157,7 @@ namespace SCIRun {
       void handleGeomObject(Graphics::Datatypes::GeometryHandle object, int port);
       // Remove all SCIRun 5 objects.
       void removeAllGeomObjects();
+      bool hasObject(std::string object);
```
dcwhite: Change parameter to `const std::string&` instead of pass by value

src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -1221,6 +1221,16 @@ namespace SCIRun {
       mCore.renormalize(true);
     }
 
+    //----------------------------------------------------------------------------------------------
+    bool SRInterface::hasObject(std::string object)
+    {
+      for (auto it = mSRObjects.begin(); it != mSRObjects.end(); ++it)
```
dcwhite: If you want, the more modern way to write this is:
`return std::find_if(mSRObjects.begin(), mSRObjects.end(), [&object](const Object& obj) { return obj.mName == object; }) != mSRObjects.end();`
where `Object` is the type of element in `mSRObjects` (which I forgot)

src/Interface/Modules/Render/ES/SRCamera.cc
```diff
@@ -141,7 +141,9 @@ namespace SCIRun {
     //----------------------------------------------------------------------------------------------
     void SRCamera::setClippingPlanes()
     {
-      mRadius = (mSceneBBox.get_max() - mSceneBBox.get_min()).length() / 2.0f;
+      //we could divide by 2.0 for tightest bound but here we divide by 1.0 to accomidate
+      //for the clipping plane visualization
+      mRadius = (mSceneBBox.get_max() - mSceneBBox.get_min()).length() / 1.0f;
```
dcwhite: What does the divide by one do? Simply convert to float without changing the value? To do that use a `static_cast<float>`--the divide by 1 just looks weird

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -585,8 +585,22 @@ QColor ViewSceneDialog::checkColorSetting(std::string& rgb, QColor defaultColor)
...
+}
+
+void ViewSceneDialog::updateModifiedGeometries()
+{
+  //if we are looking for a new geoetry the ID will have changed therefore we can find the
+  //geometrys that have changed and only remove those
+  newGeometryValue(false);
+}
+
```
dcwhite: I like the new names! Good comment use too.

Haydelj:
I just found a bug with glyph shading which might be related to the changes I made I'm going to investigate further.

Haydelj:
Looks like it exists on master as well

Haydelj:
It was just a shader that was never updated for multi light support

dcwhite:
- [ ] Add color scheme for the 6 plane widgets. Modify label color and glyph color to match. Later we can make this cusomizable.

dcwhite:
6coloredPlanes

Issue #991, opened 2015-05-26
Module porting: ShowAndEditDipoles


dcwhite:

aylakhan:
Module documentation also missing.

dcwhite:
Let's set up a meeting about widgets soon.

dcwhite:
For next testing lunch.

PR #1958, opened 2019-06-28
Show And Edit Dipoles Module and Fuzzy Search

RubioJr9:
Issues:
#991 

Reviewers:
@dcwhite @jessdtate 

ShowAndEditDipoles module is complete. Added option to move all dipole positions together.
Created Geometry Composite which acts like a CompositeWidget but you can add both widgets and regular geometry. Added fuzzy search to filter options in the Module Selector.


src/Graphics/Datatypes/GeometryImpl.h
```diff
@@ -287,7 +287,24 @@ namespace SCIRun {
 
       typedef boost::shared_ptr<GeometryObjectSpire> GeometryHandle;
 
+      class SCISHARE Composite : public GeometryObjectSpire
```
dcwhite: Rename class to CompositeGeometryObject

RubioJr9: Done

RubioJr9:
Made changes to fuzzy search:
-Ignores all special symbols
-Only case sensitive if you use upper case. ie. "aB" will match to "AB", but not "ab"
-Checks up to 8 permutations of search if you use spaces. ie. "mesh get" will fuzzy search both "meshget" and "getmesh"
- * character uses wildcards search instead of fuzzy search

I will add documentation for both dipoles and fuzzy search on Monday.

RubioJr9:
I got my changes in and added a couple test networks. I won't be making any more commits to this branch unless either of you have requests or we find any bugs.

dcwhite:
On my Qt5 Mac build, the selection action (red highlight->draggable) doesn't match up with the mouse click location, it seems to be translated. I will check Windows/Qt5.

RubioJr9:
> On my Qt5 Mac build, the selection action (red highlight->draggable) doesn't match up with the mouse click location, it seems to be translated. I will check Windows/Qt5.

Are you seeing this with other widgets as well?

dcwhite:
Looks like yes. Bounding box same problem. Separate issue then

src/Interface/Application/TreeViewCollaborators.cc
```diff
@@ -78,9 +80,81 @@ void HideItemsNotMatchingString::operator()(QTreeWidgetItem* item)
...
+{
+  int patternIndex = 0;
+
+  for(int t = 0; t < text.length(); t++)
+  {
+    bool isUpperChar = pattern[patternIndex] < 97;
+
+    // Counts as a match if letter found. Only case sensitive if search char is upper case
+    if((isUpperChar && (text[t] == pattern[patternIndex]))
+       || (!isUpperChar && (std::tolower(text[t]) == std::tolower(pattern[patternIndex]))))
```
dcwhite: On Windows `tolower` needs the `#include <cctype>` header.

dcwhite:
Looks like Windows/Qt5 is OK, at least with the "OtherWidget" network. The other one with two VSes, shows nothing when executed.

RubioJr9:
> On my Qt5 Mac build, the selection action (red highlight->draggable) doesn't match up with the mouse click location, it seems to be translated. I will check Windows/Qt5.

I'm not seeing this issue on my machine. I built with qt 5.13.0 on OSX 10.13.6

src/Interface/Application/TreeViewCollaborators.cc
RubioJr9: It's added now.

dcwhite:
Windows/Qt5 builds fine now, and I had to re-add ViewScenes to get the test network working mentioned above. That's probably a separate issue too, so I think this one can go. @jessdtate needs to approve.

dcwhite:
@RubioJr9 Real build errors coming from Travis, please investigate and fix: https://travis-ci.org/SCIInstitute/SCIRun/builds/559064365?utm_source=github_status&utm_medium=notification

src/Modules/Legacy/Visualization/ShowAndEditDipoles.cc
```diff
@@ -212,7 +212,7 @@ void ShowAndEditDipoles::loadData()
   }
 
   bool zeroVector = false;
-  for(int i = 0; i < scale_.size(); i++)
+  for(uint i = 0; i < scale_.size(); i++)
```
dcwhite: Can you use standard size_t here? uint is a compiler-specific typedef

RubioJr9: Good point

RubioJr9: I changed all ints to size_t since they were all positive and many were used to access arrays we were using size_t for.

dcwhite:
@jessdtate Build is green, you can approve again now.

PR #1963, opened 2019-07-09
Red ViewScene Part 2

Haydelj:
Fixed the issue causing the new red viewScene by preventing the garbage collector from running when there are unfulfilled geometry promises. Also fixes a segfault in SRInterface and removes unused assets from assetbootstrap.


src/Interface/Modules/Render/ES/Core.cc
```diff
@@ -151,7 +151,19 @@ void ESCore::execute(double currentTime, double constantFrameTime)
...
+      if(comp.second->getNumComponents() > 0)
+      {
+        if(mComponentIDNameMap.find(comp.first) != mComponentIDNameMap.end() &&
+           mComponentIDNameMap.at(comp.first) == "ren:GeomPromise")
+        {
+          run = false;
+          break;
+        }
+      }
+    if(run) runCompleteGC();
```
dcwhite: Two issues here: there's already a `if (runGC)` check, so this is like a `if (reallyActuallyRunGC)` check. Can the checks be combined (issue 1), and extracted into one function (issue 2)? Also, a typo in the cout line below, which should probably be a LOG_DEBUG or LOG_TRACE call instead.

src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -834,6 +834,9 @@ namespace SCIRun {
 
       DEBUG_LOG_LINE_INFO
 
+      if(!mCore.getStaticComponent<ren::StaticVBOMan>() ||
```
dcwhite: This check should be extracted into a function as well. It's good practice for refactoring anyway; don't add any more complexity than is needed (less to refactor later).

Haydelj: I feel as though this one is a bit short for its own function. I did, however, notice that we cast a shared pointer to a weak pointer only to lock it right after am I mistaken in thinking this is pointless?

src/Interface/Modules/Render/ES/Core.cc
```diff
@@ -110,6 +110,18 @@ std::string ESCore::toString(std::string prefix) const
   return output;
 }
 
+bool ESCore::hasGeomPromise()
```
dcwhite:
```suggestion
bool ESCore::hasGeomPromise() const
```

src/Interface/Modules/Render/ES/Core.cc
```diff
@@ -110,6 +110,18 @@ std::string ESCore::toString(std::string prefix) const
...
+{
+  for(auto& comp : mComponents)
```
dcwhite:
```suggestion
  for(auto& comp : mComponents)
  { // confusing multi-line for block should have braces
```

src/Interface/Modules/Render/ES/Core.cc
```diff
@@ -110,6 +110,18 @@ std::string ESCore::toString(std::string prefix) const
...
+    if(comp.second->getNumComponents() > 0)
+    {
+      if(mComponentIDNameMap.find(comp.first) != mComponentIDNameMap.end() &&
+         mComponentIDNameMap.at(comp.first) == "ren:GeomPromise")
+        return true;
+    }
```
dcwhite:
```suggestion
    }
  }
```

src/Interface/Modules/Render/ES/Core.h
```diff
@@ -49,6 +49,7 @@ namespace Render {
   void runGCOnNextExecution(){runGC = true;}
 
 private:
+  bool hasGeomPromise();
```
dcwhite:
```suggestion
  bool hasGeomPromise() const;
```

PR #1967, opened 2019-07-22
Show Field Tests

Haydelj:
On going


dcwhite:
@Haydelj This doesn't build on Windows since ShowField doesn't use the newfangled way of declaring algorithm name objects (`ALGORITHM_PARAMETER_DECL`). I'll make the fix locally and we can merge it separately, since you mentioned it would be painful to add more commits to this one.

PR #1969, opened 2019-07-25
Glyphs

RubioJr9:
-Added boxes, rings(toruses), and superellipsoids
-Added option to uniformly scale secondary glyph values
-Added option to render bases of arrows and cones and scale now scales secondary glyph value
-Comets now use a (almost) hemisphere, reducing memory
-Comets are now moved back so the sphere is at the base of the vector
-Disks have radius and width swapped so they are flat instead of long cylinders
-Fixed scaling on ellipsoids
-Fixed normals on order 2 ellipsoids and boxes
-Added error handling for undefined behavior in ShowFieldGlyphsPortHandler


RubioJr9:
Found an issue from SCIRun 4 with negative eigenvalues on order 2 tensors. Will fix and update branch

RubioJr9:
Changes are in. Also added a new test network

dcwhite:
@RubioJr9 Ready for conflict fixes

RubioJr9:
Don't merge yet. I'm seeing a scaling issue with boxes

RubioJr9:
I'm done making changes

jessdtate:
disks primary and secondary attributes are flipped
comet location should be the center of the sphere

RubioJr9:
Ready for testing. I edited the original comment to have all the changes

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -523,228 +567,632 @@ void GlyphGeom::generateSphere(const Point& center, double radius, int resolutio
...
+
+  // Rotate and translate points
+  Vector p1 = Vector(trans * Point(eigvals * Vector(-1.0, 1.0, 1.0)));
+  Vector p2 = Vector(trans * Point(eigvals * Vector(-1.0, 1.0, -1.0)));
+  Vector p3 = Vector(trans * Point(eigvals * Vector(1.0, 1.0, 1.0)));
+  Vector p4 = Vector(trans * Point(eigvals * Vector(1.0, 1.0, -1.0)));
+  Vector p5 = Vector(trans * Point(eigvals * Vector(-1.0, -1.0, 1.0)));
+  Vector p6 = Vector(trans * Point(eigvals * Vector(-1.0, -1.0, -1.0)));
+  Vector p7 = Vector(trans * Point(eigvals * Vector(1.0, -1.0, 1.0)));
+  Vector p8 = Vector(trans * Point(eigvals * Vector(1.0, -1.0, -1.0)));
```
dcwhite: This is a good block to extract into a class/function. The 8 constant vectors then could be declared static constexpr in the new class, saving time and memory, and you'd reduce duplication. Calling code would look something like
```
BoxBuilder bb;
auto box = bb.makeSideVectors(trans, eigvals);
//....
//changed to 0-based
generateBoxSide(box[6], box[7], box[2], box[3], x_vec, node_color);
...
```

dcwhite: I'll leave it to you to figure out what type the `auto` should be :)

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -523,228 +567,632 @@ void GlyphGeom::generateSphere(const Point& center, double radius, int resolutio
...
+
+  // Rotate norms
+  Vector x_vec = rotate * Vector(1, 0, 0);
+  Vector y_vec = rotate * Vector(0, 1, 0);
+  Vector z_vec = rotate * Vector(0, 0, 1);
```
dcwhite: This block can be refactored into the `BoxBuilder` class as well.
```
auto xyz = bb.makeNorms(rotate);
//...
generateBoxSide(box[6], box[7], box[2], box[3], xyz[0], node_color);
```

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -523,228 +567,632 @@ void GlyphGeom::generateSphere(const Point& center, double radius, int resolutio
...
+  }
+  else
+  {
+    return (double)(pow(e, x));
+  }
+}
+
+void GlyphGeom::generateSuperEllipsoid(const Point& center, Tensor& t, double scale, int resolution, const ColorRGB& color, bool normalize, double emphasis)
+{
+  double zeroThreshold = 0.000001;
```
```suggestion
  static const double zeroThreshold = 0.000001;
```

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -523,228 +567,632 @@ void GlyphGeom::generateSphere(const Point& center, double radius, int resolutio
...
+      eigvectors[1] = zero_norm;
+    }
+    else if(eig_z_0)
+    {
+      zero_norm = Cross(eigvectors[0], eigvectors[1]);
+      eigvectors[2] = zero_norm;
+    }
+  }
+
+  Transform rotate(Point(0.0, 0.0, 0.0), eigvectors[0], eigvectors[1], eigvectors[2]);
```
dcwhite: This is duplicated above--make a new function `transformFromEigenVectors` or something

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -523,228 +567,632 @@ void GlyphGeom::generateSphere(const Point& center, double radius, int resolutio
...
+      indices_.push_back(3 + offset);
+    }
+  }
+  for(int jj = 0; jj < 6; jj++) indices_.pop_back();
+}
+
+void GlyphGeom::generateTorus(const Point& p1, const Point& p2, double major_radius, double minor_radius,
+                              int resolution, const ColorRGB& color)
+{
+  std::cout << "minor, major rad: " << minor_radius << ", " << major_radius << std::endl;
```
dcwhite: Remove

src/Interface/Modules/Visualization/ShowFieldGlyphsDialog.cc
```diff
@@ -54,6 +54,7 @@ ShowFieldGlyphsDialog::ShowFieldGlyphsDialog(const std::string& name, ModuleStat
   createExecuteInteractivelyToggleAction();
 
   connect(defaultMeshColorButton_, SIGNAL(clicked()), this, SLOT(assignDefaultMeshColor()));
+  // connect(tensorSuperquadricsEmphasisSlider_, SIGNAL(valueChanged(int)), parent, SLOT(setSuperquadricEmphasis(int)));
```
dcwhite: Remove unless it's coming soon

src/Modules/Visualization/ShowFieldGlyphs.cc
```diff
@@ -734,8 +778,9 @@ void GlyphBuilder::renderTensors(
       }
 
     int neg_eigval_count = 0;
-
     int tensorcount = 0;
+    double vectorThreshold = 0.001;
```
dcwhite:
```suggestion
    static const double vectorThreshold = 0.001;
```

src/Modules/Visualization/ShowFieldGlyphs.cc
```diff
@@ -734,8 +778,9 @@ void GlyphBuilder::renderTensors(
...
+    double pointThreshold = 0.01;
```
dcwhite:
```suggestion
    static const double pointThreshold = 0.01;
```

src/Graphics/Glyphs/GlyphGeom.cc
RubioJr9: `Vector p5 = Vector(trans * Point(-eigvals.x(), -eigvals.y(), eigvals.z());`
What if I did this instead? This gets rid of instantiating the vector and doing the extra multiplication operation without needing a new class

dcwhite: It's still a high degree of duplication

dcwhite: Anytime there's a long list of similar-looking variables ending in 1,2,3,4,..., that is a smell and should be replaced with a function returning an array

src/Graphics/Glyphs/GlyphGeom.cc
RubioJr9: I changed it to get the vectors from the transform and avoided the multiplication entirely.

src/Graphics/Glyphs/GlyphGeom.cc
RubioJr9: I'm using arrays for for both points and normals now. I refactored it so the points are generated in a loop because we previously had the permutations hard coded to get the 8 points. Is there any issues with this method?

`std::vector<Vector> points;
  for(int x = -1; x < 2; x+=2)
  {
    for(int y = -1; y < 2; y+=2)
    {
      for(int z = -1; z < 2; z+=2)
      {
        Point translated_point = trans * Point(x * eigvals.x(), y * eigvals.y(), z * eigvals.z());
        points.push_back(Vector(translated_point));
      }
    }
  }`

dcwhite: A new-style better way to write each loop is:
```
for (int x : {-1, 1})
  for (int y : {-1, 1})
    for (int z : {-1, 1})
{
 ...
}
```

dcwhite: And I'd still at the very least extract this into a separate function.

dcwhite: And you can make the body a one-liner:
```
        points.emplace_back(trans * Point(x * eigvals.x(), y * eigvals.y(), z * eigvals.z()));
```

RubioJr9: I like that loop style much better. Why should we separate point generation into another function? I don't see any place we could reuse that code. Every other glyph has their points being generated in the same function and boxes have a much simpler implementation.

RubioJr9: I put it into a separate function. I decided against separating point generation for other glyphs because we can't without duplicating the nested for loop. It's too intertwined with generating normals unlike boxes. Plus we would have to pass along tons of parameters.

src/Graphics/Glyphs/GlyphGeom.h
```diff
@@ -93,11 +107,17 @@ namespace SCIRun {
...
-      void generateCone(const  Core::Geometry::Point& p1, const  Core::Geometry::Point& p2, double radius, int resolution, bool renderBase, const Core::Datatypes::ColorRGB& color1, const Core::Datatypes::ColorRGB& color2);
-      void generateDisk(const  Core::Geometry::Point& p1, const  Core::Geometry::Point& p2, double radius1, double radius2, int resolution, const Core::Datatypes::ColorRGB& color1, const Core::Datatypes::ColorRGB& color2);
       void generateSphere(const Core::Geometry::Point& center, double radius, int resolution, const Core::Datatypes::ColorRGB& color);
-      void generateBox(const Core::Geometry::Point& center, Core::Geometry::Tensor& t, double scale);
-    void generateEllipsoid(const Core::Geometry::Point& center, Core::Geometry::Tensor& t, Core::Geometry::Vector& scaled_eigenvals, int resolution, const Core::Datatypes::ColorRGB& color);
+      std::vector<Core::Geometry::Vector> generateBoxPoints(Core::Geometry::Transform& trans, Core::Geometry::Vector& eigvals);
```
dcwhite:
```suggestion
      static std::vector<Core::Geometry::Vector> generateBoxPoints(const Core::Geometry::Transform& trans, const Core::Geometry::Vector& eigvals);
```

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -523,228 +567,631 @@ void GlyphGeom::generateSphere(const Point& center, double radius, int resolutio
...
+    double temp = eigvals[0];
+    eigvals[0] = eigvals[1];
+    eigvals[1] = temp;
+    Vector tempVec = eigvectors[0];
+    eigvectors[0] = eigvectors[1];
+    eigvectors[1] = tempVec;
+  }
+}
+
+std::vector<Vector> GlyphGeom::generateBoxPoints(Transform& trans, Vector& eigvals)
```
```suggestion
std::vector<Vector> GlyphGeom::generateBoxPoints(const Transform& trans, const Vector& eigvals)
```

dcwhite: Almost there!

RubioJr9: Sorry, I thought I did that.

src/Modules/Visualization/ShowFieldGlyphs.cc
```diff
@@ -167,29 +177,41 @@ void GlyphBuilder::addGlyph(
...
       break;
     case RenderState::GlyphType::ARROW_GLYPH:
-      glyphs.addArrow(p1, p2, radius, ratio, resolution, node_color, node_color);
+      glyphs.addArrow(p1, p2, radius, ratio, resolution, node_color, node_color, render_base1, render_base2);
       break;
     case RenderState::GlyphType::DISK_GLYPH:
-      glyphs.addDisk(p1, p2, radius, resolution, node_color, node_color);
+    {
+      Vector dir = (p2 - p1);
+      Point new_p2 = p1 + dir * radius * 2.0;
```
jessdtate: This line isn't quite right.  The scaling of the length isn't staying consistent with the radius

src/Modules/Visualization/ShowFieldGlyphs.cc
```diff
@@ -167,29 +177,41 @@ void GlyphBuilder::addGlyph(
...
+      glyphs.addDisk(p1, new_p2, dir.length()*0.5, resolution, node_color, node_color);
       break;
+    }
     case RenderState::GlyphType::RING_GLYPH:
-      BOOST_THROW_EXCEPTION(AlgorithmInputException() << ErrorMessage("Ring Geom is not supported yet."));
+    {
+      double length = (p2 - p1).length();
+      double major_radius = length * 0.5;
+      double minor_radius = length * radius * 2.0;
```
jessdtate: same here

src/Modules/Visualization/ShowFieldGlyphs.cc
RubioJr9: Scaling on this and rings should be good now

PR #1976, opened 2019-08-12
Show field refactor

Haydelj:
Implemented texture mapping for color mapping for faces. Refactored and optimized face geometry generation. Added new shaders that are more versatile and use better naming conventions that will eventually fully replace the old ones but are currently only used by faces.


src/Modules/Visualization/ShowField.cc
```diff
@@ -430,6 +430,96 @@ void GeometryBuilder::renderFaces(
 
 
 
+inline static void writeAtributeToVBO(const Point& point, spire::VarBuffer* vboBuffer)
+{
+  vboBuffer->write(static_cast<float>(point.x()));
```
dcwhite: These could be simplified with a template as well:
```
template <typename T>
void writeFloats(spire::VarBuffer* vboBuffer, std::initializer_list<T> ts)
{
  for (const T& t : ts)
   vboBuffer->write(static_cast<float>(t);
}
```
then each of these can pass a list, avoiding duplication:
```
writeFloats(vboBuffer, {point.x(), point.y(), point.z()});
```

src/Modules/Visualization/ShowField.cc
```diff
@@ -430,6 +430,96 @@ void GeometryBuilder::renderFaces(
...
+  vboBuffer->write(static_cast<float>(point.y()));
+  vboBuffer->write(static_cast<float>(point.z()));
+}
+
+inline static void writeAtributeToVBO(const Vector& vector, spire::VarBuffer* vboBuffer)
+{
+  vboBuffer->write(static_cast<float>(vector.x()));
```
dcwhite: See above

src/Modules/Visualization/ShowField.cc
```diff
@@ -430,6 +430,96 @@ void GeometryBuilder::renderFaces(
...
+  vboBuffer->write(static_cast<float>(vector.y()));
+  vboBuffer->write(static_cast<float>(vector.z()));
+}
+
+inline static void writeAtributeToVBO(const ColorRGB& color, spire::VarBuffer* vboBuffer)
+{
+  vboBuffer->write(static_cast<float>(color.r()));
```
dcwhite: See above

src/Modules/Visualization/ShowField.cc
```diff
@@ -430,6 +430,96 @@ void GeometryBuilder::renderFaces(
...
+  vboBuffer->write(static_cast<float>(color.g()));
+  vboBuffer->write(static_cast<float>(color.b()));
+  vboBuffer->write(static_cast<float>(1.0f));
+}
+
+inline static void writeAtributeToVBO(const glm::vec2& coords, spire::VarBuffer* vboBuffer)
+{
+  vboBuffer->write(static_cast<float>(coords.x));
```
dcwhite: See above--although, the function I wrote ought to be a member function of VarBuffer

src/Modules/Visualization/ShowField.cc
```diff
@@ -430,6 +430,96 @@ void GeometryBuilder::renderFaces(
...
+}
+
+template<typename ... Params>
+static void writeTri(
+  spire::VarBuffer* vboBuffer,
+  spire::VarBuffer* iboBuffer,
+  uint32_t& iboIndex,
+  const Params& ... params)
+{
+  for(int i = 0; i < 3; ++i)
```
dcwhite: I'll have more to say on this after I build it and look at the usage

src/Modules/Visualization/ShowField.cc
```diff
@@ -448,754 +538,293 @@ void GeometryBuilder::renderFacesLinear(
...
-      else if (fld->is_vector())
-      {
-        fld->get_value(vvals[0], *fiter);
-        face_colors[0] = map->valueToColor(vvals[0]);
-      }
-      else if (fld->is_tensor())
-      {
-        fld->get_value(tvals[0], *fiter);
-        face_colors[0] = map->valueToColor(tvals[0]);
+        case 0b000: writeTri(vboBuffer, iboBuffer, iboIndex, points); break;
```
dcwhite: Please use `enum` labels for these, we're not writing assembly code here

src/Modules/Visualization/ShowField.cc
```diff
@@ -448,754 +538,293 @@ void GeometryBuilder::renderFacesLinear(
...
-  // Accessing the pointers like this is contrived. We only do this for speed since we will be
-  // using the pointers in a tight inner loop.
-  auto iboBuffer = iboBufferSPtr.get();
-  auto vboBuffer = vboBufferSPtr.get();
+  std::vector<Point> points(numNodes);
+  std::vector<Vector> normals(numNodes);
+  std::vector<glm::vec2> textureCoords(numNodes);
+  std::vector<double> svals(numNodes);
+  std::vector<Vector> vvals(numNodes);
+  std::vector<Tensor> tvals(numNodes);
```
dcwhite: Would the below be faster/better memory/cache usage if this were allocated as one vector of structs, one struct instance for each node?

src/Modules/Visualization/ShowField.cc
```diff
@@ -448,754 +538,293 @@ void GeometryBuilder::renderFacesLinear(
...
-  size_t vboSize = static_cast<size_t>(mesh->num_faces() * sizeof(float) * nodes.size() * numAttributes);
-
-  // Construct VBO and IBO that will be used to render the faces. Once again, IBOs are not strictly
-  // needed. But, we may be able to optimize this code somewhat.
-  // todo Switch to unique_ptrs and move semantics.
-  std::shared_ptr<spire::VarBuffer> iboBufferSPtr(new spire::VarBuffer(iboSize));
-  std::shared_ptr<spire::VarBuffer> vboBufferSPtr(new spire::VarBuffer(vboSize));
+  int numNodes = nodes.size();
+  bool useQuads = (numNodes == 4);
+  int writeCase = ((int)useQuads << 2) | ((int)useNormals << 1) | ((int)useColorMap);
```
dcwhite: This should definitely be a function returning a typed enum value with the correct label that makes the switch readable

src/Modules/Visualization/ShowField.cc
```diff
@@ -448,754 +538,293 @@ void GeometryBuilder::renderFacesLinear(
...
+  size_t passNumber = 0;
+  size_t facesLeft = mesh->num_faces();
 
-  while (fiter != fiterEnd)
+  while(facesLeft > 0)
   {
-    interruptible->checkForInterruption();
+    const static size_t maxFacesPerPass = 1 << 24;
+    int facesLeftInThisPass = std::min(facesLeft,maxFacesPerPass);
+    facesLeft -= facesLeftInThisPass;
```
dcwhite: This logic should be extracted into a `partitionFaces` function, which then can pass the correct subset of faces to the actual renderFaces call. It will most likely change in the future

src/Modules/Visualization/ShowField.cc
```diff
@@ -448,754 +538,293 @@ void GeometryBuilder::renderFacesLinear(
...
+            normals[i] = -normals[i];
+      }
 
-      if (fld->is_scalar())
+      if(useColorMap)
       {
-        fld->get_value(svals[0], cells[0]);
-
-        if (cells.size() > 1)
+        // Element data (Cells) so two sided faces.
```
dcwhite: Part of this or a near-future refactoring should be to extract the coloring functions separately from the spatial functions

src/Modules/Visualization/ShowField.cc
```diff
@@ -430,6 +430,96 @@ void GeometryBuilder::renderFaces(
...
+  writeIndexToIBO(iboIndex + 2, iboBuffer);
+  writeIndexToIBO(iboIndex + 2, iboBuffer);
+  writeIndexToIBO(iboIndex + 3, iboBuffer);
+  writeIndexToIBO(iboIndex + 0, iboBuffer);
+  iboIndex += 4;
+}
+
+
+
+static float valueToFloat(double scalar)
```
dcwhite: These color conversions need to be exposed and better documented. Could you extract a class, say, `ShowFieldDataColorConverter` which contains these 3 functions?

dcwhite:
This is part 1 of (at least) 3 of my review

src/Modules/Visualization/ShowField.cc
Haydelj: I actually found a better way to do it so those are unused now. I will remove them

src/Modules/Visualization/ShowField.cc
Haydelj: If I had it my way...

dcwhite: You can put all the assembly you want in the shader files, i ignore those...

src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -1295,25 +1226,20 @@ namespace SCIRun {
     //----------------------------------------------------------------------------------------------
     void SRInterface::addTextToEntity(uint64_t entityID, const SpireText& text)
     {
-      if (text.name == "")
-        return;
-
-       //texture man
+      if (text.name == "") return;
```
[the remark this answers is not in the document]
Haydelj: Out of curiosity why do it this way? Is it faster?

src/Graphics/Datatypes/GeometryImpl.h
```diff
@@ -145,6 +145,25 @@ namespace SCIRun {
         std::vector<uint8_t>                  bitmap;
       };
 
+      struct SpireTexture2D
```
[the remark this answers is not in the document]
Haydelj: As is there is a multitude of things that we need to do to actually add a texture to an entity, therefore, I feel more comfortable just letting the function I wrote in SRInterface handle all of that.

src/Modules/Visualization/ShowField.cc
Haydelj: Hmm, interesting point. For svals, vvals, and tvals we should only ever use one of them in the loop so these are almost certainly better left apart.  For the textureCoords, normals, and points it's a bit more interesting. Having them interleaved in memory could improve performance for copying them to the vbo however it could hurt performance for copying them from the vmesh since the writes will no longer be sequential. This from my experience can be even more important than keeping reads sequential. The only way to know for sure is to try though

src/Interface/Modules/Render/ES/SRInterface.cc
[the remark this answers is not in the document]
dcwhite: It ought to be a bit faster, since it's just a member function call versus constructing a new string and calling an operator

src/Modules/Visualization/ShowField.cc
dcwhite: (1) Since  
> svals, vvals, and tvals we should only ever use one of them in the loop
you're doing 3x the allocation here than necessary? That needs to be fixed then.
(2) You're right about testing the memory layout for the required parts. Profile it and see

src/Modules/Visualization/ShowField.cc
```diff
@@ -448,754 +538,293 @@ void GeometryBuilder::renderFacesLinear(
...
-  size_t iboSize = static_cast<size_t>(mesh->num_faces() * sizeof(uint32_t) * (nodes.size() - 2) * 3);
-  size_t vboSize = static_cast<size_t>(mesh->num_faces() * sizeof(float) * nodes.size() * numAttributes);
-
-  // Construct VBO and IBO that will be used to render the faces. Once again, IBOs are not strictly
-  // needed. But, we may be able to optimize this code somewhat.
-  // todo Switch to unique_ptrs and move semantics.
-  std::shared_ptr<spire::VarBuffer> iboBufferSPtr(new spire::VarBuffer(iboSize));
-  std::shared_ptr<spire::VarBuffer> vboBufferSPtr(new spire::VarBuffer(vboSize));
+  int numNodes = nodes.size();
+  bool useQuads = (numNodes == 4);
```
dcwhite: Is all the quad logic truly for only a 4-node mesh? Would it make much difference if tris were  used for such a small case?

src/Modules/Visualization/ShowField.cc
```diff
@@ -430,6 +430,96 @@ void GeometryBuilder::renderFaces(
...
+{
+  iboBuffer->write(index);
+}
+
+template<typename ... Params>
+static void writeTri(
+  spire::VarBuffer* vboBuffer,
+  spire::VarBuffer* iboBuffer,
+  uint32_t& iboIndex,
+  const Params& ... params)
```
dcwhite: I applaud the template attempt but there's an easier+clearer way to handle the optional arguments.

The points vector is always passed in. So make that a normal parameter first.

Then the question is how to add 0,1, or 2 of normals and textureCoords. Just make those boost::optional<vector>s, and pass in `{}` for the cases where you don't need them. This function then looks like
```
static void writeTri(
  spire::VarBuffer* vboBuffer,
  spire::VarBuffer* iboBuffer,
  uint32_t& iboIndex,
const std::vector<Point>& points,
 boost::optional<std::vector<Vector>> normals,
  boost::optional<std::vector<glm::vec2>> textureCoords)
{
for 0->2
   write point
   if (normals)
      write normal
   if (textureCoords)
      write Texture
```
with no template. The switch looks like
```
        case JUST_POINTS: writeTri(vboBuffer, iboBuffer, iboIndex, points, {}, {}); break;
        case WITH_COLOR: writeTri(vboBuffer, iboBuffer, iboIndex, points, {}, textureCoords); break;
        case WITH_NORMALS: writeTri(vboBuffer, iboBuffer, iboIndex, points, normals, {}); break;
        case WITH_COLOR_AND_NORMALS: writeTri(vboBuffer, iboBuffer, iboIndex, points, normals, textureCoords); break;
```

Easier to read, yes? No need for "janky" voided initializer_list expanded parameter pack.

dcwhite: Just make sure the vectors aren't copied when passing them as boost::optional parameters. Doc here: https://www.boost.org/doc/libs/1_63_0/libs/optional/doc/html/boost_optional/tutorial/performance_considerations.html

Reading that, you can use optional references:
`boost::optional<const std::vector<Vector>&> normals,`
etc

src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -1295,25 +1226,20 @@ namespace SCIRun {
...
       ss << "FontTexture:" << entityID << text.name << text.width << text.height;
       std::string assetName = ss.str();
 
       ren::Texture texture;
 
-      spire::CerealHeap<ren::Texture>* contTex =
-        mCore.getOrCreateComponentContainer<ren::Texture>();
-      std::pair<const ren::Texture*, size_t> component =
-        contTex->getComponent(entityID);
+      spire::CerealHeap<ren::Texture>* contTex = mCore.getOrCreateComponentContainer<ren::Texture>();
```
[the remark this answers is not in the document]
Haydelj: I think I prefer these types to be explicit. Unless you have a strong preference I'd like to keep these how they are.

src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -1324,6 +1250,33 @@ namespace SCIRun {
       mCore.addComponent(entityID, texture);
     }
 
+    //----------------------------------------------------------------------------------------------
+    void SRInterface::addTextureToEntity(uint64_t entityID, const SpireTexture2D& texture)
+    {
+      if (texture.name == "") return;
```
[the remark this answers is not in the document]
Haydelj: see above

dcwhite:
@Haydelj That's OK about auto usage--those are just suggestions

src/Modules/Visualization/ShowField.cc
Haydelj: This would require some fairly large code changes to test and I'm reasonably certain it won't be faster so I'm going to go ahead and mark this resolved

src/Modules/Visualization/ShowField.cc
Haydelj: I don't follow

src/Modules/Visualization/ShowField.cc
dcwhite: What about comment (1)?

src/Modules/Visualization/ShowField.cc
dcwhite: The structure is 
```while (stuffToDo())
{ doStuff(); }
```
I'm saying extract each of those two parts as a separate function since they are independent.

dcwhite: Independent from each other, and reusable for nodes/edges partitioning.

dcwhite:
@Haydelj Also, any performance improvements on the BruteForceTest from this change?

src/Modules/Visualization/ShowField.cc
dcwhite: Part (2) can wait but it should be easy to extract an object that encapsulates the set of vectors.

src/Modules/Visualization/ShowField.cc
Haydelj: Won't that kinda have to look like this?

  inline int getWriteCase(bool writeQuads, bool writeNormals, bool writeTexCoords)
  {
    if(writeQuads)
    {
      if(writeNormals)
      {
        if(writeTexCoords)
        {
          return QUAD_NORMALS_TEXCOORDS;
        }
        else
        {
          return QUAD_NORMALS;
        }
      }
      else
      {
        if(writeTexCoords)
        {
          return QUAD_TEXCOORDS;
        }
        else
        {
          return QUAD;
        }
      }
    }
    else
    {
      if(writeNormals)
      {
        if(writeTexCoords)
        {
          return TRI_NORMALS_TEXCOORDS;
        }
        else
        {
          return TRI_NORMALS;
        }
      }
      else
      {
        if(writeTexCoords)
        {
          return TRI_TEXCOORDS;
        }
        else
        {
          return TRI;
        }
      }
    }
  }

dcwhite: No, you can set up the enum class to have the same values that the `int` has now, then just construct it:
```
EnumTypeName writeCase(bool writeQuads, bool writeNormals, bool writeTexCoords)
{
  return EnumTypeName(((int)useQuads << 2) | ((int)useNormals << 1) | ((int)useColorMap));
}
```

dcwhite: If it had to look like that do you think I'd suggest it? 😝

Haydelj: Can we talk about this one in person?

dcwhite: That way the binary values are hidden and turn into named constants (remember no magic numbers!). This is done all the time in Qt with "flag enums" that have values of powers of 2, and are bit-shifted and bit-arithmetic'ed together in different ways

dcwhite: https://stackoverflow.com/questions/1448396/how-to-use-enums-as-flags-in-c

src/Modules/Visualization/ShowField.cc
Haydelj: Not quite. This would defeat the whole point of me writing it the way I did. I wanted each one to be its own function so I could move the conditional statements out of the write functions (to replace with the switch statement). This was done for performance reasons since this is the tightest loop of the face generation code. And while the void initializer is janky it was again done for performance reasons. Writing a recursive template destroyed performance so I used an initializer list instead (hopefully there will be a cleaner way of doing this in the future but currently there is not).

Haydelj:
No there weren't any noticeable changes to brute force performance. The node generation with spheres and edge generation with cylinders take orders of magnitude longer than anything else and I didn't touch that code yet. There was about a 20% improvement in large latvol generation

src/Modules/Visualization/ShowField.cc
Haydelj: numNodes refers to the number of nodes in a face ie either 3 for triangular faces or 4 for quadrilateral faces. I'm fairly comfortable with just constructing all three

dcwhite:
@jessdtate I'm still working on the super-generic double-refactoring for the main loop logic, but @Haydelj and I had a good discussion about it and definitely want to try it out. This version of the PR ought to be merged first though, since a lot of other things need to be worked on related to ShowField. So let's try to get it tested and merged in by Tuesday.

dcwhite:
Also, this crashes with the basic latvol test on MacOS 10.10 (the build machine):
```

Thread 14 Crashed:
0   libsystem_platform.dylib      	0x00007fff860a6fc9 _platform_memmove$VARIANT$Unknown + 41
1   libc++.1.dylib                	0x00007fff8d4adf25 std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::__init(char const*, unsigned long) + 91
2   libCore_Datatypes.dylib       	0x0000000110965c72 SCIRun::Core::Datatypes::ColorMap::getColorMapName() const + 18
3   libModules_Visualization.dylib	0x000000010e2f428e SCIRun::Modules::Visualization::detail::GeometryBuilder::renderFacesLinear(boost::shared_ptr<SCIRun::Field>, boost::optional<boost::shared_ptr<SCIRun::Core::Datatypes::ColorMap> >, SCIRun::Core::Thread::Interruptible*, SCIRun::RenderState, boost::shared_ptr<SCIRun::Graphics::Datatypes::GeometryObjectSpire>, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 1166
4   libModules_Visualization.dylib	0x000000010e2f15cb SCIRun::Modules::Visualization::detail::GeometryBuilder::renderFaces(boost::shared_ptr<SCIRun::Field>, boost::optional<boost::shared_ptr<SCIRun::Core::Datatypes::ColorMap> >, SCIRun::Core::Thread::Interruptible*, SCIRun::RenderState, boost::shared_ptr<SCIRun::Graphics::Datatypes::GeometryObjectSpire>, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 267
5   libModules_Visualization.dylib	0x000000010e2edd86 SCIRun::Modules::Visualization::detail::GeometryBuilder::buildGeometryObject(boost::shared_ptr<SCIRun::Field>, boost::optional<boost::shared_ptr<SCIRun::Core::Datatypes::ColorMap> >, SCIRun::Core::GeometryIDGenerator const&, SCIRun::Core::Thread::Interruptible*) + 2054
6   libModules_Visualization.dylib	0x000000010e2ec946 SCIRun::Modules::Visualization::ShowField::execute() + 342
7   libDataflow_Network.dylib     	0x000000010e7011ce SCIRun::Dataflow::Networks::Module::executeWithSignals() + 414
8   libInterface_Application.dylib	0x000000010b72e3bf non-virtual thunk to SCIRun::Gui::ModuleWidget::executeWithSignals() + 79
9   libEngine_Scheduler.dylib     	0x000000010ce85106 SCIRun::Dataflow::Engine::DynamicExecutor::ModuleExecutor::run() + 278
10  libCore_Datatypes_Legacy_Field.dylib	0x000000010fa469b6 boost::(anonymous namespace)::thread_proxy(void*) + 54
11  libsystem_pthread.dylib       	0x00007fff91fc405a _pthread_body + 131
12  libsystem_pthread.dylib       	0x00007fff91fc3fd7 _pthread_start + 176
13  libsystem_pthread.dylib       	0x00007fff91fc13ed thread_start + 13
```

src/Modules/Visualization/ShowField.cc
```diff
@@ -445,757 +486,327 @@ void GeometryBuilder::renderFacesLinear(
...
-  std::vector<Vector> vvals;
-  std::vector<Tensor> tvals;
-  std::vector<ColorRGB> face_colors;
 
-  if (fld->basis_order() < 0 || state.get(RenderState::USE_DEFAULT_COLOR))
-  {
-    colorScheme = ColorScheme::COLOR_UNIFORM;
-  }
-  else if (state.get(RenderState::USE_COLORMAP))
+  auto realColorMap = colorMap.get();
```
dcwhite: There's something just not right with this block (524-532). It should be rewritten into one function and tested separately.

dcwhite: It's leading to the crash with grayscale coloring.

dcwhite: If I set a default color in ShowField it does not crash

Haydelj:
@jessdtate I can expose the shaders if you want that in this pr and I can write the toon shader too if you want. The environment map shader will take some more work so probably not in this pr. Also, shininess does work (the shaders are for the most part the same as the old ones so anything that did work still works and anything that didn't still doesn't) and I can implement the emission and diffuse components either right now or after the QOpenGLWidget

src/Modules/Visualization/ShowField.cc
dcwhite: Right, move the creation of `map` to the block where it is used, and rename it. That should fix the crash on 10.10. Then, be more explicit about the need to create a default colormap just to create a coordinateMap--can you just compute the scale and shift?. Plus the constant parameters for the coordinateMap are magic numbers. I'd like to see a function call `createCoordinateMap` that encapsulates all that.

dcwhite: @Haydelj required review changes above

jessdtate:
We don't have to do those things in this PR, but it's hard to test the shaders without them exposed.  We can make some issues for those.

PR #1979, opened 2019-08-27
QOpenGLWidget 

Haydelj:
Partially working implementation of the new QOpenGLWidget to replace the deprecated QGLWidget class


dcwhite:
@Haydelj Time to merge master and try Windows build

RubioJr9:
Works on Linux :confetti_ball:

src/Interface/Modules/Render/ES/SRInterface.h
```diff
@@ -348,11 +352,17 @@ namespace SCIRun {
 
       GLuint                              mFontTexture        {};       // 2D texture for fonts
 
+<<<<<<< HEAD
```
dcwhite: Not sure this is valid C++...

Haydelj: Hmm, what C++ version are we using?

dcwhite:
@Haydelj Qt5+Travis is ongoing, will take a while since Travis is slow and there are even more versions to test now.

RubioJr9:
Docking Fix:
```c++
int GuiApplication::run(int argc, const char* argv[])
{
  QApplication::setAttribute(Qt::AA_ShareOpenGLContexts);
  QApplication app(argc, const_cast<char**>(argv));
  ...
}
```

dcwhite:
@Haydelj I got a VS state toString dump to console at one point--might want to track that down and disable it.

dcwhite:
I'm seeing more regression test failures than before due to the screenshot not getting passed out as a matrix, but the ones I checked were still rendering fine. So this is probably the same problem with rendering/threading/state updating not all in line, but more so with the new widget structure. I'll check other systems and compare some more.

dcwhite:
@Haydelj Just got a whiole bunch of VS state dumps--please remove those before merge can happen.

Haydelj:
> @Haydelj Just got a whole bunch of VS state dumps--please remove those before merge can happen.

I removed it

dcwhite:
Waiting on #1991 so we can test Travis builds

dcwhite:
@Haydelj needs a master/qt5-dev update as we talked about

dcwhite:
@Haydelj Ready for upstream/qt5-dev merge

dcwhite:
Merge today into side branch

Issue #1962, opened 2019-07-02
SCIRun doesn't exit properly on Linux


Haydelj:
Closing SCIRun on Linux closes the Window but doesn't actually exit and return your terminal prompt.

dcwhite:
Yeah I've seen this on Pop_OS (Ubuntu variant). Always had to open a new terminal tab/window, quite annoying.

PR #1985, opened 2019-09-06
Fixed Linux closing bug

RubioJr9:
Issuse: #1962
Tested on: Arch Linux

@dcwhite @Haydelj please test on windows since you have working builds.
I will test on mac os. Other variants of Linux would be good to test on too.


src/Dataflow/Engine/Controller/NetworkEditorController.cc
```diff
@@ -102,6 +102,7 @@ NetworkEditorController::~NetworkEditorController()
 #ifdef BUILD_WITH_PYTHON
   NetworkEditorPythonAPI::clearImpl();
 #endif
+  executionManager_.stop();
```
dcwhite: This makes sense to stop explicitly since it just loops forever.

src/Interface/Application/SCIRunMainWindowQtOverrides.cc
```diff
@@ -108,6 +108,7 @@ void SCIRunMainWindow::closeEvent(QCloseEvent* event)
   {
     writeSettings();
     event->accept();
+    quit();
```
dcwhite: Is this still needed with the executionManager.stop call? It's risky since it's buried in a conditional, and might not be needed on other platforms (we don't want to call quit() twice)

RubioJr9: On second thought, this might not be necessary.

dcwhite: Test it out without the extra quit and see if it works

RubioJr9:
> We should use this change to address any other application-level threads that could be stopped more gracefully upon closing (for instance, when a network is executing and the user clicks the close button).

I think stopping the execution manager also stops modules from executing.

src/Interface/Application/SCIRunMainWindowQtOverrides.cc
RubioJr9: I don't have any problems without it. I thought I was getting a bug without that call. I reverted some code before committing so I think it came from there.

dcwhite:
I think it ought to stop more modules from executing, but does it halt current module execution threads? That codes is too old for me to remember...

RubioJr9:
I just did some tests and all the threads stop when the application is closed. I created multiple latvol > showfield > viewscene networks and they would all shut down their executions. Though if I had about 10-12 networks executing(I have 8 threads), the exit window/save network actions were being blocked and I had to wait until there were less executions left before I could interact. I think the OS might be prioritizing the executions threads over the qt threads.

RubioJr9:
I just fixed a seg fault on Mac. Things seem good on there now.

Issue #1989, opened 2019-09-16
Expanded ViewScene screenshot functionaity--share screenshot as issue


dcwhite:
Somewhat parallel to #938. 

If we add a menu to the VS screenshot button

![image](https://user-images.githubusercontent.com/1399383/64984926-07844300-d881-11e9-8d32-34ea37f752de.png)

I can think of 3 options right now to code up:
1. (quick mode--TODO this issue or later) save screenshot to user's desktop using generated filename, no pop-up dialog
2. (current slot behavior) save screenshot after file save dialog input
3. (new feature for troubleshooting, this issue) save screenshot and send in an email message, along with current network state saved as xml attachment, to a certain email address or slack endpoint for sharing/discussion.

This issue is to add the GUI parts. I'm still thinking about the email endpoint, but for now we can send to my work email.

PR #1997, opened 2019-09-27
ViewScene bug report button

RubioJr9:
Added a button in the view scene toolbar that saves a screenshot and pulls up an email template to send to @dcwhite.
Opens through the default mail client.
Requires qt5 to get system info.

Issues:
#1989

Edit:
Tested on MacOS 10.13 and Arch Linux
This won't pass travis until all builds are qt5


src/Interface/Modules/Render/Screenshot.h
```diff
@@ -46,13 +46,13 @@ namespace SCIRun
     public:
       explicit Screenshot(QGLWidget *glwidget, QObject *parent = nullptr);
```
dcwhite: This won't merge easily with https://github.com/SCIInstitute/SCIRun/pull/1979/files#diff-09b4cb29ce5bcc63aba5130a741d0f1a. I guess we can wait and see after #1979 is merged into qt5-dev.

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -299,6 +302,15 @@ void ViewSceneDialog::addScreenshotButton()
   addToolbarButton(screenshotButton);
 }
 
+//--------------------------------------------------------------------------------------------------
+void ViewSceneDialog::addBugReportButton() {
```
dcwhite: SCIRun standard is all open braces on own newline

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -2145,6 +2157,51 @@ void ViewSceneDialog::screenshotClicked()
...
+  std::string gpuVersion = reinterpret_cast<const char *>(glGetString(GL_RENDERER));
+
+  // Temporarily save screenshot so that it can be sent over email
+  takeScreenshot();
+  QImage image = screenshotTaker_->getScreenshot();
+  QString location = QDir::homePath() + QLatin1String("/scirun5screenshots/scirun_bug.png");
+  image.save(location);
+
+  // Generate email template
+  static std::string askForScreenshot = "\nIMPORTANT: Make sure to attach the screenshot of the ViewScene located at "
```
dcwhite: I'd change all this to using QString rather than std::string, then use QStringBuilder operations to look cleaner, as in https://doc.qt.io/qt-5/qstring.html#more-efficient-string-construction

src/Interface/Modules/Render/Screenshot.h
RubioJr9: I think the function for grabbing frame buffer is still the same.

src/Interface/Modules/Render/ViewScene.cc
RubioJr9: Sorry, my ide auto formats sometimes

src/Interface/Modules/Render/ViewScene.cc
RubioJr9: Great tip. It should be a lot more efficient now.

jessdtate:
do we want this option on the main vs window?  what about in the configure window?

dcwhite:
@RubioJr9 This needs a proper merge from master--the method naming changed on the new opengl widget.

RubioJr9:
It's merged properly now

dcwhite:
@RubioJr9 will move the button to the new developer tab.

dcwhite:
@RubioJr9 I merged this one at long last and just created another issue to continue my requests (#2022)

PR #2001, opened 2019-09-30
Fixed disappearing scene

RubioJr9:
It was normalizing a zero vector if movement was small.

Tested on:
- [X] Mac
- [X] Linux


src/Externals/spire/arc-ball/ArcBall.cpp
```diff
@@ -76,7 +76,13 @@ void ArcBall::setLocationOnSphere(glm::vec3 location, glm::vec3 up)
 //------------------------------------------------------------------------------
 glm::quat ArcBall::quatFromUnitSphere(const glm::vec3& from, const glm::vec3& to)
 {
-  glm::vec3 axis = glm::normalize(glm::cross(from, to));
+  glm::vec3 cross = glm::cross(from, to);
+  glm::vec3 axis = glm::normalize(cross);
+
+  // Give arbitrary non-zero vector because no rotation
+  if (glm::length(cross) == 0.0)
```
jessdtate: Shouldn't this have some tolerance?

dcwhite: I was thinking the same thing

RubioJr9: I don't think it would have been a problem because cross products won't get small enough since it is based on pixel positions. I changed it to check nan instead so it should be more robust without the arbitrary threshold.

dcwhite:
To test, move single pixel at a time.

dcwhite:
Merge today

jessdtate:
I can still get it to disappear with the zoom on mac.

dcwhite:
So no merge today?

jessdtate:
I put a cout statement.  Sometimes the angle is nan still.  When I added that check, I haven't been able to get it to disappear. 
`if(angle <= 0.00001 || std::isnan(angle))`

jessdtate:
I guess we can merge it and make the fix.

RubioJr9:
I can fix it now

RubioJr9:
Done

dcwhite:
Thanks guys

Issue #1983, opened 2019-08-30
Qt 5: Investigate all screens and module UIs for bad stylesheets


dcwhite:
From #1774

dcwhite:
Try using Qt 5.14--requires #1988

dcwhite:
I made a spreadsheet:
https://docs.google.com/spreadsheets/d/1XgJKGOi4t_aWotbahbDzBb5lH1IxGdK7H0V4HbO-7Oc/edit?usp=sharing

dcwhite:
**Analysis (2026-06-13)**

Stylesheet audit for Qt5 compatibility. A tracking spreadsheet exists (linked in comments). The issue depends on Qt 5.14+ (#1988) since some stylesheet regressions only manifest on newer Qt versions.

The scope is identified: systematically work through the spreadsheet entries. No source code investigation needed here — it's a UI polish pass. Suggest tackling alongside or after the Qt version upgrade.

Issue #648, opened 2014-09-24
Beautify Module Widget part 3: Linux


dcwhite:
Continuation of #109.

- [ ] Module widget size is not scaling 100% to name (chops off last 6 letters in MFDFSTD for instance)
- [x] Checkboxes and radio buttons are missing outlines--hard to know they're there

dcwhite:
Post-Qt5 #513

dcwhite:
@jessdtate Might be one for @RubioJr9

dcwhite:
Hey @RubioJr9 when you have a free 20 minutes, could you open up all the module UIs on your linux system and just check for any gross abnormalities? Thanks

RubioJr9:
Yeah I can do that. One thing I have noticed is that the ViewScene options window does not let me move it around.

RubioJr9:
I made it pretty comprehensive, but there are 4 critical ones listed up top.
[SCIRun_Linux_beautify.pdf](https://github.com/SCIInstitute/SCIRun/files/5575771/SCIRun_Linux_beautify.pdf)

dcwhite:
I figured out a way to fix the invisible checkbox/radio button problem. But man those Linux module layouts are stubborn. I'll see what I get done today then push the rest for next milestone.

dcwhite:
The module UIs on Linux look infinitely better NOT in 4k/hiDPI. @RubioJr9, what screen resolution did you use for your report?

RubioJr9:
@dcwhite  I was using 1440p. The scaling is much worse in 2560x1600 on my 13" macbook. Text is a lot more cut off.

PR #2004, opened 2019-10-03
qt5 Stylesheet Fixes

RubioJr9:
Fixed:
-black text on gray backgrounds
-button styling
-ShowFieldGlyphs spacing
-ViewScene look axis dropdown spacing

Related Tickets:
#1983

Reviewers:
@jessdtate @dcwhite


RubioJr9:
Just a heads up, I redid the ShowFieldGlpyhs ui and used the proper layout boxes. I think it will do better with smushing on different configurations

RubioJr9:
Made some changes to showfieldglyphs for mac os. Everything seems good except for coloring of groupbox contents. I wasn't able to figure that out before having to leave. I'm changed background colors in the stylesheet to set everything to gray. The groupbox style is being set as well but I couldn't figure out a way to exclude that or set it back to default. It looks right in qt designer but it set differently in the application.

dcwhite:
Cool, thanks

dcwhite:
Cool, thanks

src/Interface/Application/SCIRunMainWindow.cc
```diff
@@ -102,33 +102,6 @@ SCIRunMainWindow::SCIRunMainWindow()
...
 
-  setStyleSheet(
-    "background-color: rgb(66,66,69);"
-    "color: white;"
-    "selection-color: yellow;"
-    "selection-background-color: blue;"//336699 lighter blue
-    "QToolBar {        background-color: rgb(66,66,69); border: 1px solid black; color: black;     }"
-    "QProgressBar {        background-color: rgb(66,66,69); border: 0px solid black; color: black  ;   }"
-    "QDockWidget {background: rgb(66,66,69); background-color: rgb(66,66,69); }"
-    "QPushButton {"
```
dcwhite: @RubioJr9 These QPushButton stylesheet options seem to have been lost. Did they not apply anywhere so you removed them? Or were they mistakenly not copied to the app level?

RubioJr9: I'll double check but I believe they didn't have any visible effect.

RubioJr9:
Screen Shot 2019-11-07 at 2 43 10 PM
(Left is last release build; Right is this branch)

Some slight differences but the buttons don't seem to be affected

dcwhite:
Ok cool. Let's merge it

RubioJr9:
Turns out I took those out because they looked bad on linux, but had no effect on macs

dcwhite:
Ah ok. We can look tomorrow

dcwhite:
#648

PR #2010, opened 2019-10-14
Qt5 dev merge

dcwhite:
Now that the Qt4 tag is attached, the Qt5 widget work can be merged into master.


src/Modules/Render/ViewScene.cc
```diff
@@ -229,6 +229,7 @@ void ViewScene::syncMeshComponentFlags(const std::string& connectedModuleId, Mod
 
 void ViewScene::execute()
 {
+  /*
```
dcwhite: @Haydelj Was this "to be worked on later"? Not that it worked well before...

Haydelj: Yeah as of now if that code is uncommented viewScene will infinitely execute. Since it was already broken I figured we could figure out why this happens when we get around to making it work correctly

Haydelj: I can take a look at it now if you would prefer though

dcwhite: It can wait

dcwhite:
@jessdtate I need to disable the Qt4 builds on Travis before we merge this. Commit incoming. Test it out today?

jessdtate:
I can do some builds this evening

PR #2013, opened 2019-10-23
To string button

Haydelj:
Adds a developer tab and a button for printing out the state of viewScene and spire


src/Interface/Modules/Render/ViewScene.h
```diff
@@ -76,6 +76,8 @@ namespace SCIRun {
 
 
     protected Q_SLOTS:
+      void toString() const {std::cout << toString("");}
```
dcwhite: `toString` is for creating/returning a string, this should be named `print`

src/Interface/Modules/Render/ViewSceneControls.ui
```diff
@@ -6,14 +6,14 @@
...
-    <height>662</height>
+    <width>567</width>
+    <height>1101</height>
    </rect>
   </property>
   <property name="minimumSize">
    <size>
     <width>496</width>
-    <height>662</height>
+    <height>1101</height>
```
dcwhite: This minimum size is wrong. It should probably be reverted back to 662. The new tab just adds one button. If anything, the minimum width should be increased--I can smush the tabs in when I resize the config screen.

src/Interface/Modules/Render/ViewScene.h
dcwhite: Please fix

dcwhite:
Also when the button is pressed, it turns a big block of red--kind of ugly.

dcwhite:
As far as actual functionality goes, it works great!

src/Core/Application/Version.cc
```diff
@@ -0,0 +1,32 @@
+/*
```
dcwhite: Dude, stop committing this one--it is auto-generated in the bin folder now.

PR #2015, opened 2019-10-28
Send Screenshot downstream for testing fix

Haydelj:
Fixes view scenes output to contain an up to date frame


src/Interface/Modules/Render/GLWidget.cc
```diff
@@ -70,12 +70,20 @@ GLWidget::~GLWidget()
...
 {
 	spire::glPlatformInit();
-}	
+}
 
 void GLWidget::paintGL()
 {
-  mCurrentTime += updateTime;
-  mGraphics->doFrame(mCurrentTime, updateTime);
+  //set to 200ms to force promise fullfilment every frame if a good frame as been requested
```
dcwhite: How does this affect framerate? Or does it?

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -748,10 +762,29 @@ void ViewSceneDialog::newGeometryValue(bool forceAllObjectsToUpdate)
     }
   }
 
-  sendScreenshotDownstreamForTesting();
+  if(pushScreenShotToState && mGLWidget) mGLWidget->requestFrame();
```
dcwhite: Ah, ok, looks like only when a screenshot is expected.

src/Modules/Render/ViewScene.cc
```diff
@@ -51,20 +51,31 @@ using namespace SCIRun::Core::Logging;
...
-ViewScene::ViewScene() : ModuleWithAsyncDynamicPorts(staticInfo_, true), asyncUpdates_(0)
+ViewScene::ViewScene() : ModuleWithAsyncDynamicPorts(staticInfo_, true)
 {
   RENDERER_LOG_FUNCTION_SCOPE;
   INITIALIZE_PORT(GeneralGeom);
   INITIALIZE_PORT(ScreenshotDataRed);
   INITIALIZE_PORT(ScreenshotDataGreen);
   INITIALIZE_PORT(ScreenshotDataBlue);
+
+  get_state()->setTransientValue(Parameters::VSMutex, new Mutex("ViewSceneScreenShotMutex"), true);
```
dcwhite: No raw pointers/`new` allowed. This can be a `std::unique_ptr`, or just store it as an instance variable value and pass an address to the state.

src/Modules/Render/ViewScene.cc
```diff
@@ -51,20 +51,31 @@ using namespace SCIRun::Core::Logging;
...
+}
+
+ViewScene::~ViewScene()
+{
+  auto screenShotMutex = get_state()->getTransientValue(Parameters::VSMutex);
```
dcwhite: Follow the above, and, no destructor needed!

src/Modules/Render/ViewScene.cc
```diff
@@ -229,56 +237,30 @@ void ViewScene::syncMeshComponentFlags(const std::string& connectedModuleId, Mod
 
 void ViewScene::execute()
 {
-  /*
-  // hack for headless viewscene. Right now, it hangs/crashes/who knows.
+  fireTransientStateChangeSignalForGeomData();
+  auto state = get_state();
+  auto screenShotMutex = state->getTransientValue(Parameters::VSMutex);
+  auto mutex = transient_value_cast<Mutex*>(screenShotMutex);
+  mutex->lock();
```
dcwhite: Better way to lock/unlock a mutex around a scope: the `Guard` class (also in `Mutex.h`) grabs the mutex and locks it, then unlocks in the destructor.

src/Interface/Modules/Render/ViewScene.h
```diff
@@ -72,15 +72,19 @@ namespace SCIRun {
...
       void mousePressSignalForTestingGeometryObjectFeedback(int x, int y, const std::string& selName);
 
 
     protected Q_SLOTS:
       //---------------- New Geometry --------------------------------------------------------------
       void updateModifiedGeometries();
+      void updateModifiedGeometriesAndSendScreenShot();
       void updateAllGeometries();
-      void newGeometryValue(bool forceAllObjectsToUpdate);
+      void newGeometryValue(bool forceAllObjectsToUpdate, bool pushScreenShotToState);
```
dcwhite: The double-bool parameter list starts getting ugly. (client: "is it f(true, false) or f(false, true)???"). A labeled enum can be helpful, with the 4 options given distinct names.

dcwhite: I guess you already have named methods that wrap the various bool-combinations. That will work as long as the `newGeometryValue` is private, which it probably can be here.

src/Modules/Render/ViewScene.cc
dcwhite: Also, with the change to address, this code need not change--it would need to change for unique_ptr though.

src/Interface/Modules/Render/GLWidget.cc
Haydelj: This will only happen on execution and execution will halt until we get a "good frame" so whether or not this effects frame rate doesn't really matter. What it does do is accelerate execution.

dcwhite: An interesting test case then is animation using GetMatrixSlice, which triggers many executions in a row.

Haydelj:
This currently breaks widgets. @RubioJr9 and I will have to take a look at why.

jessdtate:
Breaks widgets, probably because it doesn't ever finish executing.

dcwhite:
Seems to break everything...

dcwhite:
@dcwhite Todo: add a lock log to see what's blocking what

Haydelj:
On my machine, this fix works fine with and without widgets. I'll try building it on Linux and see if that's any different.

Haydelj:
I was able to replicate the issue and the new core dump feature revealed it is being caused by a problem in the shader promise system. The issue is that the promise is fulfilled but never removed.

dcwhite:
Cool!

Haydelj:
@dcwhite @jessdtate I think I fixed the issue will y'all test it on your machines to varify

jessdtate:
The merge conflicts make it hard to test

jessdtate:
seems to behaving like you described before, ie, it will finish executing, but only if the render window is open.  This is improvement.  Could we add a check for the window being open so it doesn't get stuck?

dcwhite:
@Haydelj That's not a platform-neutral way to sleep--try https://en.cppreference.com/w/cpp/thread/sleep_for

dcwhite:
@Haydelj And do please fix the merge conflict

Haydelj:
@jessdtate @dcwhite I fixed the things y'all requested I fix. Now execution will continue even if the window is closed

dcwhite:
I get this bad crash with multiple re-executes:
`SCIRun_test(71368,0x70000efcd000) malloc: Double free of object 0x7f7fc75e29f0`
Let me get a more consistent replication script

dcwhite:
Haven't been able to replicate the crash. 

What I am seeing though, and most regression networks can be used to test this, is that the Matrix output from ViewScene doesn't work until the VS window is open. Is that expected at this point? It's true there is no longer any hang.

Haydelj:
@dcwhite yes due to the limitations of using QOpenGlWidget as our render target that is expected behavior. Qt won't let s render when the window is hidden. We could render to texture directly to get around this but I believe that would suffer from the same bug that the selection buffer rendering suffers from. It also wouldn't solve the problem that context initialization doesn't happen until the window is shown.

dcwhite:
Right, ok. Let me test some other regression network cases

dcwhite:
Regression tests are passing (well enough at least). While I'm no fan of arbitrarily throwing sleep calls around to solve deadlock problems, this does seem to work for now. Next we need to figure out how to get the screenshot data out in a guaranteed way downstream to enable some actual image-based regression testing.

dcwhite:
Actually, on Windows, this might have been a first: all regression networks passed! Before, there were always random crashes coming from ViewScene code. So this is a great improvement in functionality, assuming that double-delete crash was just a fluke...(it was on Mac btw).

PR #2020, opened 2019-11-06
fixed light ui

### LINKED ISSUE #1474: viewscene: get lights to move correctly from GUI input

Haydelj:
Closes #1474

### DIFF
--- a/src/Interface/Modules/Render/ViewSceneControls.ui
+++ b/src/Interface/Modules/Render/ViewSceneControls.ui
@@ -39,7 +39,7 @@
        </font>
       </property>
       <property name="currentIndex">
-       <number>7</number>
+       <number>3</number>
       </property>
       <widget class="QWidget" name="ObjectTab">
        <attribute name="title">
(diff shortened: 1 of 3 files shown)


allywarner:
From a user point of view, what are the slider bars supposed to do? Say somebody has never used SCIRun, where would they find out what these bars are actually controlling? The buttons to choose color are also named "PushButton." I think a more meaningful name would be appropriate so that users would know what they were for too. Something simple like "Choose Light Color."

Haydelj:
I added tooltips to the sliders labeling them Azimuth and Inclination and I fixed the button labels

allywarner:
At the bottom of the UI, there's "Click on circle to change light color/brightness." Since you've removed the circle, I would change this instruction, and probably put it at the top so the user can see it first. Also the tool tips are great, but maybe a very short explanation of what those terms mean would be helpful.

dcwhite:
Consider a table widget to add widget context but reduce duplication:
https://doc.qt.io/qt-5/qtablewidget.html

It's possible to add sliders and checkboxes and push buttons in each cell. This will make it easy to enable more than 4 lights later on.

image

dcwhite:
@Haydelj A new merge conflict to fix.

The GUI is acceptable now, we can overhaul all those tabs with nicer tables and stuff later.

allywarner:
When you update the UI again, I would add some instruction to the tool tips. "Higher does this" and "lower does this" for the Azimuth and the Inclination.

PR #2023, opened 2019-11-08
Various

dcwhite:


dcwhite:
Ahhhh, too many builds. I'll need to split up the action build file, one for push and one for pull_request.

dcwhite:
@jessdtate @allywarner Needs review. And yes, please point out the crappy description so I can improve it.

dcwhite:
I'm going to add one more commit here for @nids2001. Please wait to review until that's in

allywarner:
It would be nice to be able to filter based on UI and then use the "Starts with" within that filter.

dcwhite:
> It would be nice to be able to filter based on UI and then use the "Starts with" within that filter.

I was a bit lazy and added it as another option in the dropdown, so they can't be combined yet. What's the use case, just to reduce the result list? Mainly I added this option for developer testing of all UIs.

allywarner:
Yeah! It's always nice to be able to filter within a filter, but I can't see a lot of users needing that.

Issue #1959, opened 2019-06-30
minor z fighting


jessdtate:
**Description**

With the autorotate feature, Z fighting can be a little more obvious.  See the video

[autorotate.mov.zip](https://github.com/SCIInstitute/SCIRun/files/3332324/autorotate.mov.zip)

**To Reproduce**
1. make a network to visualize a latvol
2. align latvol to an axis
3. autorotate along axis
4. see z fightting

**Expected behavior**
it would be nice if this was smoother

**Version**
OS X 10.12

dcwhite:
Will discuss at next week's meeting

jessdtate:
Edge z fighting too.  This should be easier to fix.

jessdtate:
[edge_fighting.srn5.zip](https://github.com/SCIInstitute/SCIRun/files/3912994/edge_fighting.srn5.zip)

PR #2031, opened 2019-12-04
Edge z-fighting fix

Haydelj:
Shifts the vertices towards the camera for flat-shaded objects


dcwhite:
Fixes the second part of #1959.

PR #2034, opened 2019-12-07
ported color maps

Haydelj:
Added:
Lighthue
Don
*Tint


dcwhite:
+900 XP (100/cm)

dcwhite:
The CodeFactor review is red due to the `redundant virtual for override functions` item. I'm not too worried about that one, so we can ignore.

dcwhite:
Looks good. Nice use of functions

PR #2038, opened 2019-12-15
BBox Widget/Multiple Widget Transformations

RubioJr9:
Changes:
-Translation -- Left Click
-Translation along axis -- Right Click
-Rotation -- Left Click
-Rotate bounding box axes -- Right Click
-Scale entire widget -- Left Click
-Scale entire widget along axis -- Right Click
-Scale bidirectionally along one axis -- Left Click
-Scale unidirectionally along one axis -- Right Click
-Setting multiple transformations to one widget(ex. left click for rotation, right click for scaling)
-Multiple transformations for one widget movement(ex. scaling along axis causes scaling and translation)
-Set different colors for right and middle click on widget selection
-Set/reset position in module dialog
-Changing size of widgets in module dialog
-Update downstream
-Added orthogonalization and string operators to the Transform class
-Added oriented bounding box class
-Moved feedback datatypes to Feedback.h/Feedback.cc

Note:
Changing the output size in the menus would be kind of weird. Because now you that can rotate the mesh such that it's no longer axis aligned, calling them x,y,z axes isn't correct. If we scale along the world space x,y,z axes anyways, it will cause shearing.


src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -564,9 +582,8 @@ namespace SCIRun {
       mCore.execute(0, 50);
```
RubioJr9: After merging master, this core execution sometimes causes the scene to disappear for a moment if you shift click on viewscene. My guess is that it's from the downstream screenshot changes.

src/Core/GeometryPrimitives/Transform.h
```diff
@@ -77,6 +79,7 @@ namespace SCIRun {
 
         Transform();
         Transform(const Transform&);
+        Transform(double* pmat);
```
dcwhite:
```suggestion
        explicit Transform(double* pmat);
```

src/Graphics/Widgets/BoundingBoxWidget.cc
```diff
@@ -26,87 +26,185 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
...
+
+void BoundingBoxWidget::createWidgets(const GeometryIDGenerator& idGenerator, int widgetNum,
+                                      int widgetIter)
+{
+  std::string boxName = widgetName(BoundingBoxWidgetSection::BOX, widgetNum, widgetIter);
+  widgets_.push_back(WidgetFactory::createBox(idGenerator, diagonalLength_ * scale_ * boxScale_,
+                                              center_, corners_, bbox_));
+  widgets_[++widgetsIndex_]->setToTranslate(MouseButton::LEFT);
+
+  for(int i = 0; i < 8; i++)
```
dcwhite: Magic number --> named constant

src/Graphics/Widgets/BoundingBoxWidget.cc
```diff
@@ -26,87 +26,185 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
...
+    addCornerSphere(i, idGenerator, scaledEigvecs_[i/2], widgetNum, widgetIter);
+
+  for (int i = 0; i < 6; i++)
```
dcwhite: Magic number --> named constant

src/Graphics/Widgets/BoundingBoxWidget.cc
```diff
@@ -26,87 +26,185 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
...
+    addFaceSphere(i, idGenerator, widgetNum, widgetIter);
+
+  glm::mat4 scaleTrans = glm::mat4(1.0);
+  for(int i = 0; i < 3; i++)
+    for(int j = 0; j < 3; j++)
+      scaleTrans[i][j] = eigvecs_[i][j];
+
+  for (int i = 0; i < 6; i++)
```
dcwhite: Magic number --> named constant

src/Graphics/Widgets/Widget.h
```diff
@@ -43,23 +43,37 @@ namespace SCIRun
...
-        void setToRotate();
-        void setToTranslate();
+        void setToScale(MouseButton btn, const Core::Geometry::Vector& flipAxis);
+        void setToScaleAxis(MouseButton btn, const Core::Geometry::Vector &scaleAxis, glm::mat4 scaleTrans, int scaleAxisIndex);
+        void setToScaleAxisHalf(MouseButton btn, const Core::Geometry::Vector &scaleAxis, glm::mat4 scaleTrans, int scaleAxisIndex);
+        void setToRotate(MouseButton btn);
+        void setToTranslate(MouseButton btn);
         glm::vec3 getFlipVector();
-        WidgetMovement getMovementType();
+        glm::mat4* getScaleTransform();
```
dcwhite: Don't return pointers to internals. Use a reference instead to avoid copying, const reference if possible.

src/Graphics/Widgets/Widget.cc
```diff
@@ -61,28 +61,59 @@ void WidgetBase::setPosition(const Point& p)
...
 
-WidgetMovement WidgetBase::getMovementType()
+int WidgetBase::getScaleAxisIndex()
+{
+  return scaleAxisIndex_;
+}
+
+glm::mat4* WidgetBase::getScaleTransform()
+{
+  return &scaleTrans_;
```
dcwhite: Don't return pointers to internals. Use a reference instead to avoid copying, const reference if possible.

src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -355,6 +360,21 @@ namespace SCIRun {
     //---------------- Widgets ---------------------------------------------------------------------
     //----------------------------------------------------------------------------------------------
 
+    //----------------------------------------------------------------------------------------------
+    void SRInterface::setSelectedWidgetToRed(std::vector<WidgetHandle> objList)
```
dcwhite:
```suggestion
    void SRInterface::setSelectedWidgetToRed(const std::vector<WidgetHandle>& objList)
```

src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -610,49 +632,42 @@ namespace SCIRun {
...
-        widgetSelected_ = true;
-
-        //Calculate w value
-        float zFar = mCamera->getZFar();
-        float zNear = mCamera->getZNear();
-        float z = -1.0/(depth * (1.0/zFar - 1.0/zNear) + 1.0/zNear);
-        mSelectedW = -z;
+    void SRInterface::initializeWidgetMovement(MouseButton btn, const glm::ivec2& pos)
+    {
+      if (mSelected == "") return;
```
dcwhite:
```suggestion
      if (mSelected.empty()) return;
```

src/Interface/Modules/Render/ES/SRInterface.h
```diff
@@ -159,9 +150,11 @@ namespace SCIRun {
       void select(const glm::ivec2& pos, WidgetList& objList, int port);
       std::string &getSelection()          {return mSelected;}
       gen::Transform &getWidgetTransform() {return mWidgetTransform;}
+      void setSelectedWidgetToRed(std::vector<Graphics::Datatypes::WidgetHandle> objList);
```
dcwhite:
```suggestion
      void setSelectedWidgetToRed(const std::vector<Graphics::Datatypes::WidgetHandle>& objList);
```

src/Interface/Modules/Render/ES/SRInterface.h
```diff
@@ -119,9 +110,9 @@ namespace SCIRun {
       //       them and provide quick object feedback.
 
       //---------------- Input ---------------------------------------------------------------------
-      void inputMouseDown(const glm::ivec2& pos, MouseButton btn);
-      void inputMouseMove(const glm::ivec2& pos, MouseButton btn);
-      void inputMouseUp(const glm::ivec2& pos, MouseButton btn);
+      void inputMouseDown(Graphics::Datatypes::MouseButton btn, const glm::ivec2& pos, std::vector<Graphics::Datatypes::WidgetHandle> objList);
```
dcwhite:
```suggestion
      void inputMouseDown(Graphics::Datatypes::MouseButton btn, const glm::ivec2& pos, const std::vector<Graphics::Datatypes::WidgetHandle>& objList);
```

src/Interface/Modules/Render/ES/SRInterface.h
```diff
@@ -319,81 +316,83 @@ namespace SCIRun {
...
+      int                                              mZoomSpeed          {65};
+      MouseMode                                        mMouseMode          {MOUSE_OLDSCIRUN};  // Current mouse mode.
+
+      std::vector<uint64_t>                            mEntityList         {};
+      std::string                                      mSelected           {};       // Current selection
+      GLfloat                                          mDepth              {};
+      bool                                             mSelectionHack      {false};
+      glm::vec3                                        mOriginWorld        {};
+      glm::vec3                                        mFlipAxisWorld      {};
+      int                                              mScaleAxisIndex     {};
```
dcwhite: These should get initial values.

src/Interface/Modules/Render/ES/SRInterface.h
```diff
@@ -319,81 +316,83 @@ namespace SCIRun {
...
+      const glm::mat4*                                 mScaleTrans         {};
```
dcwhite:
```suggestion
      const glm::mat4*                                 mScaleTrans         {nullptr};
```

src/Interface/Modules/Render/ES/SRInterface.h
```diff
@@ -319,81 +316,83 @@ namespace SCIRun {
...
+      glm::vec3                                        mOriginToSpos       {};
+      glm::vec3                                        mOriginView         {};
+      glm::vec2                                        mSelectedPos        {};
+      float                                            mSelectedW          {};
```
dcwhite: These should get initial values--all numerical types below too

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -1256,15 +1269,15 @@ static std::vector<WidgetHandle> filterGeomObjectsForWidgets(SCIRun::Modules::Re
...
+std::vector<WidgetHandle> ViewSceneDialog::getObjList(const int x, const int y, MouseButton btn)
 {
   LOG_DEBUG("ViewSceneDialog::asyncExecute before locking");
   Guard lock(Modules::Render::ViewScene::mutex_.get());
   LOG_DEBUG("ViewSceneDialog::asyncExecute after locking");
 
   auto spire = mSpire.lock();
   if (!spire)
-    return;
+    return std::vector<WidgetHandle>();
```
dcwhite:
```suggestion
    return {};
```

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -1276,37 +1289,14 @@ void ViewSceneDialog::selectObject(const int x, const int y)
     if (!geomData)
     {
       LOG_DEBUG("Logical error: ViewSceneDialog received an empty list.");
-      return;
+      return std::vector<WidgetHandle>();
```
dcwhite:
```suggestion
      return {};
```
Or throw exception?

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -1276,37 +1289,14 @@ void ViewSceneDialog::selectObject(const int x, const int y)
...
-            pass.addUniform("uSpecularColor",
-              glm::vec4(0.1f, 0.0f, 0.0f, 1.0f));
-          }
-          break;
-        }
-      }
-    }
+    return objList;
   }
+  return std::vector<WidgetHandle>();
```
dcwhite:
```suggestion
  return {};
```

src/Graphics/Widgets/BoundingBoxWidget.cc
RubioJr9: There's 8 corners and 6 faces on a rectangle and this calls addCornerSphere() addFaceCylinder() so they're not really magic numbers

src/Graphics/Widgets/BoundingBoxWidget.cc
RubioJr9: Yeah, the loop with eigvecs should change

src/Graphics/Widgets/Widget.h
RubioJr9: Good tip

src/Graphics/Widgets/BoundingBoxWidget.cc
dcwhite: Any loop-ending non-named value is a magic number. True, the numbers are fixed, but a named variable there still increases readability and reduces typos:
```
for (int i = 0; i < numCorners; ++i) ...
```

dcwhite: I agree with this list of uses: https://en.wikipedia.org/wiki/Magic_number_(programming)#Accepted_uses

RubioJr9: I'll change it. I guess it is domain-specific knowledge

dcwhite: Think of it as embedding a deeper labeling of the domain as you explore it. Whoever explores next will be able to find their way more easily.

And also, vis-a-vis that list of accepted magic numbers, the 2 in `i/2` would be OK--no need for the overkill `i/numberThatDividesAllEvenNumbers`

RubioJr9:
The module/widget is done except for some UI options that might not work well with the implementation, like resizing. Also, I need to go back and update every module that uses widgets because SRInterface reads them in differently now

src/Modules/Fields/EditMeshBoundingBox.h
```diff
@@ -29,6 +29,7 @@
 #ifndef MODULES_FIELDS_EDITMESHBOUNDINGBOX_H
 #define MODULES_FIELDS_EDITMESHBOUNDINGBOX_H
 
+#include "Graphics/Widgets/BoundingBoxWidget.h"
```
dcwhite:
```suggestion
#include <Graphics/Widgets/BoundingBoxWidget.h>
```
Please change all #includes to use <> instead of ""

src/Modules/Fields/EditMeshBoundingBox.h
```diff
@@ -92,16 +99,35 @@ namespace SCIRun {
...
+        void adjustGeometryFromTransform(Core::Datatypes::MouseButton btn,
+                                         Core::Datatypes::WidgetMovement move,
+                                         const Core::Geometry::Transform& feedbackTrans);
+        void generateGeomsList();
+        void saveToParameters();
+        void loadFromParameters();
+        std::string convertForLabel(double coord);
+        void updateInputFieldAttributes();
+
+        Core::Geometry::Point mPos;
```
dcwhite: Change all these names. SCIRun naming standard (except for spire--nonstandard code) is `instanceVar_` trailing underscore

src/Modules/Fields/EditMeshBoundingBox.h
```diff
@@ -46,12 +47,17 @@ namespace SCIRun {
         public Has3OutputPorts < FieldPortTag, GeometryPortTag, MatrixPortTag >
       {
       public:
+        const int mDIMENSIONS = 3;
```
dcwhite:
```suggestion
        static const int mDIMENSIONS = 3;
```

dcwhite:
Lots more to come. Also, look into the build error related to hashtable.

Haydelj:
I branched off of this to start working on the ViewScene camera locking feature and have made heavy modifications to ViewScene/SRInterface please don't modify these files. If you find any bugs related to camera movement or see any warnings it's likely they are fixed on my branch.

RubioJr9:
I'm done with my changes. Ready for testing!

dcwhite:
@RubioJr9 Compile error due to forgotten file:
```
-- Configuring done
CMake Error at CMakeLists.txt:628 (ADD_LIBRARY):
  Cannot find source file:

    Feedback.h
```

RubioJr9:
Sorry, it's up now

jessdtate:
I'm having trouble with the widget mapping to the mouse movements again.  os x 10.15 qt 5.14.  I will try with 5.13.  Also, when a second module is added with the same input, it resets all the widgets in the network.  Is there a way to move just one side of the bbox?  The text fields in the module aren't connected at all, it seems.

RubioJr9:
I'll test it again on mac 10.14/qt 5.13.1 and see if I can replicate it.

RubioJr9:
What do you mean by trouble mapping widgets to mouse movements? I'm also seeing that adding a second module resets the data. You can move one side of a box by right clicking the green cylinder. Also what do you mean by the text fields not being connected?

jessdtate:
I have to click a different part of the window to select the widget, as if the widget positions are mapped to a bigger screen.  The text fields in the UI for center and size do not change with widget manipulation, nor do they change the bounding box when the values are changed.

RubioJr9:
Do you have the selection fix enabled in viewscene? I meant to disable the option for resizing in the module ui. Are you sure setting the center doesn't work? I haven't had issues with that. I'll get the spin boxes updating on widget translation though. Also, the data resetting when adding a second EditMeshBoundingBox to the same input is a bug with the Module class. Adding a new port tells all the downstream modules that the inputs were changed.

jessdtate:
ah yes,  I knew that we had fixed it, but I had forgotten about the option in viewscene.  And the position inputs are working.  There should be a note in the module doc that the set center button needs to be pressed to change it.  

One more thing I found, the thickness of the widget shouldn't change with the scaling.  It should stay the same size.

RubioJr9:
I can put the note in the docs. It'd be really clunky if the widget didn't change size on scale. Currently it's based on the shortest eigenvector which changes size on scaling. If there was input data that was very large or small, then you'd have to manually edit the widget size on initialization to get something sensible. Also, when you scale down, it'd make the widgets too large and occlude the object until you manually scaled it again. It'd end up looking like this image.

jessdtate:
It's not that hard to change the size.  It is inconsistently implemented.  I kind of like it when it's on the corners, with a global scaling.  It makes more sense there, but I don't like   When the edges are moved, it doesn't update real time and it looks more obviously weird. Also, it changes based on the change in scaling of one size, not the total object in these cases.  If you move only one side a lot, the widget size changes a lot without a large change in the bbox diameter.  That's frustrating to me.

I'm willing to leave as is, or with realtime update on the edge scaling, to see what people like.

dcwhite:
I suggest implementing @jessdtate's scaling suggestions first, while leaving the other method available as a user option, if that's not too difficult. Today I will look into the general module problem.

src/Core/GeometryPrimitives/BBox.h
```diff
@@ -251,6 +251,85 @@ class BBox {
...
+      Vector d = diagonal();
+      return cmin_ + (d * 0.5);
+    }
+
+  private:
+    std::vector<Vector> eigvecs_;
+    Point cmin_;
+    Point cmax_;
+    bool is_valid_;
+    bool axis_aligned_;
```
dcwhite: Remove this, it's never used, spamming warnings:
```/Users/dan/Dev/SCIRun/src/Core/GeometryPrimitives/BBox.h:329:10: warning: private field 'axis_aligned_' is not used [-Wunused-private-field]```

File: src/Graphics/Widgets/BoundingBoxWidget.h [hunk 22]
```diff
@@ -29,21 +29,107 @@ DEALINGS IN THE SOFTWARE.
...
+        std::vector<std::string> allIds_;
+        std::vector<std::string> translateIds_;
+        std::vector<std::vector<std::vector<std::string>>> translateIdsByFace_;
+        std::vector<std::vector<std::string> > translateIdsBySide_;
+        std::vector<std::vector<std::vector<std::string>>> rotateIdsByFace_;
+        std::vector<std::vector<std::vector<std::string>>> scaleIdsByFace_;
+        std::vector<std::string> rotateIds_;
+        std::vector<std::string> scaleIds_;
+        std::vector<std::vector<std::vector<std::string>>> scaleAxisIds_;
+        std::string widgetName(size_t i, size_t id, size_t iter);
```
dcwhite: As a blanket review comment, this is just way, way too many new private member variables. Come up with some classes that can break down this giant list, and give them names. No vector<vector<vector either.

dcwhite: https://refactoring.guru/smells/large-class

dcwhite:
Functionally, this is very close to done. Code-wise, there is too much to do in one PR. So I'm closing this one in order to review each piece separately. There are at least this many PRs here:

1. Transform class improvements, with tests
1. New bounding box primitive, with tests
1. New mesh function for OrientedBBox, with lots of tests
1. VS/SRInterface infrastructure to support fancier widget interaction
1. BoundingBoxWidget enhancement, which needs a big redo 

And probably more as we go a long.

dcwhite:
@Haydelj Please don't post any PRs branched off of this branch. They will be closed right away. Any work branched off of this branch will need a rebase/patch onto an updated master after this is merged in its new form.

Issue #1982, opened 2019-08-30
Qt 5: Modules stack on top of each other until you hover over


dcwhite:
From #1774

PR #2040, opened 2019-12-16
Qt5 problems, and their fixes

dcwhite:
Main quasi-fix: #1982. Still to do: better module placement method.
Also, parts of several other issues, see commit log and links below


RubioJr9:
I just tested it and I'm still seeing the stacking problem on 10.14.4

RubioJr9:
I just tested it again with the new commit. On the same mac os 10.14.4, when you drag a module, it drops on the right of the screen then snaps into the right position. I wasn't seeing that before. On linux, I got a seg fault the first time I use and execute a typical pattern. When I open it again and try that same pattern, I don't get a seg fault.

allywarner:
If I leave SCIRun for a while (about 5 minutes or so), go to another window, and come back to it the modules are on top of each other again and move when you hover over them.

jessdtate:
> If I leave SCIRun for a while (about 5 minutes or so), go to another window, and come back to it the modules are on top of each other again and move when you hover over them.
> 
> I'm testing on OSX 10.14.6 and Qt 5.13.0.

Not seeing this in 10.15 with qt5.13

PR #2049, opened 2020-01-24
Edit camera state variables

RubioJr9:
Converted CameraLookAt and CameraRotation to strings so they can be set through the python interface.


src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -522,10 +522,20 @@ void ViewSceneDialog::pullCameraLookAt()
   auto spire = mSpire.lock();
   if(!spire) return;
 
-  auto lookAt = toDoubleVector(state_->getValue(Modules::Render::ViewScene::CameraLookAt).toVector());
+  auto lookAt = pointFromString(state_->getValue(Modules::Render::ViewScene::CameraLookAt).toString());
   spire->setCameraLookAt(glm::vec3(lookAt[0], lookAt[1], lookAt[2]));
 }
 
+//--------------------------------------------------------------------------------------------------
+glm::quat ViewSceneDialog::stringToQuat(std::string &s)
```
dcwhite: Move this function to a new header/cc file, as a global function or in a new class, called ViewSceneUtility or the like.

src/Interface/Modules/Render/ViewScene.h
```diff
@@ -39,6 +39,7 @@ DEALINGS IN THE SOFTWARE.
 #include <Interface/Modules/Render/ViewSceneControlsDock.h>
 #include <Graphics/Datatypes/GeometryImpl.h>
 #include <Interface/Modules/Render/share.h>
+#include <glm/gtc/quaternion.hpp>
```
dcwhite: Part of: Move this function to a new header/cc file, as a global function or in a new class, called ViewSceneUtility or the like.

Two problems here: the glm is after the share.h. Also, we want to minimize glm includes in our header code.

src/Interface/Modules/Render/ViewScene.h
```diff
@@ -124,6 +125,7 @@ namespace SCIRun {
       void pullCameraRotation();
       void pullCameraLookAt();
       void pullCameraDistance();
+      glm::quat stringToQuat(std::string &s);
```
dcwhite:
```suggestion
      glm::quat stringToQuat(const std::string &s);
```

Move this function to a new header/cc file, as a global function or in a new class, called ViewSceneUtility or the like.

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -522,10 +522,20 @@ void ViewSceneDialog::pullCameraLookAt()
...
+{
+  std::istringstream is(s);
+  double w, x, y, z;
+  char st;
+  is >> st >> w >> st >> x >> st >> y >> st >> z >> st;
```
dcwhite: This is also a bit brittle when Python input is considered. Look into `boost::regex` to handle the intervening commas/spaces better. And add a unit test.

jessdtate:
The python behavior seems to be fine, but I can't seem to rotate the camera with the mouse.  Is anyone else seeing this?

dcwhite:
@jessdtate I tested Mac and Windows and had no problem with mouse rotation. Did you move it with python first? What were the exact testing steps?

dcwhite:
@RubioJr9 This also needs better error-checking. I set the rotation parameter to a "Point" (having just tested sending a point to CameraLookAt), and got this bizarrely rotated object (rendering a basic latvol):
image

dcwhite:
Then I set the rotation to a "5-tuple" string, and got this view, which I could not move/zoom away from (might have been what @jessdtate encountered):
image

dcwhite:
I'm going to close this one too and wait for a redo, with the changes mentioned above and better error-checking. It's not ready for milestone X.1.

RubioJr9:
How do I reopen a PR?

dcwhite:
Just make a new one.

PR #2050, opened 2020-01-29
CalculateInsideWhichField module

### LINKED ISSUE #2019: CalculateInsideWhichField module
CalculateInsideWhichField module needs the ability to set the Default Outside Value to NaN

nids2001:
closes #2019


dcwhite:
@nids2001 Cool. Please add a new test network to `ExampleNets/regression/modules` that we can use to test this out. If there's a `ReportFieldInfo` connected to the output, that's probably enough.

dcwhite:
Thanks!

src/Interface/Modules/Fields/CalculateInsideWhichFieldDialog.ui
```diff
@@ -254,6 +254,19 @@
...
+          <horstretch>0</horstretch>
+          <verstretch>0</verstretch>
+         </sizepolicy>
+        </property>
+        <property name="text">
+         <string>Use NaN</string>
+        </property>
+       </widget>
+      </item>
+      <item row="5" column="2">
```
dcwhite: One more thing to add to the UI. Check out this part of the related UI, when the Nan button is clicked, the spinner is disabled:
https://github.com/SCIInstitute/SCIRun/blob/master/src/Interface/Modules/Fields/MapFieldDataOntoElems.ui#L248

dcwhite: This can be added inside Qt Designer using the Signal/Slot interface.

allywarner: When you click the NaN checkbox, the default outside value is set to 999999999.00. Is this expected? The default value is 0.00 when opening the UI. When you uncheck the NaN checkbox, the number stays 999999999.00. Should it go back to the number it was before checking the box?

dcwhite: I think that's the behavior of the other modules with similar NaN options. We can think of a smoother transition later.

PR #2057, opened 2020-02-05
fixed screenshot bug

### LINKED ISSUE #2042: ViewScene "Take screenshot" saturates the saved screenshot (Windows 7)
**Describe the bug**
Saving a screenshot through the ViewScene button saturates the image from what the actual viewscene shows on the screen.

**To Reproduce**
1. Show a colored field in ViewScene
2. Click the "Take Screenshot" button and save as .png
3. Open the .png in an image viewer to see the saturated image

**Expected behavior**
"Take Screenshot" should save the image exactly how it looks in the ViewScene (this has been working for me in previous SCIRun 5 builds)

**Screenshots**

Result of "Take Screenshot"
saturated

What it should look like (Used Windows Snipping tool)
screenshotbug_whatitshouldlooklike

**Version**

end-of-qt4-support-220-ge8c0db661

**Desktop (please complete the following information):**
 - OS: Windows 7

Haydelj:
Fixed windows screenshot bug. Closes #2042


src/Interface/Modules/Render/Screenshot.cc
```diff
@@ -49,11 +49,21 @@ Screenshot::Screenshot(QOpenGLWidget *glwidget, QObject *parent)
...
 }
 
-QImage Screenshot::getScreenshot() {
-  return viewport_->grabFramebuffer();
+QImage Screenshot::getScreenshot()
+{
+	QImage image = viewport_->grabFramebuffer();
+	for (int j = 0; j < image.height(); ++j)
+	{
+		QRgb *row = (QRgb*)image.scanLine(j);
```
dcwhite: Cool fix. Let's make it more C++-like though:
```suggestion
		auto row = static_cast<QRgb*>(image.scanLine(j));
```

src/Interface/Modules/Render/Screenshot.cc
```diff
@@ -49,11 +49,21 @@ Screenshot::Screenshot(QOpenGLWidget *glwidget, QObject *parent)
...
+		for (int i = 0; i < image.width(); ++i)
+		{
+			((uint8_t*)&row[i])[3] = 255;
```
dcwhite: Here is a good place for a comment, or a function call. May I suggest 
```suggestion
			fixAlphaChannel(row[i]);
```

dcwhite: Or whichever pixel value you're fixing there...

dcwhite: then above
```
  static constexpr int ALPHA_INDEX = 3;
  static constexpr int OPAQUE = 255;
  using Rgba8bit = uint8_t*;
static void fixAlphaChannel(QRgb& rgb)
{
  static_cast<Rgba8bit>(&rgb)[ALPHA_INDEX] = OPAQUE;
}
```

dcwhite: And this doesn't change the mac/linux output at all? I'll run it on Windows today

Haydelj: it theoretical shouldn't. It's just setting all pixels to be fully opaque. For some reason (I'm guessing a QT bug) non-opaque pixels cause strange behavior and the QT built-in conversion function that's supposed to throw away the alpha channel seems to use the same broken code and doesn't fix the problem.

dcwhite:
Looks better, I'll test it out today

src/Interface/Modules/Render/Screenshot.cc
```diff
@@ -49,11 +49,21 @@ Screenshot::Screenshot(QOpenGLWidget *glwidget, QObject *parent)
...
-  screenshot_ = viewport_->grabFramebuffer();
+	screenshot_ = getScreenshot();
 }
 
-QImage Screenshot::getScreenshot() {
-  return viewport_->grabFramebuffer();
+QImage Screenshot::getScreenshot()
+{
+  static constexpr int ALPHA_INDEX = 3;
+  static constexpr int OPAQUE = 255;
```
dcwhite:
```suggestion
  static constexpr int OPAQUE_PIXEL_VALUE = 255;
```

src/Interface/Modules/Render/Screenshot.cc
```diff
@@ -49,11 +49,21 @@ Screenshot::Screenshot(QOpenGLWidget *glwidget, QObject *parent)
...
+	QImage image = viewport_->grabFramebuffer();
+	for (int j = 0; j < image.height(); ++j)
+	{
+    auto row = reinterpret_cast<QRgb*>(image.scanLine(j));
+		for (int i = 0; i < image.width(); ++i)
+			reinterpret_cast<uint8_t*>(row + i)[ALPHA_INDEX] = OPAQUE;
```
dcwhite:
```suggestion
			reinterpret_cast<uint8_t*>(row + i)[ALPHA_INDEX] = OPAQUE_PIXEL_VALUE;
```

dcwhite: LOL @Haydelj You'll love this one: OPAQUE is #defined in a Windows header somewhere, so the build fails (and did you really test this on Windows??). So please change the identifier. @jessdtate will need to re-review.

dcwhite: I suppose the Windows build was fine before the constexpr was pulled out. I really must get the github Windows build working...

Haydelj: I wrote the fix on windows but I made your suggested style changes on Linux.

dcwhite: Yeah cool. @ajanson confirmed it fixes Win7, so this should be good to go once the rename is in.

PR #2063, opened 2020-02-09
Superquadric tensor fix

RubioJr9:
- Superquadric glyphs need to change axis alignment if they're more linear than planar.
Before the fix:
superquadric_bug
After the fix:
superquadric_fixed

- Changed superquadric emphasis range from 0-1 to 0-10. We were limitied to very ellipsoidal shapes.
- Changed superquadric emphasis default from 0.85 to 3. The creator used 3 and 6 as examples and said [3 is a a good default choice](http://www.cs.utah.edu/~gk/papers/vissym04/web/slide020.html). 

- I incorrectly named superquadric tensors as superellipsoids so I fixed it. [Superellipsoids](https://en.wikipedia.org/wiki/Superellipsoid) are a subset of [superquadric surfaces](https://en.wikipedia.org/wiki/Superquadrics). I thought superellipsoid was a more specific name, however the creator named it [superquadric tensor glyphs](http://www.cs.utah.edu/~gk/papers/vissym04/web/index.html). I changed the name in ShowFieldGlyphs but left in a line so old files will still display the same.

Edit: The above links to the slides are dead. The paper can be found here http://www.sci.utah.edu/publications/gk04/vissym04kindlmann.pdf


dcwhite:
@RubioJr9 CodeFactor flagged the new method as too complex. I'm going to agree with the failure on this one--it should be refactored to be smaller. There are a lot of duplicated chunks that can be pulled out.

https://github.com/SCIInstitute/SCIRun/blob/92a424a4aa5e73ff7bb98859b64cba41515f3722/src/Graphics/Glyphs/GlyphGeom.cc#L996-L1137

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -1057,8 +1060,9 @@ void GlyphGeom::generateSuperEllipsoid(const Point& center, Tensor& t, double sc
 
   double cl = (eigvals[0] - eigvals[1]) / (eigvals[0] + eigvals[1] + eigvals[2]);
   double cp = 2.0 * (eigvals[1] - eigvals[2]) / (eigvals[0] + eigvals[1] + eigvals[2]);
-  double A = spow((1.0 - cl), emphasis);
-  double B = spow((1.0 - cp), emphasis);
+  bool linear = cl >= cp;
+  double A = linear ? spow((1.0 - cp), emphasis) : spow((1.0 - cl), emphasis);
```
dcwhite: Some more specific critiques.

Don't do computation inside a ternary operator. They should be used sparingly, in simple expression. For this case, pull out the part that actually varies--the cp/cl variable--and put that in a ternary if you must, then do the computation once in code. Above all, try to ascribe to DRY--Don't Repeat Yourself.

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -1077,27 +1081,47 @@ void GlyphGeom::generateSuperEllipsoid(const Point& center, Tensor& t, double sc
...
-        const float nnx = spow(nr[i], 2.0-B) * spow(nx, 2.0-A);
-        const float nny = spow(nr[i], 2.0-B) * spow(ny, 2.0-A);
-        const float nnz = spow(nz[i], 2.0-B);
-        Vector normal = rotate * Vector( nnx, nny, nnz );
+        double nnx, nny, nnz;
+        if(linear)
+        {
+          nnx =  spow(nz[i], 2.0-B);
+          nny = -spow(nr[i], 2.0-B) * spow(ny, 2.0-A);
+          nnz =  spow(nr[i], 2.0-B) * spow(nx, 2.0-A);
```
dcwhite: Similar pattern of computation as lines 1090-1098.

RubioJr9:
> @RubioJr9 CodeFactor flagged the new method as too complex. I'm going to agree with the failure on this one--it should be refactored to be smaller. There are a lot of duplicated chunks that can be pulled out.
> 
> https://github.com/SCIInstitute/SCIRun/blob/92a424a4aa5e73ff7bb98859b64cba41515f3722/src/Graphics/Glyphs/GlyphGeom.cc#L996-L1137

I just pushed a commit. Let me know what you think of the TensorGlyphBuilder class. I reduced the redundancy quite a bit. I'm going to need to be able to evaluate specific points of the superquadric surface so the refactor of that function was going to be needed.

I would also like to do the same and make builders for vectors and scalars. I think the GeomData could be expanded upon to take the function buildObject() from GlyphGeom. That would make GlyphGeom only have addGlyph() functions so that could be renamed to GlyphFactory.

dcwhite:
Neat, CodeFactor is happy. I will look more closely tomorrow!

dcwhite:
Another Windows build error: https://github.com/SCIInstitute/SCIRun/runs/489264304?check_suite_focus=true#step:6:4413

src/Core/GeometryPrimitives/Tensor.cc
```diff
@@ -380,6 +387,26 @@ void Tensor::set_outside_eigens(const Vector &e1, const Vector &e2,
   have_eigens_ = 1;
 }
 
+double Tensor::eigenValueSum()
```
dcwhite: These can all be `const`

src/Core/GeometryPrimitives/Tensor.cc
```diff
@@ -380,6 +387,26 @@ void Tensor::set_outside_eigens(const Vector &e1, const Vector &e2,
...
+{
+  return l1_ + l2_ + l3_;
+}
+
+double Tensor::linearCertainty()
```
dcwhite:
```suggestion
double Tensor::linearCertainty() const
```

src/Core/GeometryPrimitives/Tensor.cc
```diff
@@ -380,6 +387,26 @@ void Tensor::set_outside_eigens(const Vector &e1, const Vector &e2,
...
+{
+  return (l1_ - l2_) / eigenValueSum();
+}
+
+double Tensor::planarCertainty()
```
dcwhite:
```suggestion
double Tensor::planarCertainty() const
```

src/Core/GeometryPrimitives/Tensor.cc
```diff
@@ -380,6 +387,26 @@ void Tensor::set_outside_eigens(const Vector &e1, const Vector &e2,
...
+{
+  return 2.0 * (l2_ - l3_) / eigenValueSum();
+}
+
+double Tensor::sphericalCertainty()
```
dcwhite:
```suggestion
double Tensor::sphericalCertainty() const
```

src/Core/GeometryPrimitives/Tensor.h
```diff
@@ -125,7 +127,12 @@ class SCISHARE Tensor {
   double val(size_t i, size_t j) const { return mat_[i][j]; }
   double& val(size_t i, size_t j) { return mat_[i][j]; }
 
+  double linearCertainty();
```
dcwhite: Add consts here

src/Graphics/Glyphs/GeomData.h
```diff
@@ -0,0 +1,52 @@
...
+#define Graphics_Glyphs_GEOM_DATA_H
+
+#include <Core/GeometryPrimitives/Vector.h>
+#include <Core/Datatypes/Color.h>
+#include <Core/Math/TrigTable.h>
+#include <Graphics/Glyphs/share.h>
+
+namespace SCIRun {
+namespace Graphics {
+struct GeomData
```
dcwhite: Already too many vague names with Geometry and Data...how about `GlyphConstructionParameters`?

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -311,29 +316,29 @@ void GlyphGeom::generateCylinder(const Point& p1, const Point& p2, double radius
...
+    geomData_.points.push_back(r1 * p + Vector(p1));
+    geomData_.colors.push_back(color1);
+    geomData_.normals.push_back(normals);
+    geomData_.numVBOElements++;
+    geomData_.points.push_back(r2 * p + Vector(p2));
+    geomData_.colors.push_back(color2);
+    geomData_.normals.push_back(normals);
+    geomData_.numVBOElements++;
+
+    geomData_.indices.push_back(0 + offset);
```
dcwhite: Now that there's a new class, this block (334-339) can be a function of `GlyphConstructionParameters`. Will again reduce duplication and you can introduce a name to explain the magic numbers.

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -311,29 +316,29 @@ void GlyphGeom::generateCylinder(const Point& p1, const Point& p2, double radius
...
-    normals_.push_back(normals);
-    numVBOElements_++;
-
-    indices_.push_back(0 + offset);
-    indices_.push_back(1 + offset);
-    indices_.push_back(2 + offset);
-    indices_.push_back(2 + offset);
-    indices_.push_back(1 + offset);
-    indices_.push_back(3 + offset);
+    geomData_.points.push_back(r1 * p + Vector(p1));
```
dcwhite: These four lines also make a method--it's the same 4 modifications.

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -454,52 +459,52 @@ void GlyphGeom::generateCylinder(const Point& p1, const Point& p2, double radius
       std::sin(strip_angle * strips) * crx;
     p.normalize();
     Vector normals((length * p + (radius2-radius1)*n).normal());
-    points_.push_back(radius1 * p + Vector(p1));
-    colors_.push_back(color1);
-    normals_.push_back(normals);
-    points_.push_back(radius2 * p + Vector(p2));
-    colors_.push_back(color2);
-    normals_.push_back(normals);
+    geomData_.points.push_back(radius1 * p + Vector(p1));
```
dcwhite: Yeah, these are now method calls on the new class.

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -540,14 +545,14 @@ void GlyphGeom::generateSphere(const Point& center, double radius, int resolutio
         v2 = 1;
       }
 
-      indices_.push_back(0 + offset);
-      indices_.push_back(v1 + offset);
-      indices_.push_back(v2 + offset);
-      indices_.push_back(v2 + offset);
-      indices_.push_back(v1 + offset);
-      indices_.push_back(3 + offset);
+      geomData_.indices.push_back(0 + offset);
```
dcwhite: Another method

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -1152,81 +793,81 @@ void GlyphGeom::generateTorus(const Point& p1, const Point& p2, double major_rad
...
 {
-  points_.push_back(Vector(p1));
-  points_.push_back(Vector(p2));
-  points_.push_back(Vector(p3));
-  points_.push_back(Vector(p4));
-  colors_.push_back(color);
-  colors_.push_back(color);
-  colors_.push_back(color);
-  colors_.push_back(color);
+  geomData_.points.push_back(Vector(p1));
```
```suggestion
  geomData_.addAllPoints({p1, p2, p3, p4});
```

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -1152,81 +793,81 @@ void GlyphGeom::generateTorus(const Point& p1, const Point& p2, double major_rad
...
+  geomData_.points.push_back(Vector(p2));
+  geomData_.points.push_back(Vector(p3));
+  geomData_.points.push_back(Vector(p4));
+  geomData_.colors.push_back(color);
```
```suggestion
  geomData_.addColorNTimes(color, 4);
```

src/Graphics/Glyphs/GeomData.h
RubioJr9: I went with GlyphConstructor and moved buildObject() here

src/Graphics/Glyphs/GlyphGeom.cc
RubioJr9: There's now a function addVertex(point, normal, color) which replaces adding each of those to a vector and also increments numVBOElements

src/Graphics/Glyphs/GlyphGeom.cc
RubioJr9: There's now a function addIndices() which takes 3 index values to form a triangle. Also that offset value can be set and is held internally now.

dcwhite:
@RubioJr9 Merge conflict with this one.

src/Core/GeometryPrimitives/Tensor.h
dcwhite: Still no consts?

src/Graphics/Glyphs/GlyphGeomUtility.h
```diff
@@ -0,0 +1,68 @@
...
+};
+
+inline double GlyphGeomUtility::spow(double e, double x)
+{
+  // This for round off of very small numbers.
+  if (abs(e) < 1.0e-6)
+    e = 0.0;
+
+  if (e < 0.0)
+    return (double)(pow(abs(e), x) * -1.0);
```
dcwhite:
```suggestion
    return static_cast<double>(std::pow(std::abs(e), x) * -1.0);
```

src/Graphics/Glyphs/GlyphGeomUtility.h
```diff
@@ -0,0 +1,68 @@
...
+  else
+    return (double)(pow(e, x));
```
dcwhite:
```suggestion
    return std::pow(e, x);
```
No cast needed here?

src/Core/GeometryPrimitives/Tensor.h
RubioJr9: I swear I added this. I'll do it now

src/Graphics/Glyphs/GlyphGeomUtility.h
RubioJr9: This function was directly copied over from GlyphGeom. I'll add it since it seems like it should be there

dcwhite:
@jessdtate @allywarner Next

dcwhite:
@jessdtate should still review this one, given his tensor knowledge

PR #2068, opened 2020-02-13
Oriented bbox

RubioJr9:
-Added OrientedBBox
-Renamed BBox to AxisAlignedBBox
-Added abstract class BBox
-Added tests to cover OrientedBBox
-Extracted mesh bounding box extending in the get_bounding_box() function so get_oriented_bounding_box can use the same.


src/Core/Algorithms/BrainStimulator/ElectrodeCoilSetupAlgorithm.cc
```diff
@@ -792,7 +792,7 @@ boost::tuple<DenseMatrixHandle, FieldHandle, FieldHandle, VariableHandle> Electr
 
         if (compute_third_output) /// a lot of code is inside this if (make sure it creates proper results)
         {
-          BBox proto_bb = prototype->vmesh()->get_bounding_box();
+          AxisAlignedBBox proto_bb = prototype->vmesh()->get_bounding_box();
```
dcwhite: Here's where `auto` would've been your friend. With it, future return-type changes do not force all these call-site updates.

dcwhite:
@RubioJr9 This one will be a pain to merge with my changes, which will go in first. Just a head(ache)s-up. You might want to work on less intrusive features for another week or so.

dcwhite:
Also, I'm not sure the virtual base `BBox` is a good idea, since there are field algorithms that may depend on it being fast and simple. That'll be a @jessdtate question.

dcwhite:
@RubioJr9 I think you ought to run these big object/class-level changes by me on slack first before going and modifying a lot of files like this. We could've examined some smaller use cases with tests, to show performance is not affected; and peripheral refactoring (a la `auto`) could've made the transition cleaner.

dcwhite:
@RubioJr9 I'll keep this one open for the time being, but please don't add any more PRs branched off of here--they will likely need another redo.

dcwhite:
Another thing to test: field serialization. Changing a type name shouldn't matter, but changing the layout of the object might break all existing field files (note: this would be bad).

RubioJr9:
I don't see how merging this will be hard. The only overlap is the class rename which will be easy to merge. I tested out some test networks that use ReadField to get fields from the test data folder and those work. I'm going to test performance.

RubioJr9:
In that case, I can rename those manually after your merge. It's not a good reason for abandoning a more standardized naming convention.

dcwhite:
I'll probably have you redo this one again, using typedefs for the old bounding box name so not so many files need changing.

PR #2070, opened 2020-02-14
Move VS settings

dcwhite:
Example of pre-refactoring settings code change.


RubioJr9:
Looks good to me! Also, I tested this on Linux

PR #2072, opened 2020-02-19
Custom color map

### LINKED ISSUE #2071: ColorMap infrastructure for user-defined colormaps

Haydelj:
Added custom color map capabilities and refactored some code. Closes #2071


src/Interface/Modules/Visualization/CreateStandardColorMap.ui
```diff
@@ -233,6 +233,65 @@
...
+       <number>2</number>
+      </property>
+      <property name="rightMargin">
+       <number>2</number>
+      </property>
+      <property name="bottomMargin">
+       <number>2</number>
+      </property>
+      <item>
+       <widget class="QPushButton" name="customColorButton0_">
```
allywarner: What do you mean by Custom Color 0 and Custom Color 1? One seems to be grayscale and the other is color? It doesn't seem straightforward to the user

Haydelj: @allywarner Custom Color 0 is the color at the 0th element in the colormap and Custom Color 1 is the value of the 1st element in the colormap. Perhaps Custom Color Min and Custom Color Max could be more straight forward if that doesn't make sense to most people? Also, there should be no such limitation to either of them they use the exact same default color selectors provided by the target OS.

dcwhite:
@jessdtate This is second in line behind #2073.

dcwhite:
@Haydelj Please rebuild with BUILD_TESTING turned on--you broke ColorTests code in ScalarTests.cc. I should add this to the CI build first.

Haydelj:
@dcwhite I made the suggested changes. If we could get this in asap that would be great. I think color maps would be a good first modification to the Opsray renderer and this pr makes significant changes to the colorMap class.

dcwhite:
Cool. I'll check it out today.

dcwhite:
@jessdtate should definitely comment on the usability of the UI as well. This PR is next in the queue.

allywarner:
I didn't see the buttons when I opened the CSCM UI! They turned on and off great! My comments are to maybe have tool tips for what "shift" and "resolution" are doing? I know these are for every colormap, but I don't know what they're doing! :D And for colormap min the color dialog shows up as black (because minimum) but could we have this a little higher than black so we can see the color dialog options? It just looks black in the circle when you select color and for a new user that might feel incorrect.
Screen Shot 2020-03-26 at 11 26 48 AM

dcwhite:
@Haydelj @allywarner I think the general solution to your request is to click the "?" button and be led to the module's doc page (in this case it is inadequate, but that's a separate issue). There's not much room in a tooltip for a good explanation. 

I agree with changing the min/max color defaults to something other than black/white--pick your two favorite colors as defaults :)

allywarner:
@Haydelj @dcwhite I like the documentation idea! ;)

dcwhite:
@jessdtate This one needs your sign off

Issue #2062, opened 2020-02-08
Higher-resolution spheres develop a gap


dcwhite:
Testing spheres in ShowField, I found at resolution 15, 16, and 17, a gap in the two halves appears. It goes away at 18. Scale didn't seem to matter.

![image](https://user-images.githubusercontent.com/1399383/74092084-c7ff0300-4a7c-11ea-8e76-d88b9bf15f20.png)

RubioJr9:
This bug happen with resolution 6 and some other low ones. It happened from looping over double values instead of ints. The error in floating point precision caused the loop to end one step early for certain values.

PR #2073, opened 2020-02-22
Sphere Glyph Fix

RubioJr9:
Ticket #2062

Fixed the gaps by changing the for loops to loop integers instead of doubles. The loop was relying on a <= operator which was no good because of floating point precision.

When you test this, use transparency to verify there's no redundant strips.
I tested all resolutions 3-20.


dcwhite:
@jessdtate After merging the new build stuff into all the PRs, this one gets the distinction of going first. Please review.

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -527,18 +529,13 @@ void GlyphGeom::generateSphere(const Point& center, double radius, int resolutio
...
-      //preserve vertex ordering for double sided rendering
-      int v1 = 1, v2 = 2;
-      if(theta < M_PI)
-      {
-        v1 = 2;
-        v2 = 1;
-      }
+      bool flipVertexOrder = theta_strip >= resolution;
+      int v1 = flipVertexOrder ? 1 : 2;
+      int v2 = flipVertexOrder ? 2 : 1;
```
```suggestion
      int v1 = 2, v2 = 1;
      if (flipVertexOrder)
         std::swap(v1, v2);
```

dcwhite: This reads a bit better to me

dcwhite: But now the boolean seems inverted, since the "original order" is 2,1 and the "flipped" is 1,2--change the name of the boolean?

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -506,17 +506,19 @@ void GlyphGeom::generateCylinder(const Point& p1, const Point& p2, double radius
...
 void GlyphGeom::generateSphere(const Point& center, double radius, int resolution, const ColorRGB& color)
 {
-  double num_strips = resolution;
-  if (num_strips < 0) num_strips = 20.0;
+  if (resolution < 3) resolution = 3;
   double r = radius < 0 ? 1.0 : radius;
   Vector pp1, pp2;
-  double theta_inc = /*2. */ M_PI / num_strips, phi_inc = 0.5 * M_PI / num_strips;
+  double d_res = resolution;
+  double theta_inc = M_PI / (double)resolution, phi_inc = 0.5 * M_PI / (double)resolution;
```
```suggestion
  double theta_inc = M_PI / d_res, phi_inc = 0.5 * M_PI / d_res;
```

dcwhite: That was the purpose of `d_res` right?

RubioJr9: Yeah, thanks for catching that

src/Graphics/Glyphs/GlyphGeom.cc
RubioJr9: Agreed, the name is misleading

dcwhite: Is this getting changed? If so we can probably merge this one later today.

RubioJr9:
Yeah I'll change it

dcwhite:
👍 🥇

dcwhite:
@jessdtate This one is ready for you.

PR #2075, opened 2020-02-25
Widget testing and refactoring, SRInterface refactoring, transform testing

### LINKED ISSUE #2056: Refactor SRInterface (part 3) and Widget library
White board pic coming soon

dcwhite:
Closes #2056 

Also updates the stale github workflow parameters.


dcwhite:
To test: any module that creates widgets. Test all mouse interaction in ViewScene (translate, rotate, scale).
- [ ] GenerateSinglePointProbe
- [ ] GeneratePointSample
- [ ] ShowAndEditDipoles
- [ ] EditMeshBoundingBox

dcwhite:
@allywarner I added a checklist of what to test above.

RubioJr9:
I went through the checklist and I don't have any issues. I like the glyph builder abstraction. What are your thoughts on removing the SphereWidget, DiskWidget, etc. classes and have the create functions in WidgetFactory create a plain widget with the name and position that's given since that's all that is needed of the <Shape>Widget classes?

dcwhite:
That might lose some of the type structure that will come in handy when we get into defining the more complex interactions. It's true those two are mostly the same, but it's not a big deal to keep them around, for now.

dcwhite:
I'll merge this today, but AFTER I make the official X tag (which will happen around 1:30pm).

Haydelj:
@dcwhite @RubioJr9 if you just click on the bounding box (not click and drag) with shift held you can get into a state where it shows red as though it's selected but nothing is selected

RubioJr9:
> @dcwhite @RubioJr9 if you just click on the bounding box (not click and drag) with shift held you can get into a state where it shows red as though it's selected but nothing is selected

I'm seeing this as well with point samples and bbox. It does not happen with the arrowWidget.

dcwhite:
I was seeing something like this too, but wasn't sure what could've caused it. I guess I merged this too soon, but we should be able to figure that out pretty fast.

PR #2089, opened 2020-03-29
OrientedBBox 1: OrientedBBox class with Tests

RubioJr9:
First PR of PR #2068 for EditMeshBoundingBox widget


src/Core/GeometryPrimitives/OrientedBBox.h
```diff
@@ -0,0 +1,119 @@
...
+#include <Core/Utils/Legacy/Assert.h>
+#include <Core/GeometryPrimitives/BBox.h>
+#include <Core/GeometryPrimitives/Point.h>
+#include <Core/GeometryPrimitives/Vector.h>
+#include <Core/GeometryPrimitives/share.h>
+
+namespace SCIRun {
+namespace Core {
+namespace Geometry {
+class OrientedBBox
```
dcwhite:
```suggestion
class SCISHARE OrientedBBox
```

src/Core/GeometryPrimitives/OrientedBBox.h
```diff
@@ -0,0 +1,119 @@
...
+{
+public:
+  OrientedBBox(const Core::Geometry::Vector &e1, const Core::Geometry::Vector &e2, const Core::Geometry::Vector &e3);
+  /// Expand the bounding box to include point p
+  inline void extend(const Point &p)
```
dcwhite: Let's un-inline all of these functions--move them to .cc file

src/Core/GeometryPrimitives/OrientedBBox.h
```diff
@@ -0,0 +1,119 @@
...
+        cmin_[iDim] = projectedPos;
+        cmax_[iDim] = projectedPos;
+      }
+    }
+  }
+
+  /// Extend the bounding box on all sides by a margin
+  /// For example to expand it by a certain epsilon to make
+  /// sure that a lookup will be inside the bounding box
+  inline void extend(double val)
```
dcwhite: ditto

src/Core/GeometryPrimitives/OrientedBBox.h
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+
+  inline bool valid() const
+  { return is_valid_; }
+
+  inline Vector diagonal() const
+  {
+    //TODO: needs invariant check, or refactoring.
+    ASSERT(is_valid_);
```
dcwhite: With only the one constructor, isn't the object always valid? If so, remove this line.

src/Core/GeometryPrimitives/Tests/BBoxTests.cc
```diff
@@ -0,0 +1,76 @@
...
+
+using namespace SCIRun::Core::Geometry;
+
+TEST(BBoxTests, ExtendPerformance)
+{
+  BBox bbox;
+  auto start = std::chrono::steady_clock::now();
+
+  for(int i = 0; i < 1000000; ++i)
+    bbox.extend(Point(rand(), rand(), rand()));
```
dcwhite: CodeFactor complaint:
 Consider using rand_r(...) instead of rand(...) for improved thread safety. lines of code = 1
[Read Up]
Found in src\Core\GeometryPrimitives\Tests\BBoxTests.cc:44
44

    bbox.extend(Point(rand(), rand(), rand()));

src/Core/GeometryPrimitives/Tests/BBoxTests.cc
```diff
@@ -0,0 +1,76 @@
...
+
+  auto end = std::chrono::steady_clock::now();
+  auto diff = end - start;
+  std::cout << "1,000,000 Extend Calls" << " : " << std::chrono::duration<double, std::milli>(diff).count() << " ms\n";
+}
+
+TEST(BBoxTests, Constructor)
+{
+  BBox bbox;
```
dcwhite: This test should have some EXPECTs that describe the default constructor's expected behavior. If members are uninitialized, that should be fixed.

src/Core/GeometryPrimitives/Tests/OrientedBBoxTests.cc
```diff
@@ -0,0 +1,133 @@
...
+using namespace SCIRun::Core::Geometry;
+
+const static double errorOfMargin = 1.0e-10;
+const static Vector x_axis = Vector(1,0,0);
+const static Vector y_axis = Vector(0,1,0);
+const static Vector z_axis = Vector(0,0,1);
+
+TEST(OrientedBBoxTests, Constructor)
+{
+  OrientedBBox obbox(x_axis, y_axis, z_axis);
```
dcwhite: Same here--need some EXPECT statements.

dcwhite:
@RubioJr9 That's why I wanted just one PR at a time...this one needs changes, forcing you to merge with yourself, never fun...

src/Core/GeometryPrimitives/OrientedBBox.h
RubioJr9: The object is not valid on construction

src/Core/GeometryPrimitives/Tests/BBoxTests.cc
RubioJr9: rand_r() requires a seed and these values don't need to be thread safe

dcwhite: Let's use a fixed seed so the test is consistent across platforms. Completely random unit tests are not a good idea.

src/Core/GeometryPrimitives/OrientedBBox.h
dcwhite: OK, same design as original BBox--but don't use a debug assert then, throw an actual exception.

dcwhite: And then test this case in a unit test!

src/Core/GeometryPrimitives/Tests/BBoxTests.cc
RubioJr9: I'll switch to a fixed seed but it's arbitrary for this test case

dcwhite:
Ack, sorry, `rand_r` is not portable, so Windows build is broken again 😢 If you can keep the seed part, but ditch rand_r for rand, then it should all work.

dcwhite:
@jessdtate @allywarner This one is next to review

PR #2091, opened 2020-03-30
Ospray 2 build

dcwhite:
Fixes up external library structure so Ospray2 is usable.


dcwhite:
@jessdtate @allywarner To test this, just build from scratch normally--all should be fine. Then enable WITH_OSPRAY in Superbuild, and all externals should be built, SCIRun starts rebuilding, and then it should fail with Ospray-related code errors, which @Haydelj will pick up with in his Ospray2.0 work.

dcwhite:
This does not handle any other future build errors, which there will be, and for sure doesn't solve the packaging--but Ospray 2.0 does provide a nicer CMake system so hopefully those errors will be easier to figure out.

dcwhite:
@jessdtate @allywarner Finally updated this one--ready to review

Issue #2095, opened 2020-04-06
ViewScene frame flashes out with shift + click


allywarner:
OSX 10.14.6 on SCIRun v5.0-beta.X-183-gfe7ff5217. 
In any ViewScene, the object flashes out when you are holding shift and click about 6 times. This will happen every 6 clicks. Doesn't change how things are working.

dcwhite:
Good catch. This is probably the same flashing phenomenon as with widget selection, but much easier to replicate. I got it with just a gray latvol. @RubioJr9 is currently investigating the exact function this is calling, and hopefully will be able to streamline the code there with some smart geometry id logic. But this shows it's a deeper issue that does not depend on the module feedback mechanism or widgets being in the scene.

RubioJr9:
It seems this issue is separate from the widget selection flickering. I think that one is being caused by the module feedback whereas this one was from a delay in core execution.

jessdtate:
you don't need widgets to see this flicker, but you do have to shift click.

dcwhite:
Fixed with https://github.com/SCIInstitute/SCIRun/pull/2096

PR #2096, opened 2020-04-07
ViewScene Flicker Fix for Non-Widget Selection

RubioJr9:
This fixes issue #2095 about flickering when shift+click on non-widget object or background. 
This is NOT a fix for flickering when widgets are clicked.

-Removed delay in core execution during widget selection


dcwhite:
Can @Haydelj chime in with any knowledge of why there was a 50 ms delay to begin with?

allywarner:
It works when shift+clicking in the background. Testing on ShowAndEditDipole and there is still the flickering, but that is to be expected with this change. I'm also wondering why it was originally set to 50 ms!

Haydelj:
@dcwhite @allywarner I already talked to @RubioJr9 about this but I'll put this here since y'all seem interested. The 50 is not a delay It's the number of ticks(in theory 1 tick = 1ms but not necessarily)  to advance the core by before renormalizing and executing. This clock is what's used to decide which systems to run that way we don't run all of the more expensive systems on the same frame. As for why it was set to 50 maybe whoever wrote it thought the core's clock needed to advance to ensure the render system ran (it doesn't). I agree it should be zero in fact @RubioJr9 said we had already done that on 0 in the edit mesh bounding box PR. That said I'm not entirely sure why setting it to 50 caused a problem. Maybe something to look into if we get a chance

src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -566,7 +566,7 @@ void SRInterface::runGCOnNextExecution()
       updateCamera();
       updateWorldLight();
 
-      mCore.execute(0, 50);
+      mCore.execute(0, 0);
```
dcwhite, line 569: After digging into this a bit more, this is the only hard-coded number case of calling mCore.execute. Instead of passing magic numbers and changing said numbers, make a new function to wrap this special call--give it a name that explains why it's now 0,0, if you like. Something like `executeWithNoDelayToDetermineSelectedWidget`

dcwhite: As discussed today: remove first parameter entirely, then add the new function with expressive name.

dcwhite:
@RubioJr9 There's a not-so-useful unit test that is now not compiling: SCIRun/src/Interface/Modules/Render/Tests/SRInterfaceTests.cc. Not sure why the CI build missed it. Please fix

dcwhite:
Ugh, I know why: there's no workflow that includes both BUILD_TESTING and building the Qt UI. I'll need to add that today.

dcwhite:
@RubioJr9 I took care of that silly compilation error, in my PR.

RubioJr9:
Thanks @dcwhite

PR #2103, opened 2020-04-20
Programmable port prototype--basic looping available. Also updates Windows build to Python 3.8

dcwhite:
* New Windows build script: `windowsBuild.bat`
* Initial implementation of auto-ViewScene-screenshotting from command line
* Other fixes


dcwhite:
Adding interns for review since there's a lot to check here. @Haydelj @RubioJr9 @nids2001 Also because, no one else has looked at it ⏳ @jessdtate @allywarner

src/Core/Datatypes/MetadataObject.h
```diff
@@ -0,0 +1,54 @@
...
+
+namespace SCIRun {
+namespace Core {
+namespace Datatypes {
+
+  class SCISHARE MetadataObject : public Datatype
+  {
+  public:
+    explicit MetadataObject(const std::string& programData);
+    std::string program() const { return programData_; }
```
Sailanarmo, line 43: Just a nit-picky thing, you can also return a reference to `programData_` like:

`const std::string& program() { return programData_; }

dcwhite: Yeah I usually return strings like that, I guess I was typing too quickly here

src/Core/Datatypes/MetadataObject.h
```diff
@@ -0,0 +1,54 @@
...
+    virtual void process(const std::string& modId);
+    std::string dynamic_type_name() const override { return "MetadataObject"; }
+    Datatype* clone() const override;
+  protected:
+    std::string programData_;
```
Sailanarmo, line 48: Why protected instead of private? There is a getter function `program()` Although I don't see a setter anywhere or how `programData_` is eventually initialized.

src/Core/Datatypes/MetadataObject.h
dcwhite: Also, I was thinking the string return type would be a placeholder for whatever this class was going to do (it’s still prototype), so I’ll probably move that method soon anyway.

src/Core/Python/PythonDatatypeConverter.cc
```diff
@@ -93,6 +94,25 @@ namespace
...
+  for (const auto& m : matrices)
+  {
+    if (auto dense = castMatrix::toDense(m))
+      d[i++] = convertMatrixToPython(dense);
+    else if (auto sparse = castMatrix::toSparse(m))
+      d[i++] = convertMatrixToPython(sparse);
+  }
+  for (const auto& f : fields)
+    d[i++] = convertFieldToPython(f);
+  for (const auto& s : strings)
```
Sailanarmo, line 111: Good opportunity to invoke move semantics if you don't need copy of the strings:

`for (auto && s : strings)`

src/Dataflow/Engine/Controller/PythonImpl.cc
```diff
@@ -613,7 +615,7 @@ boost::shared_ptr<PyModule> PythonImpl::findModule(const std::string& id) const
 
 std::string PythonImpl::executeAll(const ExecutableLookup* lookup)
 {
-  cmdFactory_->create(GlobalCommands::DisableViewScenes)->execute();
+  //cmdFactory_->create(GlobalCommands::DisableViewScenes)->execute();
```
Sailanarmo, line 618: Temporary? Or not needed?

src/Dataflow/Network/Module.cc
```diff
@@ -471,6 +483,20 @@ bool Module::executeWithSignals() NOEXCEPT
...
+{
+  auto prog = getOptionalInputAtIndex<MetadataObject>(ProgrammablePortId());
+  if (prog && *prog)
+  {
+    //logCritical("MetadataObject found! {}", id().id_);
+    (*prog)->process(id());
+  }
+  else
+  {
+    //logCritical("\tMetadataObject NOT found! {}", id().id_);
```
Sailanarmo, line 496: Temporary? Not sure if the else statement is needed here unless there is a future use for this.

src/Interface/Application/ModuleProxyWidget.cc
```diff
@@ -581,6 +586,26 @@ QVariant ModuleProxyWidget::itemChange(GraphicsItemChange change, const QVariant
...
+  shape_ << QPointF(-100, 0) << QPointF(0, 70)
+                          << QPointF(100, 0) << QPointF(0, -70)
+                          << QPointF(-100, 0);
+  setPolygon(shape_);
+
+  setFlag(QGraphicsItem::ItemIsMovable, true);
+  setFlag(QGraphicsItem::ItemIsSelectable, false);
+  //setFlag(QGraphicsItem::ItemSendsGeometryChanges, true);
+  setFillRule(Qt::WindingFill);
+  setPen(QPen(Qt::darkGray, 5, Qt::SolidLine));
```
Sailanarmo, line 601: Assuming 5 is the width of the pen?

src/Interface/Application/ModuleProxyWidget.cc
```diff
@@ -581,6 +586,26 @@ QVariant ModuleProxyWidget::itemChange(GraphicsItemChange change, const QVariant
   return ret;
 }
 
+//TODO
+LoopDiamondPolygon::LoopDiamondPolygon(QGraphicsItem* parent) : QGraphicsPolygonItem(parent)
+{
+  shape_ << QPointF(-100, 0) << QPointF(0, 70)
+                          << QPointF(100, 0) << QPointF(0, -70)
+                          << QPointF(-100, 0);
+  setPolygon(shape_);
```
Sailanarmo, line 595: What are these numbers? Could they be made into const variables and belong privately to the class?

src/Interface/Application/ModuleWidget.cc
```diff
@@ -320,17 +325,20 @@ QGroupBox* ModuleWidgetDisplay::getButtonGroup() const
...
   {
     return mod ? mod->id() : ModuleId();
   }
+
+  QString backgroundColorByName(const QString& name)
+  {
+    if (name.contains("Loop"))
+      return moduleRGBA(13, 152, 186);
+
+    return moduleRGBA(99,99,104);
```
Sailanarmo, line 340: Same with these numbers.

src/Interface/Modules/Render/Tests/SRInterfaceTests.cc
```diff
@@ -69,5 +69,5 @@ TEST(SRInterfaceTest,
 {
   SRInterface srinterface;
 
-  srinterface.doFrame(0, 50);
+  srinterface.doFrame(50);
```
Sailanarmo, line 72: What is the 50?

src/Interface/Application/ModuleWidget.h
```diff
@@ -301,13 +307,19 @@ private Q_SLOTS:
...
   void removeOutputPortsFromWidget(int index);
+  void updateProgrammablePorts();
   QHBoxLayout* inputPortLayout_;
   QHBoxLayout* outputPortLayout_;
   bool deleting_;
   static bool networkBeingCleared_;
   const QString defaultBackgroundColor_;
   bool isViewScene_; //TODO: lots of special logic around this case.
 
+  typedef boost::bimap<QString, int> ColorStateLookup;
```
Sailanarmo, line 318: I'm really curious about the bimap and why you chose to use a bimap over `std::unordered_map` or `std::map`?

src/Core/Python/PythonDatatypeConverter.cc
dcwhite: That line by itself won't move anything; the function that uses the `s` value would need to move it, and it doesn't it just reads it and builds a new object, which means an r-value reference/move semantics won't gain anything.

src/Dataflow/Engine/Controller/PythonImpl.cc
dcwhite: This was a bug fix that might not be needed anymore, but needs testing on other major platforms, which I haven't been able to build yet. So it's in the netherworld of "commented out code hopefully soon to be deleted"--which I should use a special symbol for or something.

src/Dataflow/Network/Module.cc
dcwhite: I'll need to change the log level of those print outs. Then the else block can stay.

src/Interface/Application/ModuleProxyWidget.cc
dcwhite: More copied code that needs sprucing up--it's not enabled yet so I'll clean it up later.

src/Interface/Application/ModuleProxyWidget.cc
dcwhite: Ditto

src/Interface/Application/ModuleWidget.cc
dcwhite: Do color RGB parameters need names everywhere? I'm all against magic numbers in general, but it would be overkill to have to write this: `return moduleRGBA(NOT_MUCH_RED, SOME_GREEN, LOTS_OF_BLUE)`

src/Interface/Application/ModuleWidget.h
dcwhite: There is lookup in both directions--otherwise I'd just use a regular map

src/Interface/Application/ModuleWidget.cc
Sailanarmo: I think I was too deep into looking at this. The function clearly say RGBA in it so I think the numbers are self explanatory in this case. It would be overkill.

src/Interface/Modules/Render/Tests/SRInterfaceTests.cc
dcwhite: Now that’s a magic number that we should label. It’s the milliseconds in between frames, more or less. I’ll have the graphics interns change it.

RubioJr9: This test says it renders an empty frame which isn't quite true. The 50 milliseconds as input is the same thing that caused the flicker every 6 clicks. It might work if we change it to 300, which we could just make a function to render an empty frame instead.

dcwhite: I think "empty" here meant nothing was piped into ViewScene in this test, so there would be nothing to show...other than axes I guess.

Haydelj: Since the axis is loaded by the promise fulfillment system we aren't guaranteed it will exist (from my experience it generally won't). What I think this test is doing is just bootstrapping the core. Since there is no geometry the rendering step should be skipped. If this wasn't the case I'd say we should remove this since otherwise, we'd be rendering to a random context since the doFrame code doesn't expect the right context is already bound. It shouldn't matter what the number we execute with is setting it to 0 should be fine especially since we already have a function for that

src/Dataflow/Engine/Controller/PythonImpl.cc
RubioJr9: What about 'TODO DELETE if X works'?

dcwhite: Yeah, there's a lot of commented code that fits that category--lots of work to reference the correct issue though...#lazyprogrammer

RubioJr9:
I just built this on Arch. firstLoop.srn5 gives that invalid bbox exception, but secondLoop.srn5 works without any problems. However, I built this again with headless mode and executing secondLoop.srn5 gives the following errors:

[PYTHON ERROR]   File "<script>", line 3, in <module>
[PYTHON ERROR] name 'scirun_set_module_state' is not defined

and

[PYTHON ERROR]   File "<script>", line 2, in <module>
[PYTHON ERROR] name 'scirun_get_module_input_value_by_index' is not defined

dcwhite:
With your headless build, which python version is used? Can you try `SCIRun -i` and see if you can get a python prompt?

dcwhite:
It might just be missing the API import. There's code to include that automatically, but perhaps it's broken on linux/headless

RubioJr9:
Python 3.5.6. Is there a minimum verison required?

PR #2106, opened 2020-04-27
ViewScene Flicker Fix for Widget Selection and Widget Selection Fix

RubioJr9:
Bug: The other flickering issue listed in #2095
I also fixed a call to addCylinder in GlyphGeom which was stopping ShowAndEditDipoles from executing. This must have happened from my changes to the GlyphGeom class.

The flickering bug was fixed by ViewSceneDialog's and GLWidget's mouse press events being disabled during the module ViewScene's execution.

Instead of a flash, widget selection is being ignored in those instances. To get this to happen, you have to click immediately after releasing a widget. When widget selection is being ignored, mouse movement calls the mouse press event so it will select the widget if there is mouse movement after execution ends.


src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -119,7 +119,7 @@ void GlyphGeom::addCylinder(const Point& p1, const Point& p2, double radius, int
                             const ColorRGB& color1, const ColorRGB& color2,
                             bool renderBase1, bool renderBase2)
 {
-  addCylinder(p1, p2, radius, radius, color1, color2, renderBase1, renderBase2);
+  addCylinder(p1, p2, radius, radius, resolution, color1, color2, renderBase1, renderBase2);
```
dcwhite, line 122: Here you see the problem with multiple many-parameter overloads of the same function. These would be good to convert to builder pattern like with (some of the) widgets and glyphs you started already.

src/Modules/Render/ViewScene.cc
```diff
@@ -233,6 +234,9 @@ void ViewScene::syncMeshComponentFlags(const std::string& connectedModuleId, Mod
 
 void ViewScene::execute()
 {
+  auto state = get_state();
+  state->setValue(IsExecuting, true);
```
dcwhite, line 238: This should change to a scoped object--set state value to true in constructor, then set to false in destructor which ensures it gets called on exit of this function. Since execute might return early, or throw, with such an object you don't need to worry about multiple setValue(false) calls scattered around. See `ag Scoped` for examples.

src/Interface/Modules/Render/GLWidget.h
```diff
@@ -91,6 +92,7 @@ public Q_SLOTS:
   void closeEvent(QCloseEvent *evt);
 
 private:
+  bool canClickMouse_;
```
dcwhite, line 95:
```suggestion
  bool canClickMouse_ {true};
```

src/Interface/Modules/Render/GLWidget.h
```diff
@@ -71,6 +71,7 @@ class GLWidget : public QOpenGLWidget
   void setLockPanning(bool lock);
   void setLockRotation(bool lock);
   void requestFrame() {mFrameRequested = true;}
+  void allowMousePresses(bool allow);
```
dcwhite, line 74:
```suggestion
  void setAllowMousePresses(bool allow);
```

src/Interface/Modules/Render/GLWidget.cc
```diff
@@ -116,10 +116,19 @@ void GLWidget::mouseMoveEvent(QMouseEvent* event)
...
+  {
+    makeCurrent();
+    auto btn = getSpireButton(event);
+    mGraphics->inputMouseDown(event->x(), event->y(), btn);
+    event->ignore();
+  }
+}
+
+//------------------------------------------------------------------------------
+void GLWidget::allowMousePresses(bool allow)
```
dcwhite, line 129:
```suggestion
void GLWidget::setAllowMousePresses(bool allow)
```

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -928,13 +928,21 @@ void ViewSceneDialog::resizingDone()
...
+  if (canSelectWidget())
   {
+    mouseButtonPressed_ = true;
+    mGLWidget->allowMousePresses(false);
     selectObject(event->x(), event->y());
     updateModifiedGeometries();
   }
 }
 
+bool ViewSceneDialog::canSelectWidget()
```
dcwhite, line 940:
```suggestion
bool ViewSceneDialog::canSelectWidget() const
```

src/Interface/Modules/Render/ViewScene.h
```diff
@@ -239,6 +239,7 @@ namespace SCIRun {
       void pushCameraState();
 
       //---------------- Widgets -------------------------------------------------------------------
+      bool canSelectWidget();
```
dcwhite, line 242:
```suggestion
      bool canSelectWidget() const;
```

src/Modules/Render/ViewScene.cc
RubioJr9: If I create a scoped class, is there a way to save it to state so that the interface can see it?

src/Graphics/Glyphs/GlyphGeom.cc
RubioJr9: I converted this one. There's clipping planes which calls addSphere and addCylinder but I think that one should call those functions instead of the builders directly. addSphere doesn't use the builder constructor because it is neither a vector or a tensor. I just left the function generateSphere to do that.

src/Interface/Modules/Render/GLWidget.h
```diff
@@ -93,10 +94,10 @@ public Q_SLOTS:
...
   Render::MouseButton getSpireButton(QMouseEvent* event);
 
-  Render::RendererPtr                   mGraphics          {};  ///< Interface to spire.
-  QTimer*                               mTimer             {};
-  double                                mFrameTime         {0.0};
-  bool                                  mFrameRequested    {false};
+  Render::RendererPtr                   graphics_          {};  ///< Interface to spire.
+  QTimer*                               timer_             {};
+  bool                                  frameRequested_    {false};
+  bool                                  canClickMouse_     {true};
```
RubioJr9, line 100: I changed these from mVarName to varName_ because it is not a part of spire. I deleted mFrameTime because it is not being used anywhere

src/Graphics/Glyphs/GlyphGeom.cc
```diff
@@ -119,7 +119,11 @@ void GlyphGeom::addCylinder(const Point& p1, const Point& p2, double radius, int
                             const ColorRGB& color1, const ColorRGB& color2,
                             bool renderBase1, bool renderBase2)
 {
-  addCylinder(p1, p2, radius, radius, color1, color2, renderBase1, renderBase2);
+  VectorGlyphBuilder builder(p1, p2);
+  builder.setResolution(resolution);
+  builder.setColor1(color1);
+  builder.setColor2(color2);
+  builder.generateCylinder(constructor_, radius, radius, renderBase1, renderBase2);
```
dcwhite, line 126: 👍

src/Modules/Render/ViewScene.cc
```diff
@@ -233,6 +245,9 @@ void ViewScene::syncMeshComponentFlags(const std::string& connectedModuleId, Mod
 
 void ViewScene::execute()
 {
+  auto state = get_state();
+  auto executionReporter = ScopedExecutionReporter(state);
```
dcwhite, line 249: You got it, cool. Aren't those types neat?

src/Modules/Render/ViewScene.h
```diff
@@ -158,6 +159,14 @@ namespace Render {
     void syncMeshComponentFlags(const std::string& connectedModuleId, Dataflow::Networks::ModuleStateHandle state);
 
     ActiveGeometryMap activeGeoms_;
+
+    class ScopedExecutionReporter
```
dcwhite, line 163:
```suggestion
    class SCISHARE ScopedExecutionReporter
```

I'm not sure this is necessary for nested local classes but it wouldn't hurt

src/Modules/Render/ViewScene.cc
RubioJr9: Yeah, I didn't know about it. It's very useful

dcwhite:
@RubioJr9 first of several self-merges

dcwhite:
This should probably be merged and tested along with #2107, since the coloring is still wrong when the widget doesn't move.

RubioJr9:
The widget color doesn't affect the flickering.

dcwhite:
Right, but it’s annoying to test one thing while the other is broken—they should just be tested together.

RubioJr9:
Here's a branch with #2107 merged in https://github.com/RubioJr9/SCIRun/tree/flicker_and_selection_fix

dcwhite:
Check cylinder widgets here

dcwhite:
@jessdtate @RubioJr9 Cylinders in old EMBB widget DO work on this branch.

dcwhite:
However, I'm getting a VS-destructor crash on Windows. Will try to get a stack trace.

dcwhite:
@RubioJr9 This one does need a merge

dcwhite:
Merge looks ugly

RubioJr9:
This is merged now. I fixed some of the widget bugs I was getting with the new ViewScene locking changes. Also, I made it so that when a widget is selected, it is selected between all ViewScene's that also contain that widget. Since we can't directly use those screen space coordinates, we'll have to send transformations from the main ViewScene to the others. I didn't want to tack that on to this PR.

dcwhite:
Cool. I’ll test this next

RubioJr9:
Sounds good. I think that syncing widget movements between ViewScenes would be best to tackle with the arrow widget changes.

dcwhite:
@RubioJr9 Now I'm getting the VS-destructor-crash-after-any-widget-selection crash on Mac too. @jessdtate can you test this one? I'm on 10.14. Basically widgets are unusable unless you always make sure to save the network before loading a new one.

dcwhite:
Funnily enough, with two VSes active showing the same widget, it does not crash. So something fishy is going on with the shared contexts and whatever texture creation stuff is happening in `select`. I still haven't tested the clean up code insertion, that will happen soon.

dcwhite:
I ran the new test network. After translating one widget in one scene (by shift-clicking), any click (without shift) would translate the widget in the other scene. This will be confusing for users, and is probably not what we want. How does one get out of this "shift mode"? If the mouse cursor changed, that would be another story. As it is, the translation can occur in both, but only ever with a shifted-interaction.

RubioJr9:
SCIRun isn't crashing on my machines with widget selection before saving the network and isn't getting stuck in shift mode. An issue that returned is that widgets will briefly snap back to their original position on mouse release. I tested on macOS 10.14 and Linux

dcwhite:
Not before saving, before clearing/exiting. What’s your GPU?

RubioJr9:
I'm not seeing any problems after clearing the network either. Maybe I don't understand the issue fully. I have an integrated chip, Intel Iris 5100.

dcwhite:
It could be a driver problem; all machines I've seen it on have a discrete card. Both NVidia and AMD though, so no narrowing down there. I'll test the latest change for the non-shifting problem; the crash is a separate issue.

dcwhite:
I'm still getting the "shift-mode-active-without-shift-held" problem on Windows with your test network.

RubioJr9:
Have you only seen this on Windows? Are you stuck in shift mode sometimes or always?

dcwhite:
Nope, just happened with latest on the Mac too. And it seems like you need to do several shift-clicks before it gets stuck, it's not immediate.

RubioJr9:
I still can't replicate on 10.14 or linux. Do you have a test network?

dcwhite:
It's any network with widgets

RubioJr9:
So I disabled shiftdown when the window goes out of focus and it fixed the issue you had. I could make a function to share shiftdown status with other ViewScenes with the code that Jacob made, but perhaps we should have a ViewScene manager that handles these interactions.

dcwhite:
Cool, I'll test today. Once these behaviors solidify we can look at refactoring the new VS manager code and centralizing these various interactions. Should I also make a separate issue for the cursor change idea?

RubioJr9:
Yeah, that sounds good. I didn't want to lump in any more stuff into this PR

src/Modules/Render/ViewScene.h
```diff
@@ -158,6 +159,14 @@ namespace Render {
     void syncMeshComponentFlags(const std::string& connectedModuleId, Dataflow::Networks::ModuleStateHandle state);
 
     ActiveGeometryMap activeGeoms_;
+
+    class SCISHARE ScopedExecutionReporter
+    {
+      Dataflow::Networks::ModuleStateHandle state_;
+    public:
+      ScopedExecutionReporter(Dataflow::Networks::ModuleStateHandle state);
```
dcwhite, line 167: Sorry missed these easy improvements earlier
```suggestion
      explicit ScopedExecutionReporter(Dataflow::Networks::ModuleStateHandle state);
```

src/Modules/Render/ViewScene.cc
```diff
@@ -54,6 +54,17 @@ MODULE_INFO_DEF(ViewScene, Render, SCIRun)
 
 Mutex ViewScene::mutex_("ViewScene");
 
+ViewScene::ScopedExecutionReporter::ScopedExecutionReporter(ModuleStateHandle state)
```
dcwhite, line 57:
```suggestion
ViewScene::ScopedExecutionReporter::ScopedExecutionReporter(ModuleStateHandle state) : state_(state)
```

src/Modules/Render/ViewScene.cc
```diff
@@ -54,6 +54,17 @@ MODULE_INFO_DEF(ViewScene, Render, SCIRun)
...
+{
+  state_ = state;
```
dcwhite, line 59:
```suggestion
```

RubioJr9:
This is ready again

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -1197,10 +1223,31 @@ void ViewSceneDialog::keyReleaseEvent(QKeyEvent* event)
...
   case Qt::Key_Shift:
     shiftdown_ = false;
+    updateCursor();
     break;
   }
 }
 
+void ViewSceneDialog::focusOutEvent(QFocusEvent* event)
+{
+  shiftdown_ = false;
```
dcwhite, line 1233: This is buggy--when I focus out, a pointer cursor changes to open hand (which it should never be outside of a VS window). Probably just do this instead:
```suggestion
  setCursor(Qt::ArrowCursor);
```

dcwhite: OK for some reason I forgot the network editor also uses the hand-grab cursor for module selection. It just seemed a bit weird with the transition. Is there a way to customize the color of the hand for ViewScene?

RubioJr9: Focus out only happens when you click away from the window. The [cursor doc](https://doc.qt.io/qt-5/qcursor.html) says that it only supports colored cursor for X11. Mac and Windows don't use X11 for qt apps and Linux usage is going down because it is being replaced by Wayland.

dcwhite: Maybe these should be part of hoverEnter/Leave events then? We’ll discuss today at meeting

Issue #2081, opened 2020-03-06
Shift-click and release widget in ViewScene without moving mouse keeps object red


dcwhite:
@dcwhite @RubioJr9 if you just click on the bounding box (not click and drag) with shift held you can get into a state where it shows red as though it's selected but nothing is selected

_Originally posted by @Haydelj in https://github.com/SCIInstitute/SCIRun/pull/2075#issuecomment-592683343_

dcwhite:
A quick and sloppy fix is to update all geometries on mouseRelease. However this is very inefficient. Also, the color is not reset correctly, a reset widget will turn light gray instead of the previous color (after turning red and not moving).

PR #2065, opened 2020-02-11
Edit camera state variables

RubioJr9:
Converted CameraLookAt and CameraRotation to strings so they can be set through the python interface.

Edit 2/10:
Camera Rotation
-the quaternion is now being normalized so it can no longer be given garbage numbers
-using regex 
-wrong formatting sets to default

Camera Distance
-enforce positive, non-zero numbers

PR #2107, opened 2020-04-29
Fixed Widget Color Restoration

RubioJr9:
Issue #2081

I fixed this by making a class that stores the previous colors and sets them back to those colors on release. To work around having to update all geometries, I appended an increment to every geometry id so we can increment the id when the widget's color needs to be restore.


dcwhite:
Switch color back on release = sounds like another Scoped object :) I haven't looked at the code yet, do you think that would apply? It would span multiple function calls, but that can be solved by putting the scope object on the heap as a member variable.

RubioJr9:
That might work. Are you suggesting we should explicitly destruct the scoped object on release?

dcwhite:
Right, but use a smart pointer, so it would look like `colorResetter_.reset()`

RubioJr9:
Done!

src/Core/Datatypes/Geometry.h
```diff
@@ -57,14 +57,16 @@ namespace Datatypes
     //weird signature--need to avoid raw ptrs and shared_from_this for now
     virtual void addToList(GeometryBaseHandle handle, GeomList& list);
 
-    const std::string& uniqueID() const { return objectName_; }
+    const std::string uniqueID() const { return objectName_ + idIncrement_; }
+    void changeID() { idIncrement_ = std::to_string(std::stoi(idIncrement_)+1); }
```
dcwhite, line 61: @RubioJr9 @Haydelj YANRQ (yet another naive renderer question): for this incrementing of object id, do you think we should use an atomic object? Meaning, could multiple ViewScenes containing the same object introduce a race for incrementing the id? In that case, we'd have undefined ids, which would be bad.

dcwhite: For instance, what if the same widget were in two ViewScenes, and they were linked, and the user clicked one while auto-spinning another, or something like that...

RubioJr9: I think there could be a race condition if a mouse press and release is being done in a linked viewscene to get the widget to move in both viewers, but I doubt Jacob set it up that way because that would fail if the camera wasn't the exact same and handling window sizes would be messy. Regardless, it wouldn't hurt to use an atomic just in case.

Haydelj: Yes, the viewSceneLocking code shouldn't interact with the widget selection code at all

dcwhite:
@jessdtate This one next after #2065

RubioJr9:
Here's a branch with #2106 merged in https://github.com/RubioJr9/SCIRun/tree/flicker_and_selection_fix

dcwhite:
This one works great. Thank goodness for destructors and C++ lifetime determinism. Let's merge it today

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -1354,7 +1381,8 @@ void ViewSceneDialog::selectObject(const int x, const int y)
 
     if (selectedWidget_)
     {
-      colorWidgetRed(selectedWidget_);
+      widgetColorChanger_ = boost::make_shared<ScopedWidgetColorChanger>(selectedWidget_);
```
dcwhite, line 1384: Here's actually a better way to structure the logic: create the scoped object and pass the `selectedWidget_` in all the time. The scoped object will then test for null and change the color if not null. Then the destructor resets the color if the not null again. That can eliminate the if statements in this function, and the need to allocate the scoped object conditionally

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -60,6 +60,61 @@ using namespace SCIRun::Render;
...
 
+namespace SCIRun {
+namespace Gui {
+  class SCISHARE ScopedWidgetColorChanger
+  {
+  public:
+    ScopedWidgetColorChanger(WidgetHandle widget)
+    : widget_(widget)
+    {
+      backupColorValues();
```
dcwhite, line 71:
```suggestion
      if (widget_) { backupColorValues(); colorWidgetRed(); }
```

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -60,6 +60,61 @@ using namespace SCIRun::Render;
...
+    }
+
+    ~ScopedWidgetColorChanger()
+    {
+      colorWidget(previousAmbientColor_, previousDiffuseColor_, previousSpecularColor_);
```
dcwhite, line 76:
```suggestion
      if (widget_) colorWidget(previousAmbientColor_, previousDiffuseColor_, previousSpecularColor_);
```

src/Interface/Modules/Render/ViewScene.cc
RubioJr9: Since there will be other colors added when middle and right click are added from the features in the bounding box widget, I put a switch statement to allow green and blue as well.

dcwhite: ok

dcwhite:
@jessdtate This one next--no merge conflict!

dcwhite:
@jessdtate Ready to merge?

Issue #1890, opened 2019-03-15
Display axis glyph at Cartesian origin


ajanson:
SCIRun 4 had the option to display the axis glyph at 0,0,0 in world space with arrow glyphs that were one unit in length. I would like this option to exist in v5. What is really missing is any representation of the origin in world space to help with scale and positioning in the ViewScene. 

See a similar issue [#1435](https://github.com/SCIInstitute/SCIRun/issues/1435) that places an axis glyph on the screen but not in the world.

jessdtate:
One of the things that can replicate this functionality is to create a field with six points at the origin with the various unit vectors, then use ShowFieldGlyph.

dcwhite:
A module can be written to create this and output it. "CreateAxisField?"

jessdtate:
Yeah, that's a good way to implement it.

dcwhite:
Will discuss at next week's meeting

PR #2082, opened 2020-03-07
View scene lock view

Haydelj:
Lock view prototype. Mostly done but I expect some issues to arise from thorough testing

PR #2108, opened 2020-05-01
Added ShowOrientationAxes module

RubioJr9:
Issue: #1890

I added a module to show orientation axes in world space. I set the default to be at the origin with length 1 for each arrow glyph, but I added an interface to change the position and the scaling.


src/Modules/Visualization/ShowOrientationAxes.cc
```diff
@@ -0,0 +1,182 @@
...
+                                             impl_(new ShowOrientationAxesImpl)
+{
+  INITIALIZE_PORT(InputField);
+  INITIALIZE_PORT(OutputGeom);
+}
+
+void ShowOrientationAxes::setStateDefaults()
+{
+  auto state = get_state();
+  state->setValue(Scale, 1);
```
dcwhite, line 159: I think the state code handles int->double, but it's good practice to have these match type exactly.
```suggestion
  state->setValue(Scale, 1.0);
```

src/Modules/Visualization/ShowOrientationAxes.cc
```diff
@@ -0,0 +1,182 @@
...
+  state->setValue(X, 0);
```
dcwhite, line 160:
```suggestion
  state->setValue(X, 0.0);
```
same for Y and Z

dcwhite:
Next for @jessdtate--use the pre-built version

dcwhite:
@RubioJr9 Any new module needs at least one test network! Please add one so I can test it.

RubioJr9:
I added a test network but with the changes from the looping modules, it added a required field input port to this module that isn't used.

dcwhite:
There's a new optional metadata port on every module, it should not have added a field port though.

dcwhite:
Also, the unit test is failing on the module count. Time to up the number. Go to https://github.com/SCIInstitute/SCIRun/blob/master/src/Modules/Factory/Tests/ModuleReplaceBuilderTests.cc and bump both numbers on L53-55 by 5.

src/Modules/Visualization/ShowOrientationAxes.cc
```diff
@@ -0,0 +1,182 @@
...
+  auto state = get_state();
+  state->setValue(Scale, 1.0);
+  state->setValue(X, 0.0);
+  state->setValue(Y, 0.0);
+  state->setValue(Z, 0.0);
+}
+
+void ShowOrientationAxes::execute()
+{
+  auto input = getRequiredInput(InputField);
```
dcwhite, line 167: @RubioJr9 Here's the required input port...did you merge this funny or something? Or forget about it...

dcwhite: Or maybe just leftover copy-paste...this is why the test network is useful 😄

RubioJr9: I have no idea where that came from. I tested it before the merge and I wasn't getting any errors about a required input port.

dcwhite: That field input port is also in the module header file, is it supposed to be an optional input? Seems unlikely it would appear in both files by mistake.

RubioJr9: Yeah I must have copied it over by accident. I'll remove it now

dcwhite:
@jessdtate Next while @RubioJr9 merges #2082 into his other ones

dcwhite:
@RubioJr9 This one has the messed-up merge-master button, can you update your branch manually?

RubioJr9:
> @RubioJr9 This one has the messed-up merge-master button, can you update your branch manually?

Mine had the button and merged without conflicts

dcwhite:
Thanks

RubioJr9:
No problem

allywarner:
@RubioJr9 @dcwhite What do you think about a "reset" button to reset the X, Y, Z, and scale back to what you started with? I was also thinking of a "fit scale to object" or something. I tested this module on a skull segmentation and the orientation widget was inside the skull and too small to see. Of course I can change the scale and could see where it was, but it might be nice to have a quick button to push to have that happen? Or just have this calculated as the origin is calculated?

dcwhite:
I thought of something similar. Reset to origin should be easy, auto-scale would need screen space data which the module won’t have.

RubioJr9:
I'll add a reset button. We could give a field input to the module, take the bounding box of the field, and scale the orientation glyph based on that.

allywarner:
Whatever is easier! It's really easy to use now which is great!

RubioJr9:
I can do both if it sounds useful. It'll be pretty straightforward to implement.

allywarner:
I think it'd be useful!

dcwhite:
@RubioJr9 Do you want to add the buttons as part of this PR, or a later one? Either one is fine with me

RubioJr9:
I'll just add it to this one. It shouldn't take too long

RubioJr9:
@allywarner If I base the size of orientation axes based on a field input, should I scale it to the shortest or longest axis?(as in x, y, z) I could also scale each one individually but the axes won't be equal sizes.

dcwhite:
@RubioJr9 Are you building each of these recent changes? Every build is red. Just wondering...

Also, just scale to the smallest dimension/size, to avoid large arrows sticking out weirdly. And it should be a checkbox option, "scale to input mesh"

RubioJr9:
I think I forgot to stage an include statement. I'll have this fixed in a push in the next couple of hours.

dcwhite:
OK cool. Happens to everybody, but be extra careful on "live PR" branches.

RubioJr9:
Sorry, I'll be more careful. Let me know what you think of the new UI.

allywarner:
I like it! Thank you!

dcwhite:
* Does the scale checkbox invalidate the scale doublespinbox as far as scaling? Or can they both be active at the same time? If not, then checking the box should disable the spin box.
* Same with position: if the field position invalidates the 3 position spinboxes, disable them when the checkbox is checked. Also, they should update to show what the current position actually is when set via the field data.

src/Interface/Modules/Visualization/ShowOrientationAxesDialog.h
```diff
@@ -0,0 +1,55 @@
...
+   DEALINGS IN THE SOFTWARE.
+*/
+
+
+#ifndef INTERFACE_MODULES_VISUALIZATION_SHOW_ORIENTATION_AXES_DIALOG_H
+#define INTERFACE_MODULES_VISUALIZATION_SHOW_ORIENTATION_AXES_DIALOG_H
+
+#include "Interface/Modules/Visualization/ui_ShowOrientationAxes.h"
+#include <Interface/Modules/Base/ModuleDialogGeneric.h>
+#include <Interface/Modules/Visualization/share.h>
```
Sailanarmo, line 29: Just a nit-picky thing. Should stick to one standard when including files.

Sailanarmo:
@dcwhite I'm not seeing where you are requesting a change. Where exactly is that? I may have missed it.

src/Interface/Modules/Visualization/ShowOrientationAxesDialog.h
dcwhite: That's SCIRun's standard--the quoted includes are only used on generated Qt ui_ headers. Everything else should be in angle brackets. If you see other counterexamples, please do pick at them.

dcwhite:
I made GUI design questions in the comment above.

dcwhite:
https://github.com/SCIInstitute/SCIRun/pull/2108#issuecomment-640936830

RubioJr9:
The scale factor stacks on top of the field scale if there is field input. I'll make changes next to gray out the position and update it.

dcwhite:
The scale stacking is fine I suppose, is there a way to indicate that in the GUI? Any ideas @jessdtate?

RubioJr9:
What about a checkbox for applying the scale factor from the spinbox?

dcwhite:
Add more cases to test network:
- [ ] Default case
- [ ] Adjusted case without input field
- [ ] field cases: 2 different to show different position/scaling options

RubioJr9:
image
I put another checkbox next to the scale factor for disabling that value. The position spin boxes are now disabled when the input field is used. Also, the checkboxes and buttons are set to execute the module.

allywarner:
Being a bit nit-picky here, but would you mind making the vertical spacer smaller? Group boxes would be cool too, but that requires a little more work. The spacer should be quick!

RubioJr9:
> Being a bit nit-picky here, but would you mind making the vertical spacer smaller? Group boxes would be cool too, but that requires a little more work. The spacer should be quick!

Good idea. I'll do that now

PR #2113, opened 2020-05-13
Clang-Format Code Styling

RubioJr9:
This format file allows configuring the code style across the project. It is made by the LLVM group that made Clang. It is supported by most IDE's(emacs, vim, xcode, visual studio, atom) with an extension. This also allows IDE's to auto-format the code. I set it up to follow SCIRun's current code style as closely as I could.

NOTE: I set it to sort include statements, but added a rule that searches for includes that contain 'share.h' and always lists them at the bottom(Less broken Windows builds!).

This [document](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) explains the configuration of every setting I used. You can load the .clang-format file into this [configurator](https://zed0.co.uk/clang-format-configurator/) and see what the changes look like. However, the brace wrapping and share.h priority doesn't work in this tool.


dcwhite:
Super cool! Can it be fed into the clang-tidy GitHub action? I was just looking at that one today

RubioJr9:
> Super cool! Can it be fed into the clang-tidy GitHub action? I was just looking at that one today

Yup! On the clang-tidy documentation, it says we can set the flag --format-style='file'  (literally 'file', not a placeholder), and it will find the .clang-format file in the closest parent folder.

Feel free to make any changes to the format file.

Also, clang-tidy has a feature that does a [project-wide refactor](https://www.kdab.com/clang-tidy-part-1-modernize-source-code-using-c11c14/) if you have a compilation database, which I did manage to create with bear to get IDE features. Though it'd be nice to get the compilation database made without bear because it is not available on Windows and requires turning off System Integrity Protection in Mac OS. You also have to stop the build after externals are finished, go to the bin/SCIRun folder, then restart with bear. Do you know if there's a way to create an optional cmake flag to set for building in bin/SCIRun?

dcwhite:
So a flag to run make from bin, but only build bin/SCIRun, skipping externals? It sounds doable, but could get messy. Could you just `make SCIRun_external` from bin instead?

RubioJr9:
Not quite. There is a flag called CMAKE_EXPORT_COMPILE_COMMANDS that will generate the compilation database that'd be a better solution than bear. The problem is that I would need to set that for building in bin/SCIRun. I've tried enabling this flag in bin/ and it creates an empty json file.

dcwhite:
I'd love to help with this, but this seems like a crossover point into something that you need to figure out, since it's quite linux-specific and won't be supported officially anyway (unless it makes it into an CMake option). If we can have it running as part of the Github action to check new files, that will have to do for now.

RubioJr9:
The way I have it working right now with bear is Mac/Linux specific, however that flag for exporting compile commands is [built into cmake](https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html) and should work on all of the platforms. I don't think this will be useful for Github Actions, but it would simplify setting up IDE features, make it work on Windows, and require less permissions on Macs. I just thought this would be a good place to discuss this since it is about clang tools.

dcwhite:
Ah, looks like this can do automated modernization like auto and usings. I'm sold (there are several issues for changing the code for C++11), but it's gotta wait until we get the PR list down.

RubioJr9:
I agree, we should wait until there's not a lot of changes waiting. It should quiet down when Jacob starts his other internship. I managed to get the compile commands flag working through CMake. The compilation database is ~27x larger than the one made with bear so it probably has a more detailed build log. I'm going to make a flag to disable/enable it through the Superbuild.

PR #2115, opened 2020-05-15
Quick merge!

dcwhite:
@jessdtate Here are BBox fixes for various places, including ShowString.

The ReadFile code is work in progress, but can be merged now, no effect.


dcwhite:
👍 going in today

PR #2118, opened 2020-05-15
Compilation Database

RubioJr9:
-Added a flag called GENERATE_COMPILATION_DATABASE that enables or disables the cmake option CMAKE_EXPORT_COMPILE_COMMANDS in the src/ folder.
I left the flag on as default because there are no slowdowns in build times. There was a ~10% speedup which I assume is within the margin of error. The file only takes up 3.5Mb. To get this to work on existing builds, you have to call cmake again in bin/, delete bin/SCIRun, and compile again. After that, the compilation database will update on recompilation.

-Added a symbolic link in the base folder so extensions can find the compilation database automatically.

Edit: Removed qt5 detection


dcwhite:
@RubioJr9 Mac CI build is broken...couldn't find Qt 😆 😢

dcwhite:
This will also break the Windows build script I just made 😭

dcwhite:
@RubioJr9 Breaks windows packaging too--can you change this to Linux only, and off by default? You can keep it on for your fork if you want.

dcwhite:
To fix the other platforms, you'll have to add back the original Qt5_PATH variable.

RubioJr9:
I'm waiting for my mac build to finish, but it seems we can use the same directory for Qt5_PATH for CMAKE_PREFIX_PATH so changing the github builds might be pretty simple. The cmake one is built in, but the Qt5_PATH name is more descriptive so I can go either way.

RubioJr9:
It didn't work out on macOS so I'm going to make it linux only and see if somebody on another distribution like Ubuntu can try it.

RubioJr9:
I removed the auto detection of Qt5 altogether since it wasn't working on Ubuntu, as well as Mac and Windows. I did consolidate the qt5 find_package statements into 1 line so I left that in there.

dcwhite:
This looks fine. I'll test it this week.

dcwhite:
@jessdtate This one is ready for quick review too. Maybe worth a build, just to see if there is any impact.

jessdtate:
I built it without problems

PR #2121, opened 2020-05-20
Tensors: Changed Eigen Solver and Added functionality for Uncertainty Glyphs

RubioJr9:
I changed the eigen solver for tensors because the current function solved them did not maintain the orientation of given eigenvectors. To be able to implement uncertainty glyphs in my next pull request, I needed these to be consistent when 2 eigenvalues are equivalent.

Note: These images are made with the Regression/Render/negativeTensors.srn5 network

Old RGB conversion that relied on primary eigenvector's direction. As you can see in the image, the primary eigenvector may or may not be axis aligned depending on which eigenvalues are negative.
oldNegativeTensors

New RGB conversion that checks if there are equivalent eigenvalues. 
newNegativeTensors

I also added some functionality to the tensor class and tensor glyph builder so I can use it to create the uncertainty tensor glyph.


src/Core/GeometryPrimitives/Tensor.cc
```diff
@@ -334,29 +339,65 @@ Vector Tensor::operator*(const Vector& v) const
...
-
-  tenEigensolve_f(eval, evec, ten);
-
-  e1_ = Vector(evec[0], evec[1], evec[2]);
-  e2_ = Vector(evec[3], evec[4], evec[5]);
-  e3_ = Vector(evec[6], evec[7], evec[8]);
-  l1_ = eval[0];
-  l2_ = eval[1];
-  l3_ = eval[2];
+  DenseMatrix dm = DenseMatrix(3, 3);
```
dcwhite, line 358:
```suggestion
  Eigen::Matrix3d dm;
```
Go ahead and use Eigen's fixed size matrices here for speed.

src/Core/GeometryPrimitives/Tensor.cc
```diff
@@ -334,29 +339,65 @@ Vector Tensor::operator*(const Vector& v) const
...
+  for (int i = 0; i < 3; ++i)
+    for (int j = 0; j < 3; ++j)
+      dm(i,j) = mat_[i][j];
```
dcwhite, line 361:
```suggestion
  std::copy(mat_, mat_ + SIZE_, dm.data());
```
define DIM_ = 3 and SIZE_ = DIM_ * DIM_ as static class contants, or anonymous enum values

src/Core/GeometryPrimitives/Tensor.cc
```diff
@@ -334,29 +339,65 @@ Vector Tensor::operator*(const Vector& v) const
...
   have_eigens_ = 1;
+  reorderTensorValues();
+}
+
+void Tensor::reorderTensorValues()
+{
+  if (!have_eigens_) return;
+  std::vector<Vector> eigvecs = { e1_, e2_, e3_ };
+  std::vector<double> eigvals = { l1_, l2_, l3_ };
+  std::vector<std::pair<double, Vector>> sortList(3);
```
dcwhite, line 382:
```suggestion
  std::vector<std::pair<double, Vector>> sortList = { {e1_, l1_}, {e2_, l2_}, {e3_, l3_} };
```
spell it out with just 3, yeah?

dcwhite, line 382:
```suggestion
std::map<double, Vector> sorted = { {e1_, l1_}, {e2_, l2_}, {e3_, l3_} };
```
Or why not just make a map, which will sort in the constructor

src/Core/GeometryPrimitives/Tensor.cc
```diff
@@ -334,29 +339,65 @@ Vector Tensor::operator*(const Vector& v) const
...
+    eigvals[d] = sortList[d].first;
+    eigvecs[d] = sortList[d].second;
+  }
+
+  e1_ = eigvecs[0];
+  e2_ = eigvecs[1];
+  e3_ = eigvecs[2];
+  l1_ = eigvals[0];
+  l2_ = eigvals[1];
+  l3_ = eigvals[2];
```
dcwhite, line 400:
```suggestion
auto sortedEigsIter = sorted.begin();
std::tie(e1_, l1_) = *sortedEigsIter++;
std::tie(e2_, l2_) = *sortedEigsIter++;
std::tie(e3_, l3_) = *sortedEigsIter++;
```
17 lines into 4

src/Core/GeometryPrimitives/Tensor.h
```diff
@@ -45,6 +45,7 @@
 #include <Core/Containers/Array1.h>
 #include <Core/GeometryPrimitives/Vector.h>
 #include <Core/GeometryPrimitives/share.h>
+#include <Core/Datatypes/DenseColumnMatrix.h>
```
dcwhite, line 48: This creates a circular dependency of libraries. You'll need to implement the Tensor stuff that depends on DenseMatrix math in a third library, or return lower-level objects, like a `vector<double>`. Looking at the implementation of `mandel`, there is no Eigen usage, it's just a container you need. You can use Eigen like above, but only internally.

src/Graphics/Glyphs/TensorGlyphBuilder.h
```diff
@@ -40,37 +40,66 @@
...
+  double getA();
+  double getB();
+  double computeSinPhi(int v);
+  double computeCosPhi(int v);
+  double computeSinTheta(int u);
+  double computeCosTheta(int u);
+  void setTensor(const Core::Geometry::Tensor &t);
+  Core::Geometry::Tensor getTensor() const;
+  std::vector<Core::Geometry::Vector> getEigenVectors();
+  std::vector<double> getEigenValues();
```
dcwhite, line 97: Class is getting too big, but I'll let you split it up later.

dcwhite:
Yeah the build is totally broke--note circular dependency note above. I'll convert this to draft until it's fixed.

src/Core/GeometryPrimitives/Tensor.cc
RubioJr9: neat

src/Core/GeometryPrimitives/Tensor.h
RubioJr9: In the module I'll be using it in, I'll be needing it stored as a Eigen vector to do an outer product to make it a 6x6 matrix. I was able to get it building by including <Eigen/Dense> directly. Is this bad practice?

src/Graphics/Glyphs/TensorGlyphBuilder.h
RubioJr9: Okay, I'll address this with in the next PR with the new module.

src/Core/GeometryPrimitives/Tensor.cc
RubioJr9: The copy function didn't work because the array is 2d. I instead replaced it by row so there is one for loop.

dcwhite: Ah right. There’s a place for a generic flattening operation here but I suppose that could be overkill

src/Core/GeometryPrimitives/Tensor.h
dcwhite: Including eigen directly is ok from a build perspective since it’s header only, like glm. But then intertwining type signatures with eigen’s types leads to the same problem I wrote that long post about related to returning glm types from renderer classes, find it in slack. It sounds like you need some intermediate types in a third library that can safely depend on both geometryprimitives and datatypes. Why not just make a new library called TensorVisualization, with classes for doing all these new things with tensors? That will keep the original tensor class from getting too bloated (like the field code), and the dependencies clearer.

dcwhite:
Another idea is to just move the Tensor class somewhere where it can be implemented in terms of an Eigen matrix—lots of repetitive math here. But that will break serialization, so we’d need to also write a Pio function that converts it back. We can discuss options Friday

src/Core/GeometryPrimitives/Tensor.h
RubioJr9: This mandel function only changes the notation. I'll move the functions(symmetricTensorFromNineElementArray, symmetricTensorFromMandel, symmetricTensorFromSixElementArray) that convert tensors to and from other types into another class called TensorConversion. With that, it'd still leave one usage of Eigen that I added in build_eigens_from_mat so I could use Eigen's eigen solver. The only other functions I added were a division operator and one to reorder tensor values.

dcwhite: Wouldnt it be nice to start from scratch though? With a new tensor class you can do whatever you want with it without worrying about backward compatibility or dependency management. This is a more clear-cut case than the BBox addition. It will also greatly speed up your development if you don't have to recompile all the field code every time.

RubioJr9: Recompiling field classes doesn't take too long on 24 threads, but it'd be nice to make a tensor class from scratch. Eigen has a tensor class that's a wrapper around the matrix class. It's listed in their (unsupported modules)[https://eigen.tuxfamily.org/dox/unsupported/eigen_tensors.html] however. (This)[http://eigen.tuxfamily.org/index.php?title=Tensor_support] is a list of their features.

RubioJr9: One problem is that I do need the tensor fields to use the new eigen solver in Eigen, unless I wrote an eigen solver from scratch.

dcwhite: OK how about this:
new class Tensor2/NewTensor/BetterTensor/AwesomeTensor/TarkTensor inherits from Eigen::Tensor<double,3>, lives in Core_Datatypes_Matrix to centralize the Eigen dependency.

We can add field support easily, that doesn't save the recompile cost though. Or we just add conversion functions between old and new, with care for the IO functions.

dcwhite: Then old Tensor class is closed, nothing to add there; only add new stuff to new class.

RubioJr9: Let's name it SymmetricTensor since that's all we deal with in SCIRun. The documentation says they have optimizations for symmetry so that's a plus. I'd rather not do the conversion from old tensor to new tensor because it'll have to happen every time the new tensor is needed. I'd also have to convert back to make a field. What do you have in mind with the field support?

dcwhite: We must have to add it as a field data type like the recent complex fields. But we might want to replace it, and come up with some code to handle the IO part separately.

RubioJr9: Are you suggesting that we have 2 different field types that are Tensor and NewTensor?

src/Core/GeometryPrimitives/Tensor.cc
```diff
@@ -334,29 +336,49 @@ Vector Tensor::operator*(const Vector& v) const
 		v.x()*mat_[2][0]+v.y()*mat_[2][1]+v.z()*mat_[2][2]);
 }
 
+Tensor Tensor::operator/(const double s) const
```
dcwhite, line 339:
```suggestion
//TODO: move to new tensor class soon
Tensor Tensor::operator/(const double s) const
```

src/Core/GeometryPrimitives/Tensor.cc
```diff
@@ -385,6 +426,30 @@ void Tensor::set_outside_eigens(const Vector &e1, const Vector &e2,
   e1_ = e1; e2_ = e2; e3_ = e3;
   l1_ = v1; l2_ = v2; l3_ = v3;
   have_eigens_ = 1;
+  reorderTensorValues();
+  build_mat_from_eigens();
+}
+
+Eigen::MatrixXd Tensor::mandel()
```
dcwhite, line 433:
```suggestion
//TODO: move to new Tensor class soon
Eigen::MatrixXd Tensor::mandel()
```

src/Core/GeometryPrimitives/Tensor.h
```diff
@@ -45,6 +45,7 @@
 #include <Core/Containers/Array1.h>
 #include <Core/GeometryPrimitives/Vector.h>
 #include <Core/GeometryPrimitives/share.h>
+#include <Eigen/Dense>
```
dcwhite, line 48:
```suggestion
//TODO: remove when new Tensor class arrives
#include <Eigen/Dense>
```

src/Core/GeometryPrimitives/Tensor.h
```diff
@@ -179,6 +185,23 @@ Tensor symmetricTensorFromSixElementArray(const Indexable& array)
     );
 }
 
+template <typename Indexable>
```
dcwhite, line 188:
```suggestion
//TODO: move to new Tensor class soon
template <typename Indexable>
```

src/Graphics/Glyphs/TensorGlyphBuilder.h
```diff
@@ -40,37 +40,66 @@
 
 namespace SCIRun {
 namespace Graphics {
+struct EllipsoidPointParams
```
dcwhite, line 43:
```suggestion
struct SCISHARE EllipsoidPointParams
```

src/Graphics/Glyphs/TensorGlyphBuilder.h
```diff
@@ -40,37 +40,66 @@
...
+{
+  double sinPhi;
+  double cosPhi;
+  double sinTheta;
+  double cosTheta;
+};
+
+struct SuperquadricPointParams
```
dcwhite, line 51:
```suggestion
struct SCISHARE SuperquadricPointParams
```

src/Graphics/Glyphs/TensorGlyphBuilder.h
```diff
@@ -40,37 +40,66 @@
...
+};
+
 class SCISHARE TensorGlyphBuilder
 {
 public:
   TensorGlyphBuilder(const Core::Geometry::Tensor& t, const Core::Geometry::Point& center);
   void scaleTensor(double scale);
   void reorderTensorValues(std::vector<Core::Geometry::Vector>& eigvecs, std::vector<double>& eigvals);
-  void makeTensorPositive();
+  void makeTensorPositive(bool reorder = false, bool makeGlyph = true);
```
dcwhite, line 67: Not a big fan of multiple default parameters

src/Modules/Visualization/ShowFieldGlyphsPortHandler.cc
```diff
@@ -490,35 +490,31 @@ namespace SCIRun{
...
-            colorVector = zCross;
-          }
-          else{
-            colorVector = t.get_eigenvector1();
-          }
-        } else{
-          colorVector = t.get_eigenvector1();
+        std::vector<double> eigvals(3);
+        std::vector<Geometry::Vector> eigvecs(3);
+        t.get_eigenvectors(eigvecs[0], eigvecs[1], eigvecs[2]);
```
dcwhite, line 498: Can the getters be rewritten simply to return a vector of 3 values/vectors? How much code would need to change?

src/Modules/Visualization/ShowFieldGlyphsPortHandler.cc
```diff
@@ -490,35 +490,31 @@ namespace SCIRun{
...
+        if (std::abs(eigvals[0] - eigvals[1]) < epsilon)
+        {
+          if (std::abs(eigvals[1] - eigvals[2]) < epsilon)
+            colorVector = (eigvecs[0] + eigvecs[1] + eigvecs[2]).normal();
+          else
+            colorVector = (eigvecs[0] + eigvecs[1]).normal();
         }
+        else
+          colorVector = eigvecs[0];
+
```
dcwhite, line 517: This looks like the key block of code, it would be nice to have a separate named function for it

src/Modules/Visualization/ShowFieldGlyphsPortHandler.cc
dcwhite: Or return a 3-tuple, and then use std::tie

dcwhite: if we can move to C++17/20, we could also do structured bindings:
`auto [eig0, eig1, eig3] = t.get_eigenvalues();`

RubioJr9:
Is it okay if I close this and resubmit it with the new Tensor class implemented?

dcwhite:
Let's keep it open until you have the new one ready as a PR. Then reference this one as closable.

RubioJr9:
Sounds good. I'll make sure to address all of the changes from here.

PR #2123, opened 2020-05-22
Glm Upgrade from 0.9.4.6 to 0.9.9.8

RubioJr9:
-Updated glm
-Converted to new vec_swizzle functions
-Added default constructor value for mat4's
-Changed glm::float_t to regular float

What works:
-Camera movements before and after execution
-Rendering objects
-Point widgets

I was only able to test translation widgets since the fix for dipoles is not in master yet. I'll test this when that is merged. @Haydelj do you know of any other uses of glm I should test?


dcwhite:
This will come after all the rest

src/Externals/spire/arc-look-at/ArcLookAt.cpp
```diff
@@ -103,10 +104,12 @@ void ArcLookAt::doZoom(const glm::vec2& ssPos)
 //------------------------------------------------------------------------------
 glm::mat4 ArcLookAt::getWorldViewTransform() const
 {
-  glm::mat4 finalTrafo;
-  finalTrafo[3].xyz() = -mCamLookAt;
-  finalTrafo          = mArcBall->getTransformation() * finalTrafo;
-  finalTrafo[3][2]   += -mCamDistance;
+  glm::mat4 finalTrafo(1.0);
+  for (int i = 0; i < mCamLookAt.length(); ++i)
+    finalTrafo[3][i] = -mCamLookAt[i];
```
Sailanarmo, line 109: I'm trying to get better with this magic number thing. However, I notice here on line 109 and line 112, the index [3] is used. Does this have some sort of special significance? And if it does can this number be given a name?

src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -1001,18 +1002,18 @@ uint32_t SRInterface::getSelectIDForName(const std::string& name)
...
         trans_bb = temp * trans_bb;
-        temp = glm::translate(glm::mat4(), center_bb);
+        temp = glm::translate(glm::mat4(1.0f), center_bb);
         trans_bb = temp * trans_bb;
         int index = 0;
         for (auto i : clippingPlanes_)
         {
           glm::vec3 n3(i.x, i.y, i.z);
-          double d = i.d;
+          float d = i.d;
```
Sailanarmo, line 1016: is clippingPlanes an Object? if that is the case, I recommend passing this in by `const &` ex, `for(const auto & i : clippingPlanes_)`. Unless the object is really small, however I also do not see anything that is changing the object itself. So I would suggest changing this to `const &`.

src/Externals/spire/arc-look-at/ArcLookAt.cpp
dcwhite: It seems in the graphics code 0/1/2/3 mapping to x/y/z/w coordinates is common, although there's no reason to use the raw integers everywhere. Just define a top-level enum so it looks like `finalTrafo[Coordinate.w][i] = ...` although that's a bit wordy

Sailanarmo: ah, I see. yeah, even a `Coord.w` would be a lot more clear. But if it isn't necessary right now then I wouldn't worry about it too much.

RubioJr9: The x/y/z/w convention is a bit strange for a homogeneous transformation. The literature typically uses numbers for indexing. In this case, finalTrafo[3] is the translational component of the transformation. We could do Coord.x/y/z/trans, but the bottom row is a special case so that convention could get confusing as well.

src/Interface/Modules/Render/ES/SRInterface.cc
RubioJr9: Thanks, good catch

dcwhite:
@jessdtate This one is ready to go. It should be built from scratch and then test your busiest ViewScene network.

jessdtate:
I don't have camera mouse controls with this build.  mostly rotation, but sometimes I lose all of them.  I haven't got rotation working at all.

jessdtate:
auto-rotate controls don't work either, but I can change the camera with that python console

dcwhite:
@RubioJr9 Yeah, camera is broken on Mac. I assume all your testing was on Linux? Can you do a build on your other machines?

dcwhite:
Widget drag works, camera pan and zoom are fine. Probably a bug in the creation of the rotation matrix, with the identity matrix construction change.

dcwhite:
Tested on Windows, and rotation is "off"--click-drag does execute a rotate, but the scene jumps a bit before rotating smoothly, indicating the rotation matrix is off somehow.

src/Externals/spire/arc-look-at/ArcLookAt.cpp
```diff
@@ -103,10 +104,12 @@ void ArcLookAt::doZoom(const glm::vec2& ssPos)
...
+
+  finalTrafo = mArcBall->getTransformation() * finalTrafo;
+  finalTrafo[3][2] -= mCamDistance;
```
dcwhite, line 112: Can this be unit tested separately? The camera rotation problems are probably coming from here.

src/Interface/Modules/Render/ES/SRCamera.h
```diff
@@ -80,7 +80,7 @@ namespace SCIRun{
       void rotate(glm::vec2);
 
       // P  = Projection matrix | IV = Inverse view matrix |  V  = View matrix
-      const glm::mat4& getWorldToView() const        {return mV;}
+      const glm::mat4 getWorldToView() const { return mArcLookAt->getWorldViewTransform(); }
```
dcwhite, line 83: This is another suspect change. Maybe mV was set to something else in the code?

dcwhite:
Options:
* Undo the non-required changes in the camera code and see if that fixes everything
* Restart the conversion process, first by changing those float_ts, then make a wrapper for the mat4 type to force the compiler to catch all the default-constructed cases before you switch GLM versions; weed those out and centralize the initialization, making sure rotation works on all platforms. Then do the GLM version switch again
* If neither of those reveals the error, leave SCIRun code the same, but experiment with older versions of GLM, perhaps it is a platform-specific problem on their side (entirely a possibility with the latest release of something)

RubioJr9:
> Options:
> 
>     * Undo the non-required changes in the camera code and see if that fixes everything
> 
>     * Restart the conversion process, first by changing those float_ts, then make a wrapper for the mat4 type to force the compiler to catch all the default-constructed cases before you switch GLM versions; weed those out and centralize the initialization, making sure rotation works on all platforms. Then do the GLM version switch again
> 
>     * If neither of those reveals the error, leave SCIRun code the same, but experiment with older versions of GLM, perhaps it is a platform-specific problem on their side (entirely a possibility with the latest release of something)

Great ideas. I'm going to go through the list in order and see if that fixes it. I tested again on linux as a sanity check and camera rotation works, including auto rotate. I'm building on my macbook right now so hopefully I can replicate it there.

RubioJr9:
@dcwhite This is ready to test again. Camera rotation works on mac os now.

dcwhite:
Cool!

src/Interface/Modules/Render/ES/RendererCollaborators.h
```diff
@@ -258,11 +258,11 @@ namespace SCIRun
       explicit ObjectScaleCalculator(const BasicRendererObjectProvider* s, const ScaleParameters& p);
       gen::Transform computeTransform(int x, int y) const override;
     private:
-      glm::vec3 originView_;
+      glm::vec3 originView_ {0.0};
       float projectedW_;
-      glm::vec3 flipAxisWorld_;
-      glm::vec3 originToSpos_;
-      glm::vec3 originWorld_;
```
dcwhite, line 267: With so many of these needing new zero-initialization, perhaps derive a type that default-constructs with zeros? `struct SCIVec3` or something

src/Interface/Modules/Render/ES/SRInterface.cc
```diff
@@ -828,7 +828,7 @@ uint32_t SRInterface::getSelectIDForName(const std::string& name)
         * (glm::length(originToCurrentSpos) / glm::length(originToSpos_));
 
       // Flip if negative to avoid inverted normals
-      glm::mat4 flip;
```
dcwhite, line 830: Same with this one, make it `SCIMat4`

dcwhite: It can be in a separate PR if you want

src/Interface/Modules/Render/ES/RendererCollaborators.h
RubioJr9: I agree, but that name sounds like the native Vector was duplicated. How about GLMVec3?

src/Interface/Modules/Render/ES/SRInterface.cc
RubioJr9: GLMMat4?

dcwhite: SCI or GLM are pretty redundant. How about highlighting the use case (https://en.wikipedia.org/wiki/Transformation_matrix), so TransformationMatrix or ProjectiveMatrix depending on the context?

RubioJr9:
I made wrappers for the glm vec, mat, and quat types. I added test cases for default constructors and a few functions I added to wrap around glm function calls, such as matrix transpose and vector normals. I left spire alone since it is an external.

RubioJr9:
I forgot the test folder and files. That should fix the headless builds

src/Graphics/Datatypes/GLMMat4.h
```diff
@@ -0,0 +1,46 @@
...
+#pragma once
+
+#include <Graphics/Datatypes/GLMMat.h>
+#include <glm/mat4x4.hpp>
+#include <Graphics/Datatypes/share.h>
+
+namespace SCIRun {
+namespace Graphics {
+namespace Datatypes {
+  class GLMMat4 : public GLMMat<glm::mat4>
```
dcwhite, line 37:
```suggestion
  class SCISHARE GLMMat4 : public GLMMat<glm::mat4>
```

src/Graphics/Datatypes/GLMQuat.h
```diff
@@ -0,0 +1,61 @@
...
+#pragma once
+
+#include <glm/glm.hpp>
+#include <glm/gtc/type_ptr.hpp>
+#include <Graphics/Datatypes/share.h>
+
+namespace SCIRun {
+namespace Graphics {
+namespace Datatypes {
+  class GLMQuat : public glm::quat
```
dcwhite, line 37:
```suggestion
  class SCISHARE GLMQuat : public glm::quat
```

etc for all new classes

src/Graphics/Datatypes/GLMVec.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include <glm/glm.hpp>
+#include <glm/gtc/type_ptr.hpp>
+#include <Graphics/Datatypes/share.h>
+
+namespace SCIRun {
+namespace Graphics {
+namespace Datatypes {
+  template <typename parent>
+  class GLMVec : public parent
```
dcwhite, line 38: No SCISHARE on the template ones! Just the concrete classes

dcwhite:
Testing:
- [ ] Rotation on Mac
- [ ] Rotation on Windows
- [ ] BBox widget on Mac
- [ ] BBox widget on Windows

dcwhite:
@jessdtate @RubioJr9 Widget rotation is working on Mac, however I noticed the rotation speed seems to be about half of what it used to be, with respect to trackpad motion (so widget angle / trackpad distance is lower). As a result one might need to do multiple drags to rotate an arrow all the way from one side to another side. I don't know if that's a reasonable use case, so let me know what you want to do (keep as is, increase rotation speed to old level, meet in middle), it's up to you.

RubioJr9:
The change in rotation speed was not intended. I'm looking for the cause of it right now.

src/Graphics/Datatypes/GLMVec.h
```diff
@@ -0,0 +1,63 @@
...
+      const parent& asParent() const { return static_cast<const parent&>(*this); }
+      const float* value_ptr() const { return glm::value_ptr(this->asParent()); }
+      float size() const { return this->asParent().length(); }
+      float length() const { return glm::length(this->asParent()); }
+      GLMVec<parent> normal() const { return glm::normalize(this->asParent()); }
+
+      GLMVec<parent> operator-(const GLMVec<parent>& other) { return glm::operator-(*this, other); }
+      GLMVec<parent> operator+(const GLMVec<parent>& other) { return glm::operator+(*this, other); }
+      GLMVec<parent> operator*(const GLMVec<parent>& other) { return glm::operator*(*this, other); }
+      GLMVec<parent> operator/(const GLMVec<parent>& other) { return glm::operator/(*this, other); }
```
RubioJr9, line 59: So the issue with rotation was happening because the math operators for this wrapper returned the parent class. Is this an appropriate way to overload those to return the wrapper instead?

dcwhite:
@RubioJr9 Now some merge issues as well

dcwhite:
I'll take a deeper look at the wrappers and the dependencies in general. We might try rolling back before the wrappers were added and testing again.

jessdtate:
I think there may be some build errors with the latest merge.

jessdtate:
here is a build error I get: 
/Users/jess/software/SCIRun/src/Interface/Modules/Render/ES/SRCamera.h:85:67: error: use of undeclared identifier 'mV'
      const glm::mat4 getWorldToProjection() const   {return mP * mV;}

dcwhite:
@RubioJr9 I'll correct the merge since I moved the code.

dcwhite:
This needs a commit I made to a different branch. I asked @RubioJr9 to just cherry-pick it over: https://github.com/SCIInstitute/SCIRun/commit/2f71b97378ac6a80074e8624920dccf023aa1aee

RubioJr9:
Done

dcwhite:
@allywarner @jessdtate This one is ready to test again. Please do an exhaustive ViewScene and widget test.

dcwhite (commented):
Just need to test Windows again

allywarner (approved):

dcwhite:
Soooo close. On Windows, on my new giant screen, the dipole test network output got cropped in half:
cropped

I'll try later today with my normal monitor.

dcwhite (changes requested):
See Windows view problem above.

dcwhite:
Found more problems:
- [ ] Windows switching leads to VS skewing the view. 
- [ ] Reloading VS can fix the cropping, but gets those skew factors every click away
- [ ] Mac if you click away, you can lose rotation ability

dcwhite:
My Windows problems might have gone away. I will continue testing with more machines and VS versions later this week.

dcwhite:
Surface Book Windows hiDPI build has no widget crop problems. Continuing to test other Windows machines a bit more.

dcwhite:
Officially cannot replicate this again. Let's chalk it up to driver issue + monitor size--when I switch to the new monitor again, I will give it another test. So let's merge it.

dcwhite (approved):

allywarner (dismissed):

jessdtate:
I get a weird camera glitch, probably related to numerical instability with small camera angles.  Here is a video.
[Screen Recording 2020-11-25 at 1.21.14 AM.mov.zip](https://github.com/SCIInstitute/SCIRun/files/5595295/Screen.Recording.2020-11-25.at.1.21.14.AM.mov.zip)

dcwhite:
That's yuck. Does it happen as often as the video show, like every 2-3 small drags? I will try to replicate

jessdtate:
Yeah, the video is pretty representative. It seems to happen with slower mouse movements or larger objects.  Also, my last test made it lock into a camera position.

src/Externals/spire/arc-look-at/ArcLookAt.cpp
RubioJr9: I added a test that tests these transformations: offset->rotate->offset. This does not seem to be where the problems are coming from.

RubioJr9:
I found the problem with rotations. Please test again.

RubioJr9:
Whoops, did not mean to close it

dcwhite:
@RubioJr9 I still get it sometimes. Let me look at the code a bit more, I will comment further in a bit.

dcwhite:
Couldn't get much from the code. The problem seems a bit less prevalent, like with widgets. But it still happens a lot.

RubioJr9:
Thanks for checking that. I'll try to replicate it again

src/Graphics/Datatypes/Tests/GLMTests.cc
```diff
@@ -0,0 +1,77 @@
...
+  glm::mat4 identity = glm::make_mat4(identityList);
+  ASSERT_EQ(t, identity);
+
+  glm::vec3 lookAt = glm::vec3(2, 4.5, 7);
+  t = glm::translate(t, -lookAt);
+  float translateList[16] = {1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, -2, -4.5, -7, 1};
+  glm::mat4 translatedMat = glm::make_mat4(translateList);
+  ASSERT_EQ(t, translatedMat);
+
+  const static float theta = static_cast<float>(M_PI) * 0.25f;
```
dcwhite, line 60:
```
D:\a\SCIRun\SCIRun\src\Graphics\Datatypes\Tests\GLMTests.cc(60,49): error C2065: 'M_PI': undeclared identifier [D:\a\SCIRun\SCIRun\bin\SCIRun\Graphics\Datatypes\Tests\Graphics_Datatypes_Tests.vcxproj] [D:\a\SCIRun\SCIRun\bin\SCIRun_external.vcxproj]

D:\a\SCIRun\SCIRun\src\Graphics\Datatypes\Tests\GLMTests.cc(60,28): error C2789: 'theta': an object of const-qualified type must be initialized [D:\a\SCIRun\SCIRun\bin\SCIRun\Graphics\Datatypes\Tests\Graphics_Datatypes_Tests.vcxproj] [D:\a\SCIRun\SCIRun\bin\SCIRun_external.vcxproj]
```

dcwhite: Probably needs `#include <cmath>` for Windows build to work.

dcwhite (changes requested):

RubioJr9:
I'm not seeing the jumping with either the camera rotation, widgets, and auto rotate now. The camera rotation jumps only happened on mac os so I tested on there too.

jessdtate:
I'm not seeing the jump in the camera anymore.  Widgets look fine too.   anyone else seeing it?

src/Externals/spire/arc-ball/ArcBall.cpp
```diff
@@ -101,7 +101,7 @@ glm::quat ArcBall::quatFromUnitSphere(const glm::vec3& from, const glm::vec3& to
   //TODO: check if cross is 0 before normalize. Crashes on Windows
   auto c = glm::cross(from, to);
   if (c == glm::vec3(0, 0, 0))
-    return {};
+    return glm::quat(1.0, 0.0, 0.0, 0.0);
```
dcwhite, line 104: Haha, I made that change--causing a fun bug for the more knowledgeable graphics developer to find...

dcwhite:
I'll test today

jessdtate (approved):

dcwhite (approved):

dcwhite:
I'll merge and make a follow-up issue for:
- [ ] Windows build fix (I need to look into it more)
- [ ] refactoring the ArcBall function that caused the bug

PR #2136, opened 2020-06-08
Qwt Update and fix for Qt 5.15

RubioJr9:
-Updated the Qwt version tag so it works with Qt 5.15
-Fixed a constructor that was deprecated in Qt 5.15


dcwhite:
This one is needed urgently since now the Github action build downloads Qt 5.15 for every build. @jessdtate I'll try to approve today

Issue #2083, opened 2020-03-09
Review module defaults


dcwhite:
Discussed today:
- [ ] Reader input file types -> matlab
- [ ] Writer output file types -> matlab
- [x] ClipFieldByFunction @wgood 
- [x] MFDFSTD @wgood 
- [x] RWC @wgood

wgood:
ClipFieldByMesh @wgood (The default for ClipFieldbyfunction should be elementcenter instead of onenode. )
 MFDFSTD @wgood (The default for  MFDFSTD should be Constant Mapping: Each Destination Gets Nearest Source Value instead of Linear)
 RWC @wgood (The Default should be Affine or Rigid instead of Morph)

dcwhite:
Thanks!

dcwhite:
@wgood After recent discussions with @jessdtate and @jab0707, I'm going to try making a new "SmartReader" module from scratch that will have user-defined defaults, or just default to Matlab for everything. I had trouble with modifying the ancient legacy readers, and it would be cool to just have one read module that can handle everything. It will subsume issues #1216, #289, #2110, and this one.

Issue #289, opened 2013-03-28
File reader system: use Chain of Responsibility pattern


dcwhite:
File formats. Expand

dcwhite:
Check out description of https://github.com/SCIInstitute/SCIRun/issues/1216 as well.

dcwhite:
@jab0707 here’s one of two issues for the “smart reader” idea

dcwhite:
Closed with https://github.com/SCIInstitute/SCIRun/pull/2141

PR #2141, opened 2020-06-16
AutoReadFile module

dcwhite:
New module with smarter, simpler file reading functionality. Can read most matrix and field input types.


dcwhite:
Will also close #289, for beaucoup XP

dcwhite:
@wgood This new module will read `.mat` files first as Matlab, then as SCIRun, with no need to specify the file type--just connect a field or matrix pipe and it will read it automatically. So it ought to close #2083 as well.

wgood:
Awesome, thanks so much @dcwhite!

dcwhite:
@jessdtate This is ready, with the older Qt used in the action build for you to install and try.

dcwhite:
@Sailanarmo Some template code for you to peruse

dcwhite:
@jessdtate Did the older Qt fix work?

dcwhite:
Thanks. The goal was "minimum user interaction" in the spirit of "advanced SCIRun module," but yeah, we'll see how users like it

Issue #2054, opened 2020-01-31
Improve camera state changes via python interface


dcwhite:
Be able to pass a Python list rather than a string.

dcwhite:
Closed with #2065

dcwhite:
Actually the interface still uses strings, just in a better format. Reopening

PR #2149, opened 2020-06-29
Python Lists

RubioJr9:
Issue #2054
Reverted conversion of types of CameraLookAt and CameraRotation from list to strings(issue #2065) because this is more powerful
(should we delete ViewSceneUtility now?)

-Made get/set python functions able to convert between std::vector and python's list, which also works for multi-dimensional lists
-Cleaned up some python conversion code
-Updated test network for python camera controls


src/Core/Python/PythonDatatypeConverter.cc
```diff
@@ -34,30 +34,55 @@
...
+  {
+    py::object operator()(const int& v) const { return py::object(boost::get<int>(v)); }
+    py::object operator()(const std::string& v) const { return py::object(boost::get<std::string>(v)); }
+    py::object operator()(const double& v) const { return py::object(boost::get<double>(v)); }
+    py::object operator()(const bool& v) const { return py::object(boost::get<bool>(v)); }
+    py::object operator()(const AlgoOption& v) const
+    {
+      THROW_INVALID_ARGUMENT("Conversion to AlgoOption is not implemented.");
+    }
+    py::object operator()(const VariableList& v) const
```
RubioJr9, line 69: What is AlgoOption? Is it convertible to a Python object?

src/Core/Python/PythonDatatypeConverter.cc
```diff
@@ -451,50 +476,58 @@ namespace
...
   }
+
   {
     SparseRowMatrixExtractor e(object);
-    if (e.check())
-    {
-      return makeDatatypeVariable(e);
-    }
+    if (e.check()) return makeDatatypeVariable(e);
   }
```
RubioJr9, line 531: I was able to check the type without attempting an extraction on every type by checking for the python object's class name. Do the types DenseMatrix, SparseRowMatrix, and Field have expected class names of python objects I could use to clean up the remainder of this function?

RubioJr9, line 531: I was able to check the type without attempting an extraction on every type by checking for the python object's class name. Do the types DenseMatrix, SparseRowMatrix, and Field have expected class names of python objects I could use to clean up the remainder of this function?

dcwhite: They're just particularly-laid-out lists and dicts, so checking the classname probably won't work, and might've broken matrices on line 515.

src/Core/Python/PythonDatatypeConverter.cc
dcwhite: It's just an enumerated string value

src/Core/Python/PythonDatatypeConverter.cc
RubioJr9: If their storage also uses lists, we won't be able to distinguish between a matrix and a regular list of floats. I think we should type check the inputs of the set state function. Currently, we can set any parameter to any datatype. If we verify that the parameter's datatype matches the one that is stored, we can stop user error through the interface and decide which extractors to try between fields, matrices, and lists.

There is a slight problem with type checking being based on current state types. When you save a network and change the data type, you will only be able to set an object of the type that was used when creating the network. For example, a network made before this PR will not be able to set to CameraRotation and CameraLookAt to anything besides strings. Instead, we could verify that the type of data in the file matches the type of data being set as the default state. Here, we could attempt a conversion or just set to the default value.

src/Core/Python/PythonDatatypeConverter.cc
RubioJr9: Is this like an c++ enum but with string values?

src/Core/Python/PythonDatatypeConverter.cc
dcwhite: I ran into that same issue recently, thinking there should be some type-checking for Python state-setters. It would be annoying code to write, since you'd need to handle all the different type coercions that happen within core python, like converting ints to doubles. The Python interface philosophy has always been essentially user-driven, but user-beware. 

With these data checks looking at matrices and fields, those are usually used for data coming into the InterfaceWithPython module, rather than set_state calls. So the converters could be separated and made context dependent--use your new converter for state calls, and only check for matrices etc when data is being moved around. The catch is I think IWP uses transient_state variables to move the big datatypes around, so there's more special-case code to write. 

All-in-all, this is a good idea, but a lot more test code needs to be written.

src/Core/Python/PythonDatatypeConverter.cc
dcwhite: It's just a string but there's a list of possible string values also attached to it; just look at the class

src/Core/Python/PythonDatatypeConverter.cc
RubioJr9: I'm not sure what you mean by making the converters context dependent. Also, what data is being moved around?

dcwhite:
Discussed at today's meeting. @jessdtate how much more API change could we integrate to make type-checking easier? We are already going to change the functions for bringing data back and forth, so we can probably do something with the formatting too. That would make this type of change much smoother internally. We can discuss with UQ and FwdInv toolkit users at next week's meetings.

jessdtate:
@dcwhite I think there is room for significant change.  We can discuss what you have in mind.

dcwhite:
I'll need to prototype something, but the main idea would be to add some Python class wrappers for each SCIRun datatype, so new code (with the new functions we've been discussing) would look like:
```
scirun_set_output_data(PortName, SCIRunMatrix([[1,0],[0,1]]))
```
rather than passing raw Python lists and dictionaries. I can probably make a bunch of helper functions to save typing.

RubioJr9:
Also, I think we should also break break up the get/set functionality from module look ups.

For example, the function scirun_get_module_state("ModuleName", "StateVarName") could be:
module = scirun_get_module("ModuleName")
scirunDataWrapper = module.get_state("StateVarName")

Then the function scirun_set_module_state("ModuleName", "StateVarName", someData) could be:
module.set_state("StateVarName", scirunDataWrapper)

dcwhite:
That would require the boost::python object wrappers to work, which they didn't long ago. I won't have time to test that for a while...I haven't even started on the new functions.

dcwhite:
@dcwhite will devote a whole day to get this ready to go, somehow, next week.

dcwhite:
Idea: try adding a new state setting function specifically for lists.

dcwhite:
@RubioJr9 @jessdtate As I predicted, this simple network is broken by this change (passing a dense matrix out of IWP). Please commit it with the regression networks and make sure it works with whatever change you come up with:
[simplePythonMatrix.srn5.zip](https://github.com/SCIInstitute/SCIRun/files/4956790/simplePythonMatrix.srn5.zip)

RubioJr9:
I added type inference, such as converting between number types, for state variables and converted transient state variables by using the boost extractors like the old code. In PythonDatatypeConverter, lists are converted to Variable::List instead of considering DenseMatrix. Instead of InterfaceWithPython getting the matrix handle, it now receives the Variable::List and converts to DenseMatrix there.

dcwhite:
Cool, it'll get tested today

jessdtate:
This seems to break IWP in setting the output. Is there a change in the API?  Here is the error I get. 

ERROR: Caught exception: N5boost16exception_detail10clone_implIN6SCIRun4Core24InvalidArgumentExceptionEEE Message: The rows of the input matrix must be of equal size. /Users/jess/software/SCIRun/src/Modules/Python/PythonObjectForwarder.h(155): Throw in function Datatypes::DenseMatrix SCIRun::Core::Algorithms::Python::PythonObjectForwarderImpl::convertToDenseMatrix(const Variable::List &) const [PythonModule = SCIRun::Modules::Python::InterfaceWithPython] Dynamic exception type: boost::exception_detail::clone_impl std::exception::what: The rows of the input matrix must be of equal size. [SCIRun::Core::tag_error_message*] = The rows of the input matrix must be of equal size.

RubioJr9:
I fixed the bug. I was mixing up row and column sizes for the error check. I changed the test network to be non-square because it's a more useful test case. IWP currently handles mismatching row sizes by reporting no data was sent after timing out. This is a more useful error now that it's accurate.

RubioJr9:
Still okay after merging master. This will be ready to test again if the Github builds are fine

src/Core/Algorithms/Base/AlgorithmBase.cc
```diff
@@ -203,6 +203,11 @@ Variable SCIRun::Core::Algorithms::makeVariable(const std::string& name, const V
   return Variable(Name(name), value);
 }
 
+Variable SCIRun::Core::Algorithms::makeVariable(const Name& name, const Variable::Value& value)
+{
+  return Variable(name, value);
+}
+
```
dcwhite, line 210: Not sure this is needed since it's just a plain constructor call

dcwhite (changes requested):
Small stuff

src/Core/Algorithms/Base/AlgorithmBase.cc
```diff
@@ -203,11 +203,6 @@ Variable SCIRun::Core::Algorithms::makeVariable(const std::string& name, const V
   return Variable(Name(name), value);
 }
 
-Variable SCIRun::Core::Algorithms::makeVariable(const Name& name, const Variable::Value& value)
```
dcwhite: Nice to delete this, but the references were not cleaned up--see broken build. Just replace with constructor call.

dcwhite (changes requested):

jessdtate:
I'm getting the boost build error with clang 12.0.0.  I wasn't seeing it with clang 11.

dcwhite:
I'm trying to upgrade my older Macbook's XCode. It's not cooperating just yet...

jessdtate:
I had to download and install the command line tools separately

jessdtate (approved):

dcwhite (approved):

PR #2152, opened 2020-07-09
Eigen Tensor Class

RubioJr9:
New tensor class using Eigen::TensorFixedSize

-TensorBase is an N-rank tensor inheriting from Eigen
-DyadicTensor inherits from TensorBase(dyadic as in rank 2)
-Dyadic3DTensor inherits from DyadicTensor with both dimensions as 3 and gives additional functionality
-Dyadic2DTensor and Dyadic4DTensor are typedefs of DyadicTensor
-Tests cases


dcwhite:
@RubioJr9 Looks like some Mac and Windows compiler errors w.r.t. operator namespaces most likely.

src/Core/Datatypes/DyadicTensor.h
```diff
@@ -0,0 +1,306 @@
...
+        normalizeEigenvectors();
+        haveEigens_ = true;
+        setTensorValues();
+        reorderTensorValues();
+      }
+
+      // This function is listed as something that will be added to Eigen::Tensor in the future.
+      // Usage of this function should be replaced with Eigen's asMatrix function when it is
+      // implemented.
+      Eigen::Matrix<Number, Dim0, Dim1> asMatrix()
```
dcwhite, line 129:
```suggestion
      Eigen::Matrix<Number, Dim0, Dim1> asMatrix() const
```

src/Core/Datatypes/DyadicTensor.h
```diff
@@ -0,0 +1,306 @@
...
+        return eigvecs_;
+      }
+
+      std::vector<Number> getEigenvalues()
+      {
+        if (!haveEigens_) buildEigens();
+        return eigvals_;
+      }
+
+      SCISHARE friend std::ostream& operator<<(
```
dcwhite, line 156: I'm thinking these need to be declared as friends outside of the class to make Windows/Mac happy. An inline friend declaration usually isn't enough.

src/Core/Datatypes/DyadicTensor.h
```diff
@@ -0,0 +1,306 @@
...
+        return sum;
+      }
+
+     protected:
+      const int RANK_ = 2;
+      std::vector<DenseColumnMatrixGeneric<Number>> eigvecs_;
+      std::vector<Number> eigvals_;
+      bool haveEigens_ = false;
+
+      void buildEigens()
```
dcwhite, line 239: Consider making `eigvals_` and `haveEigens_` mutable, so that almost every function here can be marked const, assuming they don't change the underlying matrix/tensor values.

src/Core/Datatypes/Dyadic3DTensor.h
```diff
@@ -0,0 +1,184 @@
...
+      return Dyadic3DTensor(array[0], array[1], array[2], array[4], array[5], array[8]);
+    }
+
+    template <typename Number>
+    Dyadic3DTensor symmetricTensorFromNineElementArray(const std::initializer_list<Number>& array)
+    {
+      std::vector<Number> vec = array;
+      if (vec.size() != 9) THROW_INVALID_ARGUMENT("This function requires 9 values.");
+      return Dyadic3DTensorGeneric<double>(vec[0], vec[1], vec[2], vec[4], vec[5], vec[8]);
+    }
```
dcwhite, line 171: These convenience functions can be made even more convenient with a variadic function, which would move the runtime size check to compile time. It would look something like:
```
template <typename ...Number>
Dyadic3DTensor buildSymmetricTensorFrom(Number ... n)
{
   return Dyadic3DTensor(n...);
}
```
There would need to be some more machinery to get all the cases working (like the 9->6 element case), but it would be much easier to use for algo writers.

src/Core/Datatypes/Dyadic3DTensor.h
```diff
@@ -0,0 +1,184 @@
...
+
+    template <typename Indexable>
+    Dyadic3DTensor symmetricTensorFromMandel(const Indexable& array)
+    {
+      const double sqrt2 = std::sqrt(2);
```
dcwhite, line 176:
```suggestion
      static const double sqrt2 = std::sqrt(2);
```

dcwhite:
I'll check the Windows build today

dcwhite:
I'm having similar difficulty with the windows build.

dcwhite:
If you want a workaround, for this set of code, you can only include them in the CMakeLists with Apple or Linux (IF NOT WIN32 or something similar). Then it can be merged and worked on separately. If a module uses it, then Windows build will need to work.

RubioJr9:
This will be used in a module in the next pr or two so I'll try to fix it now. It looks like I fixed the original windows error about std::max and std::numeric_limits. This new one is an issue with a couple Eigen functions

src/Core/Datatypes/DenseColumnMatrix.h
```diff
@@ -48,6 +48,12 @@ namespace Datatypes {
 
     explicit DenseColumnMatrixGeneric(size_t nrows = 0) : EigenBase(nrows) {}
 
+    DenseColumnMatrixGeneric(std::vector<T> vals) : EigenBase(vals.size())
```
dcwhite, line 51:
```suggestion
    DenseColumnMatrixGeneric(const std::vector<T>& vals) : EigenBase(vals.size())
```

src/Core/Datatypes/DenseColumnMatrix.h
```diff
@@ -48,6 +48,12 @@ namespace Datatypes {
...
+    {
+      for (long unsigned int i = 0; i < vals.size(); ++i)
+        (*this)[i] = vals[i];
```
dcwhite, line 54:
```suggestion
     std::copy(vals.begin(), vals.end(), data());
```

src/Core/Datatypes/DenseColumnMatrix.h
```diff
@@ -48,6 +48,12 @@ namespace Datatypes {
...
+    }
```
dcwhite, line 55: Consider a move constructor for this case too; it would need a unit test, but would be super efficient.

src/Core/GeometryPrimitives/BBoxBase.h
```diff
@@ -38,9 +38,9 @@ namespace Geometry {
 class SCISHARE BBoxBase
 {
 protected:
-  Point cmin_;
-  Point cmax_;
-  bool is_valid_;
+ bool is_valid_;
```
dcwhite, line 41: This is the wrong way to fix that warning--it might break IO. Change the constructor list to match the declared order.

src/Core/Datatypes/DenseColumnMatrix.h
RubioJr9: I was using this constructor to easily make matrices for testing. It was useful, but I'm no longer using this class in the tensor code. Do you want me to delete the changes to this file or add the suggestions?

dcwhite: Delete away!

src/Core/GeometryPrimitives/BBoxBase.h
RubioJr9: What's the difference? How can this break IO?

dcwhite: These types are read in sequential byte order. If the layout order changes, then the data will be read in the wrong order.

RubioJr9: That makes sense. I already changed this by the way

dcwhite: Cool thanks

src/Core/Datatypes/DyadicTensor.h
```diff
@@ -0,0 +1,322 @@
...
+    class DyadicTensorGeneric : public TensorBaseGeneric<Number, Eigen::Sizes<Dim, Dim>>
+    {
+     public:
+      using parent = TensorBaseGeneric<Number, Eigen::Sizes<Dim, Dim>>;
+      using VectorType = Eigen::Matrix<Number, Dim, 1>;
+      using MatrixType = Eigen::Matrix<Number, Dim, Dim>;
+
+      DyadicTensorGeneric() : parent() { parent::setZero(); }
+
+      DyadicTensorGeneric(Number val) : parent()
```
dcwhite, line 54:
```suggestion
      explicit DyadicTensorGeneric(Number val) : parent()
```

src/Core/Datatypes/DyadicTensor.h
```diff
@@ -0,0 +1,322 @@
...
+      {
+        for (size_t i = 0; i < Dim; ++i)
+          for (size_t j = 0; j < Dim; ++j)
+            (*this)(i, j) = (i == j) ? val : 0;
+      }
+
+      DyadicTensorGeneric(const std::vector<VectorType>& eigvecs) : parent()
```
dcwhite, line 61:
```suggestion
      explicit DyadicTensorGeneric(const std::vector<VectorType>& eigvecs) : parent()
```

src/Core/Datatypes/DyadicTensor.h
```diff
@@ -0,0 +1,322 @@
...
+        eigvecs_ = other.getEigenvectors();
+        eigvals_ = other.getEigenvalues();
+        haveEigens_ = true;
+      }
+
+      DyadicTensorGeneric(DyadicTensorGeneric<Number, Dim>&& other) : parent()
+      {
+        for (size_t i = 0; i < Dim; ++i)
+          for (size_t j = 0; j < Dim; ++j)
+            (*this)(index(i), index(j)) = std::move(other(index(i), index(j)));
```
dcwhite, line 89: This will just move individual doubles, or floats--not really worth it. Can you do a larger bulk move of the whole internals? Does the Eigen base class have a faster move constructor?

src/Core/Datatypes/DyadicTensor.h
```diff
@@ -0,0 +1,322 @@
...
+            (*this)(index(i), index(j)) = other(index(i), index(j));
+        eigvecs_ = other.getEigenvectors();
+        eigvals_ = other.getEigenvalues();
+        haveEigens_ = true;
+      }
+
+      DyadicTensorGeneric(DyadicTensorGeneric<Number, Dim>&& other) : parent()
+      {
+        auto otherData = other.data();
+        std::move(otherData, otherData + other.size(), this->data());
```
RubioJr9, line 88: Here's the new move statement. I confirmed this was being called from the new test I added.

dcwhite: That'll still just do a copy over the range for doubles since they're not movable one-by-one, but that's OK; if you ever make a DyadicTensorGeneric<std::thread>, for instance, this constructor will be most useful 🥇

RubioJr9: Oh so it just calls move over each item in the iteration, essentially copying the doubles?

dcwhite: Yeah it's just a range move, just like std::copy is a range copy--so it will call move on each element from begin to end. Good for large movable objects in a container, but still not moving a whole block at a time sort of thing (like std::vector's move ctor)

dcwhite: It's still an improvement, so let's just get it merged and optimize that later

RubioJr9: Sounds good. At the very least, I can add this range copy to the copy constructor.

Issue #2147, opened 2020-06-29
Red widget selection still not working


dcwhite:
Using network viewSceneLockWidgets, hold down shift, click multiple times on a bbox widget to make it red, then release shift--the widget stays red.

dcwhite:
Putting a VS in this state also makes other VSes not usable, until you hit shift again and then move the widget, releasing the "red lock." Please fix ASAP

dcwhite:
Happens on Mac and Windows

dcwhite:
Closed with #2166 (we think)

PR #2166, opened 2020-08-14
Red widget Selection fix

RubioJr9:
Issue: #2147

When a widget is deselected without movement, this restores the color without calling module feedback. When a widget is moved, the feedback is called without restoring widget color since the geometry will be regenerated.


dcwhite:
Mashing click with shift held down does work better. But I still got the redness to "stick" once or twice--I need to come up with some replication steps though. But I guess it's an improvement.

RubioJr9:
I'm not able to get it to stick to red on either my linux machine or my macbook. Would you be able to take a screen recording of it?

allywarner:
Tested on OSX 10.14.6 using ShowAndEditDipole. I couldn't get the red widget to stay! However, the flashing is back. It happens when I am clicking really quickly. If I slow down the clicks, it goes away.

RubioJr9:
So the latest commit trades screen flickering for a console error that says:
Unable to find uniform with name: uColor in shader with ID: 21
Unable to find uniform with name: hack in shader with ID: 21

This error doesn't seem to affect anything. It's an error coming from the shader used in widget selection in SRInterface. I used delays after certain events so that it happens only with large amounts of widgets(>250) when you make the widget selection function repeat very quickly.

I also cached the list of widgets so it only repopulates when there is new geometry. I also cached the previously selected widget so we do not have to search for the widget when the camera and mouse are not moved.

A technique that would fix the console error and increase performance would be to cache the selection buffer so it only has to rerender on execution or camera movement. I tried to reuse the selection buffer, but the entities are removed after selection. Leaving them in right now makes it so the selection objects appear in ViewScene.

src/Interface/Modules/Render/ViewScene.h
```diff
@@ -321,8 +323,18 @@ namespace SCIRun {
       bool                                  invertZoom_                   {};
       bool                                  shiftdown_                    {false};
       bool                                  mouseButtonPressed_           {false};
+      bool                                  frameIsFinished_              {false};
       Graphics::Datatypes::WidgetHandle     selectedWidget_;
+      Graphics::Datatypes::WidgetHandle     previousSelectedWidget_;
+      glm::mat4                             previousCameraTransform_      {0.0};
+      std::chrono::_V2::system_clock::time_point timeWidgetColorRestored_ {};
```
dcwhite, line 330: These variables definitely should not be in the main header, but in an impl class in the cc file only. They are probably what break the mac and windows builds. Please fix

RubioJr9: I made a new class in ViewScene for checking previous widget selection information.

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -149,8 +149,166 @@ namespace Gui {
...
+  if (previousSelectedWidget_)
+    return true;
+  else
+    return false;
+}
+
+//--------------------------------------------------------------------------------------------------
+bool PreviousWidgetSelectionInfo::deletePreviousWidget()
+{
+  previousSelectedWidget_.reset();
```
dcwhite, line 297: @RubioJr9 `error C4716: 'SCIRun::Gui::PreviousWidgetSelectionInfo::deletePreviousWidget': must return a value `

RubioJr9: Sorry, I'm not sure how that didn't trip up my compiler on mac os

src/Modules/Render/ViewScene.cc
```diff
@@ -273,6 +275,9 @@ void ViewScene::execute()
     }
   }
 #endif
+  state->setValue(HasNewGeometry, true);
+  state->setValue(TimeExecutionFinished, int(std::chrono::duration_cast<std::chrono::milliseconds>(
+      std::chrono::system_clock::now().time_since_epoch()).count()));
```
dcwhite, line 280:
```suggestion
  state->setValue(TimeExecutionFinished, make_a_function_for_this_fancy_time_call());
```

dcwhite:
Seems to be working pretty well now.

dcwhite:
@allywarner @jessdtate Please test the latest version again

jessdtate (approved):

PR #2172, opened 2020-09-11
Fix Html Tests

nids2001:
Fix html-proofer errors from documentation site

### DIFF
--- a/docs/scripts/cibuild.sh
+++ b/docs/scripts/cibuild.sh
@@ -2,4 +2,4 @@
 set -e # halt script on error
 
 bundle exec jekyll build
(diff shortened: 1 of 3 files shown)


docs/scripts/cibuild.sh
```diff
@@ -2,4 +2,4 @@
 set -e # halt script on error
```
dcwhite, line 2: Just delete this file. It's not used anymore; the old travis build used it on the old pages repo; now everything is just in the action yml file.

dcwhite (changes requested):

dcwhite (approved):

PR #2173, opened 2020-09-13
Fix ViewScene destructor crash after selecting a widget

dcwhite:
To test, load/create any network with a widget (e.g. src/ExampleNets/Renderer/embb2.srn5), move the widget around (shift-click-drag), then click New Network OR load a new network file OR close the whole program, and SCIRun should not crash (before, it would in any of those cases). Try all three ways to crash it.

ALSO: test with resizing the ViewScene window. Move widget around, then resize the ViewScene viewer window, then resize some more--widget should still be selectable and should not crash.


dcwhite:
@jessdtate @allywarner @RubioJr9 Urgent testing needed since it's @Haydelj's last day!!!

dcwhite:
Something in this change seems to break widget selection after resizing the VS window. Found by @allywarner. Switching to draft.

RubioJr9:
Sorry, I didn't see this post earlier. I'm building it now

dcwhite:
@allywarner ready to test again! Note updated testing instructions.

dcwhite:
I think Mac is working, but not Windows. Grr.

dcwhite:
Just kidding, tested on Windows and it seems good there too.

jessdtate (approved):

RubioJr9:
I tested on all 3 platforms. Both the destructor crash and resizing bug are gone.

dcwhite:
Thanks a lot! I'll merge this today

PR #2176, opened 2020-09-22
Fixed Widget Selection on Windows

RubioJr9:
I moved a value to transient state and am using unsigned variables for time values.


dcwhite:
Tested on Windows and Mac, and they both behave the same now (Windows being as correct as Mac now). So I think we can merge this in quickly in order to continue testing mine (#2173). Please test soon @jessdtate

src/Interface/Modules/Render/ViewScene.cc
```diff
@@ -186,19 +186,20 @@ namespace Gui {
...
-  return long(int(timeSinceEpoch(std::chrono::system_clock::now())) -time);
+  unsigned int now = timeSinceEpoch(std::chrono::system_clock::now());
+  return  now - time;
 }
 
 //--------------------------------------------------------------------------------------------------
-long PreviousWidgetSelectionInfo::timeSinceEpoch(const std::chrono::system_clock::time_point& time) const
+unsigned long PreviousWidgetSelectionInfo::timeSinceEpoch(const std::chrono::system_clock::time_point& time) const
 {
   return std::chrono::duration_cast<std::chrono::milliseconds>(time.time_since_epoch()).count();
```
dcwhite, line 204: chrono::milliseconds is "at least 45-bit integer" so you ought to change this to uint64_t rather than just unsigned long in order to avoid rollover bugs. Since it works on both of our machines for now, this can wait; maybe the windows problem was the use of the non-transient state.

dcwhite: @RubioJr9

dcwhite (approved):

jessdtate (approved):

PR #2183, opened 2020-10-16
Goodbye Cleaver 1

dcwhite:


jessdtate:
Hmm.  Github changed the layout and now I cannot find the builds that it generates.

jessdtate:
Maybe it's just not here:
`Warning: No files were found with the provided path: bin/SCIRun/SCIRun-5.0.beta.*-Darwin.pkg. No artifacts will be uploaded.`

dcwhite:
Looks like the master-deployment part of the script pushes an intermediate tag that messes up the package name, so it doesn't match and doesn't get copied. I'll need another PR to fix that.

dcwhite:
Or maybe I can just push it here...let me try

dcwhite:
@jessdtate Downloadable installer is working again

jessdtate (approved):

PR #2195, opened 2020-11-13
Update thread class

dcwhite:
Still needs some cleanup and one more thread converted. Stop button is a work in progress. Please test on Linux @RubioJr9 @nids2001 and newer Mac @jessdtate 

Also to test: 
- [ ] ShowString with ViewScene resizing. I deleted code that re-executed to keep the string object in the right place, since it seemed like it still worked ok without it, and didn't hang the test suite.


dcwhite:
Definitely something up with VS's resizeTimer connecting to ShowString, and executing from the command line--the reexecute hangs the process.

RubioJr9:
Nothing seems wrong with my linux build. Is there any specific networks you would like tested?

dcwhite:
I think the ShowString problem shows up in the regression test run but might be OK with normal usage. I also tested looping in GetMatrixSlice and that seemed OK too, but the example was trivial. What I did notice in networks with ShowString (like cleaver2Mickey.srn5), was the network execution bar would reset to 0% before all the modules were done, and ShowString was the culprit. I haven't debugged this yet. What was funny was the network executed fine, just the bar got confused, which could be why it failed in the regression setting--it probably quit too early.

dcwhite:
So try to make a complicated network with those two modules/features--loop over a matrix and pass each element to ShowString, for example, and see if you can get any weird behavior.

jessdtate (approved):

PR #2215, opened 2021-01-18
Change CPU scaling governor to performance during build

mckees:
This change alters the CPU scaling governor from powersave to performance during a build. Essentially, this lets all CPUs run at full capacity. Similar changes are implemented in the build for ChromeOS, though they require a dependency that you may or may not be interested in adding to your toolchain:
https://chromium-review.googlesource.com/c/chromiumos/platform/crosutils/+/1841944/2/build_packages

Potential risks of this change:
1. If a user has changed their settings to "performance" (unlikely since it is not default behavior), this will change their CPU governor back to powersave. 
2.  I am not super familiar with your build. If I have missed a build failure path, someone's machine could end up stuck in performance mode. I suspect that debugging sudden battery life decrease could be difficult. So your build expert(s) should take a look and make sure the that `set_powersave_mode` gets called in any build failure paths.

I did some performance timing on my i7-7820X using the `time` command to measure real time:
- 16 thread build: 7.4% improvement
- 4 thread build: 8.2% improvement
- 1 thread build: 8.3% improvement

EDIT: Tested on Ubuntu

@dcwhite 
@allywarner


dcwhite:
Cool! Can it be enabled with a flag to `build.sh`, such as `--speed-up-build` to enable easier testing on the build farm?

dcwhite (changes requested):
See above comment

PR #2222, opened 2021-01-24
Bounding box widget transformations

RubioJr9:
-Added rotation and scale to bounding box
-Fixed none type widget movement
-Refactored the transformation calculators
-Added calculators for axis translate and scale


dcwhite:
@RubioJr9 Looks like the test code doesn't compile (on Linux even), please take a look.

RubioJr9:
> @RubioJr9 Looks like the test code doesn't compile (on Linux even), please take a look.

I'm confused as to what the issue is. I think this might be it, but I haven't touched that code: "error: invalid new-expression of abstract class type ‘SCIRun::Graphics::Datatypes::StubGlyphFactory’"

dcwhite:
Need to implement base class function added to interface here:
https://github.com/SCIInstitute/SCIRun/blob/master/src/Graphics/Widgets/Tests/WidgetTestingUtility.h#L77

there are also errors in the BoundingBoxTest, I think an include or rename is needed.

dcwhite:
The tests compile now but they don't pass anymore. We can ignore that for now; I'll fix them in a follow up issue

RubioJr9:
> The tests compile now but they don't pass anymore. We can ignore that for now; I'll fix them in a follow up issue

It's because I changed BoxPosition to use eigenvectors instead. I tried compiling with Debug to test it, but the test executables didn't appear.

dcwhite:
Feedback from meeting:

1. all other sphere widgets are broken now
1. first transform seems to cancel itself
1. scale clipping to screen too close?
1. global vs other types of scaling--use cubes? diagonal cylinder?
1. move rotate to corner widget
1. add user custom widgets --LATER FEATURE
1. change mouse pointer while transforming
1. toggle transparency when other fields block a widget control
1. relative scaling of controls

dcwhite:
@RubioJr9 to fix from above:
- [ ] 1
- [ ] 2
- [ ] 4
- [ ] 5
- [ ] 7
- [ ] 9

RubioJr9:
> @RubioJr9 to fix from above:
> 
>     * [ ]  1
> 
>     * [ ]  2
> 
>     * [ ]  4
> 
>     * [ ]  7
> 
>     * [ ]  9

Progress:
* [X]  1
* [X]  2
* [X]  4
* [ ]  7
* [X]  9

dcwhite:
Same build error in test code as earlier, due to not adding factory function:
```
In file included from /home/runner/work/SCIRun/SCIRun/src/Graphics/Widgets/Tests/SphereWidgetTest.cc:31:0:
/home/runner/work/SCIRun/SCIRun/src/Graphics/Widgets/Tests/WidgetTestingUtility.h:70:22: note:   because the following virtual functions are pure within ‘SCIRun::Graphics::Datatypes::StubGlyphFactory’:
       class SCISHARE StubGlyphFactory : public AbstractGlyphFactory
                      ^~~~~~~~~~~~~~~~
In file included from /home/runner/work/SCIRun/SCIRun/src/Graphics/Widgets/Tests/WidgetTestingUtility.h:33:0,
                 from /home/runner/work/SCIRun/SCIRun/src/Graphics/Widgets/Tests/SphereWidgetTest.cc:31:
/home/runner/work/SCIRun/SCIRun/src/Graphics/Widgets/GlyphFactory.h:53:29: note: 	virtual std::__cxx11::string SCIRun::Graphics::Datatypes::AbstractGlyphFactory::superquadric(SCIRun::Graphics::Datatypes::SuperquadricParameters, SCIRun::Graphics::Datatypes::WidgetBase&) const
         virtual std::string superquadric(SuperquadricParameters params, WidgetBase& widget) const = 0;
                             ^~~~~~~~~~~~
```

dcwhite:
@RubioJr9 Please fix above

src/Interface/Modules/Render/ES/ObjectTransformCalculators.cc
```diff
@@ -29,6 +29,8 @@
 #include <Interface/Modules/Render/ES/ObjectTransformCalculators.h>
 #include <Interface/Modules/Render/ES/SRCamera.h>
 #include <glm/gtx/transform.hpp>
+#include <glm/gtx/string_cast.hpp>
+#include <cmath>
```
dcwhite, line 33:
```suggestion
#include <Core/Math/MiscMath.h>
```

dcwhite (changes requested):

jessdtate:
I don't see the clipping in my testing.  Lets get the initialization fixed asap and merge it.

dcwhite (approved):

jessdtate (approved):
I see some strange behavior with multiple viewscenes, but nothing too serious. it doesn't always sync on the first move (after is fine) and the mouse icons sometimes stays as pointer (instead of a hand) in one of the viewscenes.    I think we should merge it and fix it after the release.

PR #2237, opened 2021-03-05
ViewScene Quick Screenshot

nids2001:
Add a new quick screenshot button to ViewScene GUI. Path for Screenshot location set in global preferences.


src/Interface/Modules/Render/Screenshot.cc
```diff
@@ -88,11 +97,25 @@ void Screenshot::saveScreenshot(const QString& filename)
...
+  if (!fileName.isEmpty())
+  {
+    QMessageBox::information(nullptr, "ViewScene Screenshot", "Saving ViewScene screenshot to: " + fileName);
+    screenshot_.save(fileName);
+  }
+}
+
+QString Screenshot::screenshotFileFromPreferences() const
+{
+  return screenshotDirectory() + QString("/viewScene_%1_%2.png").arg(QDateTime::currentDateTime().toString("yyyy.MM.dd.HHmmss.zzz")).arg(index_);
```
jessdtate, line 113: Is there an easy way to add the viewscene number? it looks like it tracks the index for each viewscene window.   How about the network name?  These could be pushed until after the release.

jessdtate (approved):

dcwhite (approved):

PR #2244, opened 2021-03-18
Fix performance regression from overuse of interruption checks

dcwhite:
This didn't show up on Mac, so I missed it initially. It caused a crazy 300-1000X slow down in ShowField and any algorithm that (over)used the interruption feature using the new future/promise mechanism implemented in #2195, on Windows and Linux. That feature wasn't enabled in the GUI anyway, so I just took out all the checkForInterruption calls until I can figure out a less intrusive way to implement this. Please test and merge ASAP.

For example: ShowField on a 100^3 latvol went from 308 seconds to 1.42 seconds on my i7-10710U.

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -17,7 +17,7 @@ jobs:
         fetch-depth: 0
 
     - name: Install Qt
-      uses: jurplel/install-qt-action@v2.9.0
+      uses: jurplel/install-qt-action@v2.13.0
       env:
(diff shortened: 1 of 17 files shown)


dcwhite:
@RubioJr9 @nids2001 Linux users please test this!

RubioJr9:
I did the latvol test with 100^3 too. It took about 5 minutes on master and 1 second on this branch.

jessdtate (approved):

PR #2266, opened 2021-04-22
Turn on some warnings as errors on Mac, and fix Ospray build

### LINKED ISSUE #2111: Investigate some warnings as errors in the build
A couple ones came up recently--escape characters in a regex, and more dangerously, an infinite-recursive function call that broke dipoles. These can be marked as errors, which will halt the CI build. Just need to figure out the cmake syntax.

dcwhite:
Closes #2111


allywarner (approved):

src/Core/Algorithms/Legacy/Forward/BuildBEMatrixAlgo.cc
```diff
@@ -523,15 +519,15 @@ void BuildBEMatrixBase::make_auto_G_allocate(VMesh* hsurf, DenseMatrixHandle &h_
...
+  BuildBEMatrixBaseCompute::make_auto_G_compute(hsurf, *h_GG_, in_cond, out_cond, avInn);
 }
 
 template <class MatrixType>
 void BuildBEMatrixBaseCompute::make_auto_G_compute(VMesh* hsurf, MatrixType& auto_G,
-  double in_cond, double out_cond, double op_cond, const std::vector<double>& avInn)
+  double in_cond, double out_cond, const std::vector<double>& avInn)
 {
   //const double mult = 1/(2*M_PI)*((out_cond - in_cond)/op_cond);  // op_cond=out_cond for all the surfaces but the outermost surface which in op_cond=in_cond
   const double mult = 1/(4*M_PI)*(out_cond - in_cond);  // op_cond=out_cond for all the surfaces but the outermost surface which in op_cond=in_cond
```
jessdtate, line 533: This is the form that is is the matlab code in the FwdInvToolkit, which as far as I know is the basis of this module

jessdtate: I'm not sure where the op_cond comes from.  seems like a good idea to remove it

jessdtate (approved):

Issue #1524, opened 2016-07-13
Sphere option: make scalable as percentage of bounding box


dcwhite:

jessdtate:
I agree

dcwhite:
Will discuss at next week's meeting

jessdtate:
For modules:  GenerateSinglePointProbe, GenerateSeedsFromField, GenerateElectrode.

jessdtate:
Standardize some of the features: color selection,  projection to input field.

RubioJr9:
PR #2267 implements this for GenerateSinglePointProbeFromField and GeneratePointSamplesFromField, but converting GenerateElectrode(#1289) precedes the percentage scaling for that module.

PR #2267, opened 2021-04-24
BBox Percentage Scaling for Point Probes

RubioJr9:
Issue: #1524
Reviewers: @jessdtate @dcwhite

- Added setting to scale sphere widgets as percentage of bounding box for:
    - GenerateSinglePointProbeFromField
    - GeneratePointSamplesFromField

- Fixed the tab ordering of UI elements in GenerateSinglePointProbeFromField. Tab ordering is the order of UI elements when you hit the tab button.


jessdtate (approved):

dcwhite (changes requested):
Problems:
- [ ] If no input field, the point probe disappears (probably because bbox is invalid or zero size)--should be an error, or default back to absolute size
- [ ] The two size widgets (percentage and absolute) should enable/disable with the checkbox toggle
- [ ] The percentage double spinbox ought to be 0-100 based (still floating point though, if someone wants 40.5%

RubioJr9:
> Problems:
> 
>     * [ ]  If no input field, the point probe disappears (probably because bbox is invalid or zero size)--should be an error, or default back to absolute size
> 
>     * [ ]  The two size widgets (percentage and absolute) should enable/disable with the checkbox toggle
> 
>     * [ ]  The percentage double spinbox ought to be 0-100 based (still floating point though, if someone wants 40.5%)
> 
>     * [ ]  On mac, the new widgets are a bit smushed and need more space

I addressed all of these in the new commit

dcwhite (approved):

allywarner (approved):

jessdtate (approved):

Issue #2257, opened 2021-04-04
View scene not saving state when restarting


jab0707:
The view scene window size, camera angle, distance, rotation etc. used to save in between reopening SCIRun. This was very important in order to make sure you could take screenshots of visualization at the same size/angle in case SCIRun crashed or you wanted to change the data. Now on each restart, the view scene seems to reset to a small window and some default rotation/camera angle etc. If this is going to be the case then the ability to set the camera state variables needs to be fixed so that the original view can be restored. Otherwise, SCIRun becomes 100% harder to use for visualizations in publications as there is no consistency.

Version: v5.0-beta.Z
OS: 
![image](https://user-images.githubusercontent.com/38257234/113514763-c3719580-952d-11eb-9afe-d560ada970d6.png)

dcwhite:
Working on this together with #1939. These got broken at some point (maybe with newer Qt version) and aren't part of any test script.

PR #2272, opened 2021-04-30
Fix Camera Rotation State Variable

### LINKED ISSUE #2254: Set State for Viewscene Camera Behaving Weirdly
When using python commands to set state for the view scene camera I have encountered an issue:

Starting with the following view scene window:
image

I then open the python consol and run the following commands:

```
look = scirun_get_module_state('ViewScene:1','CameraLookAt')    
rot = scirun_get_module_state('ViewScene:1','CameraRotation')
dist = scirun_get_module_state('ViewScene:1','CameraDistance')
```

I then move the viewscene to look like this:
image

I then want to recover the original view in the viewscene from the first image. So I run the following code:
```
scirun_set_module_state('ViewScene:1','CameraRotation',rot)
scirun_set_module_state('ViewScene:1','CameraLookAt',look)
scirun_set_module_state('ViewScene:1','CameraDistance',dist)
```

Instead of getting the original view I get: 
image

It seems like the rotations is being applied to the image rather than being set. So instead of getting the intended rotation I get some compound of the current rotation and what I pass into the set field. I can check this by checking the value of CameraRotation from the beginning and the end:
```
rot
output: [0.7355649471282959, 0.3985646367073059, -0.16810239851474762, -0.5213748216629028]
```

but when I set that as the rotation I get the following from get module state:

```
scirun_get_module_state('ViewScene:1','CameraRotation')
output: [0.39856472611427307, -0.1681024432182312, -0.5213749408721924, 0.7355651259422302]
```

RubioJr9:
Fixes #2254

I changed the array accesses of glm::quat to the wxyz variables for camera state. This is needed because the data ordering is no longer consistent with the constructor ordering with the recent glm upgrade.

Here are python commands for getting and setting the camera state for testing:
rot = scirun_get_module_state('ViewScene:0','CameraRotation')
scirun_set_module_state('ViewScene:0','CameraRotation', rot)

Reviewers: @dcwhite @jessdtate 
@jab0707 you're welcome to test if you'd like

### DIFF
--- a/src/Interface/Modules/Render/ViewScene.cc
+++ b/src/Interface/Modules/Render/ViewScene.cc
@@ -881,7 +881,7 @@ void ViewSceneDialog::pushCameraRotation()
   if(!spire) return;
 
   auto q = spire->getCameraRotation();
-  state_->setValue(Modules::Render::ViewScene::CameraRotation, makeAnonymousVariableList(q[0], q[1], q[2], q[3]));
+  state_->setValue(Modules::Render::ViewScene::CameraRotation, makeAnonymousVariableList(q.w, q.x, q.y, q.z));
   pushingCameraState_ = false;
 }
 


dcwhite:
We can probably get this merged today since it's a one-liner, but I found another problem with camera state pulling: it's called in the wrong place, it's in the VS constructor and it should be in pullSpecial which is called externally. So the wrong state get pulled on start-up. I'll fix that on my branch (which ought to fix #2257).

jessdtate (approved):

dcwhite (approved):

PR #2276, opened 2021-05-07
Clean up ViewScene state management 

dcwhite:
Test ViewScene's new/fixed saved state:
- [ ] window size
- [ ] window position
- [ ] docked/undocked
- [ ] camera settings


jessdtate (approved):

jessdtate:
I think we should merge it.  This is not a big deal, docked windows default to the default size and position, instead of stacking or keeping the height.  The width seems to save correctly, which is the most import parameter with the docked windows.

PR #2283, opened 2021-05-25
Sizing Fix for EditMeshBoundingBox

RubioJr9:
The sizing in the menu for EditMeshBoundingBox is half of what it actually should be. For example, if a latvol goes from (-1,-1,-1) to (1,1,1), the current size is listed as 1 for each axis, when it actually should be 2. This was originally set to half because that is the distance from the center of the volume and that is what is used for constructing the transform matrix.


dcwhite (changes requested):
All magic numbers. 

Define one constant that explains the issue, then either multiply or divide by it at the places you changed. So either `const double WIDGET_SCALE_FACTOR = 0.5` or `const double NORM_SCALE_FACTOR = 2.0`, your pick.

RubioJr9:
Fair point. I'll update it today.

jessdtate (approved):

dcwhite (approved):

Issue #1381, opened 2016-03-28
Convert enums to scoped enums


dcwhite:
Effective Modern C++ Item 10

dcwhite:
This should be beta.

dcwhite:
**Triage 2026-06:** Convert enums to scoped enums (Effective Modern C++ Item 10) — unscoped enums pollute enclosing namespace. clang-tidy's `modernize-use-scoped-enums` would identify candidates. The most impactful ones are enums used as state keys or option selectors. Low effort per file; medium for the whole codebase. Good first contribution.

PR #2285, opened 2021-06-01
Complete viewscene state

### LINKED ISSUE #1939: ViewScene saved parameters are not applied when network is loaded
ViewScene does not use the standard widget managers, so the saved state is inconsistently applied when loading saved networks. For instance, background scene color is loaded, but not fog color. We should move as much code to the standard pulling mechanism as possible.

dcwhite:
Closes #1939 

Also made the change to scoped enums. #1381
 
And M1 mac build works, at least headless.


Sumientra:
Just tested on macOS 11.4. 

Settings that did get saved:
- background color
- zoom
- scale bar

Settings that did not get saved:
- window size
- object selection settings
- clipping plane settings
- lights settings
- orientation axes

dcwhite:
Please provide some saved network files that demonstrate those settings not loading correctly.

Sumientra:
Network: download [here](https://drive.google.com/file/d/15Qmo4bv9gfzuAogghFkOQNBtzGCW9ts6/view?usp=sharing)

Settings before saving:
Screen Shot 2021-06-04 at 11 36 01

Settings after reloading:
Screen Shot 2021-06-04 at 11 36 20

Sumientra:
Now tested same network on macOS 10.15.7. 

Settings that did get saved:
- background color
- zoom
- scale bar
- window size
- object selection settings
- clipping plane settings (but the settings do not get applied until you uncheck and check the box)
- orientation axes

Settings that did not get saved:
- lights settings

dcwhite:
Hmm, not sure why MacOS version is a factor here. Anyway, I can test on 11.3 now.

I don't think I hooked up the light settings yet, I'll check. And clipping state is a little flaky, as you found.

dcwhite:
Oh, one problem with the screenshots--you're not using the updated version, that's the milestone build.

jessdtate:
I was finally able to test this.  I'm getting a build Error:
```
/Users/jess/software/SCIRun/src/Modules/Legacy/Fields/InterfaceWithTetGenImpl.cc:341:25: error: unknown type
      name 'TETVOLMESH_E'
    FieldInformation fi(TETVOLMESH_E,CONSTANTDATA_E,DOUBLE_E);
                        ^
/Users/jess/software/SCIRun/src/Modules/Legacy/Fields/InterfaceWithTetGenImpl.cc:341:38: error: unknown type
      name 'CONSTANTDATA_E'
    FieldInformation fi(TETVOLMESH_E,CONSTANTDATA_E,DOUBLE_E);
                                     ^
/Users/jess/software/SCIRun/src/Modules/Legacy/Fields/InterfaceWithTetGenImpl.cc:341:53: error: unknown type
      name 'DOUBLE_E'
    FieldInformation fi(TETVOLMESH_E,CONSTANTDATA_E,DOUBLE_E);
                                                    ^
/Users/jess/software/SCIRun/src/Modules/Legacy/Fields/InterfaceWithTetGenImpl.cc:341:24: error: parentheses
      were disambiguated as a function declaration [-Werror,-Wvexing-parse]
    FieldInformation fi(TETVOLMESH_E,CONSTANTDATA_E,DOUBLE_E);
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jess/software/SCIRun/src/Modules/Legacy/Fields/InterfaceWithTetGenImpl.cc:341:25: note: add a pair of
      parentheses to declare a variable
    FieldInformation fi(TETVOLMESH_E,CONSTANTDATA_E,DOUBLE_E);
```

jessdtate:
Also the lighting is different in the artifact that was generated.  I'll investigate it some more tomorrow

dcwhite:
Oops never built tetgen. I'll fix that today.

jessdtate:
looks like the light color doesn't initialize properly in a previously saved network.  Once I click on the change light color button it fixes the lighting.  without doing it, it only shows the ambient component

dcwhite:
OK I moved some lighting init code around so maybe it's working differently. But the code I changed should not have been correct before. Was it working completely on master?

jessdtate:
It works fine on master, but the color was not save in the state.  It works fine if it's a new network, of if the network has been saved with the newest version.  It's only a problem with a network saved in an old version (master) and loaded in this (branch), so it's a pretty narrow case, albeit one that most people will see.

jessdtate:
This fixes the save from a new model, but it still sets to 0,0,0 with an old network.  One thing I noticed is that an old module sets the colors as an int 0-255, but the new one is a float 0-1.  I'm not sure I'd want to sacrifice accuracy, but this may be a problem for users

jessdtate:
maybe we can add a note somewhere obvious? like the red viewscene thing?

dcwhite:
Can you attach a network with int colors saved? All of mine seem to be float.

dcwhite:
I believe that was a change made a long time ago, like years.

jessdtate:
[test_light_color_save.srn5.zip](https://github.com/SCIInstitute/SCIRun/files/6690938/test_light_color_save.srn5.zip)

This network was made on the branch, so the ints probably don't matter. It appears the color selector is returning ints, but the module handles it ok.  

One thing I noticed is that I have to toggle the meta layer to get it to update the state info

jessdtate:
with old modules loaded in this branch, it is still defaulting to 0,0,0.  Why would that be?

dcwhite:
Right the metadata layer doesn't update until execute. That should be an easy improvement but do make an issue for it.

dcwhite:
> with old modules loaded in this branch, it is still defaulting to 0,0,0. Why would that be?

Can you provide an old network that does this? Or just use any old one with headlights turned on...hard to search for that in the xml files...

jessdtate:
[test_light_color_save_II.srn5.zip](https://github.com/SCIInstitute/SCIRun/files/6718591/test_light_color_save_II.srn5.zip)

Here is a test case for the default settings that probably should be reverted to white

jessdtate:
[test_light_color_save_III.srn5.zip](https://github.com/SCIInstitute/SCIRun/files/6718627/test_light_color_save_III.srn5.zip)

multiviewscene.  One with non default that should be left alone

jessdtate (approved):

Issue #1949, opened 2019-06-06
Refactor new Ospray volume/colormap logic


dcwhite:
These need some cleanup. Follows #1911

dcwhite:
**Analysis (2026-06-13)**

Refactoring/cleanup issue following #1911 (OSPRay volume + colormap implementation). The volume rendering and colormap transfer function logic was added incrementally and needs consolidation.

Same blocker as all OSPRay issues: the integration is disabled in CI and the OSPRay 2.x upgrade decision is pending. Cleanup should happen alongside or after that upgrade, not before — otherwise the refactored code may need immediate rework for the new API.

PR #2287, opened 2021-06-04
Ospray: Latvol Fix and Camera Optimization

RubioJr9:
This fits with issue #1949, but probably shouldn't close it.

The latvol construction for ospray had 2 problems that I fixed:
* The y position of the origin was not set
* The dimensions needed to be reduced by 1 to calculate spacing

The reason for the spacing change is that there is 1 less segment of space than there are boundaries. For example, a latvol with dimension 5 along an axis would have nodes at [0, 1, 2, 3, 4], but the ranges of spacing would be [0-1, 1-2, 2-3, 3-4], which has 4 segments of space.

The camera data was being recomputed before accruing every frame, so I saved camera object for reuse.


dcwhite (dismissed):

dcwhite:
```
QObject::connect: No such slot SCIRun::Gui::OsprayViewerDialog::setWidth(igeomDataTransientnt)
QObject::connect:  (sender name:   'viewerWidthSpinBox_')
QObject::connect:  (receiver name: 'OsprayViewer')
```

dcwhite:
Also, I get a crash when OsprayViewer is deleted. Please check that all cleanup is done in the destructor.

dcwhite:
OK the destructor crash can be fixed with a change in ModuleWidget.cc, around line 363:

`  isViewScene_(name == "ViewScene") //TODO`

remove the check for `name == "OsprayViewer"`. There's some extra code that gets confused now related to window position saving. Please add this to this PR @RubioJr9

jessdtate (approved):

dcwhite (approved):

PR #2290, opened 2021-06-09
Ospray: Directional Light Fix

RubioJr9:
Issue #2140
This fixes rendering meshes

Issue #2140, opened 2020-06-11
Ospray 2.0 continued


dcwhite:
Follows #2064

RubioJr9:
This is a list of things to be done or work to be considered. We can always move some of these to an Ospray 3.0 issue.

Features lacking from ViewScene:
- [ ] Glyphs(scalar, vector, tensor)
- [ ] Axis orientation glyph
- [ ] Camera Controls
- [ ] Light Controls
- [ ] Autoview(#2292)
- [ ] Camera light(spotlight from camera to look at)
- [ ] Ambient lighting(#2293)
- [ ] Widgets

Volume Rendering:
- [ ] Support for non-homogenous volumes(ie. scalar field with values that are of different values)
- [ ] Fix colormap
- [ ] Fix latvol sizing and positioning(#2287)

dcwhite:
**Analysis (2026-06-13)**

OSPRay 2.0 continued. Checklist has 5/9 items resolved (Materials, Autoview, Ambient lighting, latvol sizing/positioning, camera light refs). Remaining: Glyphs, Axis orientation glyph, Camera Controls, Light Controls, Widgets, colormap fix. All blocked on OSPRay upgrade being stable and CI-enabled.

PR #2292, opened 2021-06-22
Ospray AutoView

RubioJr9:
Issue #2140
Waiting on PRs:
- [ ] #2290 
- [ ] #2287

### DIFF
--- a/src/Interface/Modules/Render/Ospray/OSPRayCamera.cpp
+++ b/src/Interface/Modules/Render/Ospray/OSPRayCamera.cpp
@@ -109,7 +109,7 @@ void OSPRayCamera::mouseWheel(int delta)
 float OSPRayCamera::toRadians(float v)
 {
   const static float HALF_TURN_DEGREES = 180;
-  const static float TO_RADIAN_OPERATOR = glm::pi<float>() * HALF_TURN_DEGREES;
+  const static float TO_RADIAN_OPERATOR = glm::pi<float>() / HALF_TURN_DEGREES;
   return TO_RADIAN_OPERATOR * v;
 }
 
--- a/src/Interface/Modules/Render/Ospray/OSPRayCamera.cpp
+++ b/src/Interface/Modules/Render/Ospray/OSPRayCamera.cpp
@@ -31,6 +31,7 @@
 #include <glm/gtc/type_ptr.hpp>
 
 using namespace SCIRun::Render;
+using namespace SCIRun::Core::Geometry;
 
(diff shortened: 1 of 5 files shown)


src/Interface/Modules/Render/Ospray/OSPRayCamera.cpp
```diff
@@ -104,3 +105,32 @@ void OSPRayCamera::mouseWheel(int delta)
...
+  // Convert core geom bbox to AABB.
+  Core::Geometry::Point bboxMin = sceneBBox_.get_min();
+  Core::Geometry::Point bboxMax = sceneBBox_.get_max();
+  glm::vec3 min(bboxMin.x(), bboxMin.y(), bboxMin.z());
+  glm::vec3 max(bboxMax.x(), bboxMax.y(), bboxMax.z());
+
+  spire::AABB aabb(min, max);
+
+  // The arcball class expects fov in radians
+  lookat_.autoview(aabb, toRadians(fovy_));
```
RubioJr9, line 129: Another usage of a spire data class(AABB) because we use spire's arcball class.

src/Interface/Modules/Render/Ospray/OSPRayCamera.cpp
```diff
@@ -104,3 +105,32 @@ void OSPRayCamera::mouseWheel(int delta)
...
+  pos_ = lookat_.getPos();
```
jessdtate, line 130: Autoview should change the lookat to the center of the BBox

jessdtate (changes requested):

src/Interface/Modules/Render/Ospray/OSPRayCamera.cpp
RubioJr9: I made this change, but I think you meant the target since lookat is the arcball object. I'll change that naming.

RubioJr9:
Found a camera bug where an object that's offset from the origin is moved to the wrong direction after changing the camera after initialization. I'll fix this on this branch since it's causing issues with autoview.

src/Interface/Modules/Render/Ospray/OSPRayCamera.cpp
RubioJr9: I'm wrong. The class is called ArcLookAt

jessdtate:
merge conflicts in this PR now.

dcwhite (approved):

jessdtate (approved):

Issue #2006, opened 2019-10-09
Adjust alpha hues in createstandardcolormap, part 3


dcwhite:
Improve the functionality of the alpha graph editor:
- [ ] Delete alpha points
- [ ] Move alpha points
- [ ] Maybe more ideas

dcwhite:
**Analysis (2026-06-13)**

Two alpha-graph editor items remain open: delete points and move points are done. The 'maybe more ideas' item is a placeholder. The alpha editor is part of CreateStandardColorMap — this can be closed or the remaining item replaced with concrete follow-ups.

Issue #1349, opened 2016-03-03
CreateStandardColorMap improvements, part 2: improve performance of GUI


dcwhite:
See #898

dcwhite:
This might be defunct, only appearing on old hardware. Needs more testing. The code is still crappy, however.

dcwhite:
@Haydelj If you have any ideas for improvement of the color map preview generation in Qt, this would be the issue to discuss them.

dcwhite:
To be tested again after the new alpha editor logic is in.

dcwhite:
> This might be defunct, only appearing on old hardware. Needs more testing. The code is still crappy, however.

Still applies after new point GUI changes are in

dcwhite:
**Triage 2026-06:** `CreateStandardColorMap` GUI performance — the colormap preview recomputes and redraws on every slider event. The performance fix is to debounce slider changes (delay redraw by 50ms) using a `QTimer::singleShot`. Low effort; would also fix the new point-GUI regression mentioned in the last comment.

Issue #1127, opened 2015-07-15
SCIRun crashes when changing the colormaps


allywarner:
In the vis lab today, I had the view scene window open and was changing colormaps in CreateStandardColorMap several times and SCIRun crashed. First the view button the the ViewScene module wouldn't work so I deleted it, made a new one and then my pipes wouldn't connect. I reloaded the same network and it worked again. Then I started playing with the color maps and the modules became see through. The graphics card on my computer is an Intel Iris 1536 MB.

dcwhite:
That module has serious performance issues when changing its settings. There's an issue for improving that part of it, which would probably avoid your crash. @cbrightsci

dcwhite:
**Triage 2026-06:** SCIRun crash when changing colormaps rapidly — the stale bot fired twice. Rapid colormap changes while ViewScene is rendering creates a race condition on the colormap object shared between the UI thread and the render thread. The data race fix (#2486) for `SimpleMapModuleState::transientStateMap_` is related. If the colormap is passed through the transient state map, fixing #2486 may also fix this.

PR #2311, opened 2021-07-16
CreateStandardColorMap Refactor

RubioJr9:
Changes:
- Fixed value scaling for alpha values in ColorMap
CreateStandardColorMap changes:
- Set UI size to fixed and changed custom color map buttons
- Fixed value range for color map preview so the selection range is always in [0,1]
- Preview points can be reselected and dragged
- Preview points can be deleted with middle mouse button too, so it can be cancelled when you haven't dropped a point yet
- Preview points are clamped to the nearest neighbors, which solves the problem of having multiple points at the same x value


dcwhite:
Issues to test this with:
- [ ] #2006 
- [ ] #1349 
- [ ] #1127

RubioJr9:
> Issues to test this with:
> 
>     * [ ]  [Adjust alpha hues in createstandardcolormap, part 3 #2006](https://github.com/SCIInstitute/SCIRun/issues/2006)
> 
>     * [ ]  [CreateStandardColorMap improvements, part 2: improve performance of GUI #1349](https://github.com/SCIInstitute/SCIRun/issues/1349)
> 
>     * [ ]  [SCIRun crashes when changing the colormaps #1127](https://github.com/SCIInstitute/SCIRun/issues/1127)

Thanks for linking those. We should be able to close #2006 with this. I'll test performance on my old macbook to see if it's still necessary. I didn't get any crashes on linux, so I'll test Mac

jessdtate (approved):

dcwhite (changes requested):
Screen Shot 2021-12-10 at 2 24 57 PM
Screen Shot 2021-12-10 at 2 22 49 PM

Layout of the UI needs some fixing on Mac (and likely elsewhere)

dcwhite:
Top is PR, bottom is original.

dcwhite (approved):

PR #2317, opened 2021-07-26
Fix boost naming for mac & unix debug builds

RubioJr9:
The platform suffix was missing from the names.
Also for the non-python builds, 'y' needed to be removed from the debug postfix

Tested with & without python on mac and linux


dcwhite:
This will need lots of testing, and also after #2318

jessdtate (approved):

dcwhite (approved):
Tested on M1 arm in release and debug, and both built fine (headless still, of course).

PR #2318, opened 2021-07-28
ViewScene config as pop-up widgets

dcwhite:
The ViewScene configuration window has been broken up into many small widgets that pop-up on hover/click of the now expanded ViewScene toolbar. The color GUI has also been much improved.

Additionally, the first step was taken to convert boost smart pointers to std smart pointers, by updating a typedef. And the C++ standard was set using modern CMake to C++17, which will obviate a few older compilers that a few users might have still been using.


dcwhite:
Forgot to mention the new code uses some of https://github.com/commontk/CTK/, thanks to @jcfr for introducing it.

jessdtate:
I have a couple suggestions:

- I think the blank space in the top left corner between the icons is visually awkward.  Could we move one of the toolbars to the corner? 
- the light positions, mostly the head light, could use a reset button since the it's difficult to set the precise position in the UI. the reset could also include the color if that's easier.  Alternatively, a combo box with the position value would work.  
- Autorotate could use a stop button (in the middle), or by clicking on the same direction

dcwhite:
Autorotate stop/start is on the main button (the one with the icon)

The rest I will add for sure!

jcfr:
Few preliminary nitpicks:
* changes related to c++ standard, use of shared pointer should be done in a dedicated PR
* integration of CTK should be fixed to work using external project instead of copying the sources (idea is to avoid `src/Interface/Modules/Base/CustomWidgets/CTK/Core/ctkValueProxy.h`)

dcwhite:
I agree with both of those points. Reasoning behind the opposite decision: Our PR-review rate is extremely slow, so I tend to lump features and std-refactoring together. Ideally, I would submit multiple PRs on different branches, but if I did that the PRs would stack up even higher with likely a backlog of a year or more. Also, I tried setting up CTK as an external project, we have the cmake machinery for that, but I ran into several problems, which I've forgotten, so this is a "CTK-lite" attempt to just get a few widgets copied in. Normally I wouldn't do that either, but I was in a rush to start using them. See Qwt as an external--I've had many problems with that one over the years, and opted for fast-and-messy for this dependency.

jessdtate (approved):

PR #2339, opened 2021-11-18
One ported module and one quick issue, plus Windows build fix

dcwhite:


src/Modules/Legacy/Fields/ApplyFilterToFieldData.cc
```diff
@@ -25,88 +25,76 @@
...
-        if(!(dilate_algo_.run(input,output))) return;
-      }
-      else
+      
+    }
+    if (doErode)
+    {
+      ErodeFieldDataAlgo erode;
+      erode.set(Variables::MaxIterations, iters);
+      if (!erode.runImpl(input, output))
```
jessdtate, line 90: This implementation will ignore the dilated volume in the case when both both dilate and erode are true

jessdtate (changes requested):

src/Modules/Legacy/Fields/ApplyFilterToFieldData.cc
dcwhite: Let me fix

GPayne (approved):

src/Interface/Modules/Fields/ApplyFilterToFieldDataDialog.cc
```diff
@@ -0,0 +1,50 @@
...
+#include <Dataflow/Network/ModuleStateInterface.h>  ///TODO: extract into intermediate
+#include <Core/Algorithms/Base/AlgorithmVariableNames.h>
+
+using namespace SCIRun::Gui;
+using namespace SCIRun::Dataflow::Networks;
+using namespace SCIRun::Core::Algorithms;
+using namespace SCIRun::Core::Algorithms::Fields;
+
+ApplyFilterToFieldDataDialog::ApplyFilterToFieldDataDialog(const std::string& name, ModuleStateHandle state,
+  QWidget* parent /* = 0 */)
```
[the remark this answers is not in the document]
dcwhite: The default value of `nullptr` is in the header file. This is leftover from probably the original dialog class code, where in the source file the default argument is commented out as a reminder that it is defaulted in the declaration, and that was from ~2012, so no prevalent use of nullptr yet. It's been copy-pasted like this ever since.

jessdtate (approved):

PR #2343, opened 2021-11-21
Composite part 2

dcwhite:
CompositeModuleWithStaticPorts has a useful GUI now and can be tested.

I will make a copy of the module with typed ports instead of wildcard ports, so we can test both at the lunch. Once this one is approved, I'll probably ram that change through since it will not affect any other code.

Still to come (next lunch, not by tomorrow): the subnet module UIs are working but don't have all the connections of a regular UI dialog yet (context menus, button bars, dockability).


GPayne (approved):

jessdtate (approved):
it looks like changes to the UIs in the composite widget are not reflected in the xml in the field, which means they are not saved with the network file.  For the next round I think.

dcwhite:
Correct, I put that (state saving) in the list in https://github.com/SCIInstitute/SCIRun/issues/2344. I might just be able to implement it this morning though...stay tuned...

PR #2345, opened 2021-11-22
Basic python port data functions

dcwhite:
To be elaborated on at the testing lunch


jessdtate (dismissed):
can be hard to find the syntax, but works

GPayne (approved):

PR #2264, opened 2021-04-17
Tensor Class Update and Usage in Glyph Generation

RubioJr9:
Issue #2129

- Swapped in the new tensor class in GlyphGeom and TensorGlyphBuilder.

- Added new functions and constructors to the tensor class.
- Changed the superquadric tensor normal equation to the one from the teem package.
- Generalized the mandel function for tensors of any dimension.
- Tensors now have an OrderState object to keep track of when the eigenvalues are sorted and the eigenvectors are in a right handed coordinate system.
- Extracted the superquadric glyph generation from the function for generating superquadric tensor glyphs, so any portion of the superquadric glyph space can be utilized.
- Changed the EigenSolver for the square matrix case, which is much, much faster.
- Updated and added new test cases.

- Changed the superquadric tensor normal equation to the one from the teem package.
- Removed the unnecessary thresholding in the sign power function in GlyphGeomUtility.

PR #2363, opened 2022-01-10
Fix for Holes in Superquadric Tensor Glyphs

RubioJr9:
This fixes the small holes in the superquadric glyphs. This bug was because the round off was removed. I introduced it in PR #2264. This just reverts that change.


src/Graphics/Glyphs/GlyphGeomUtility.h
```diff
@@ -56,6 +56,11 @@ class SCISHARE GlyphGeomUtility
 
 inline double GlyphGeomUtility::spow(double e, double x)
 {
+  const auto EPSILON = 1.0e-6;
```
dcwhite, line 59:
```suggestion
  static const auto EPSILON = 1.0e-6;
```

dcwhite (changes requested):

jessdtate (approved):

dcwhite (approved):

PR #2369, opened 2022-01-28
Show Uncertainty Glyphs

RubioJr9:
-Added module to visualize tensor field uncertainty using glyphs
-Added uncertainty glyph builder class to tensor glyph builder file
-Added test network
-Moved some files from Core::Algorithm that should be in Graphics::Datatypes

Note: The glyph surface is not mathematically correct, so I put an experimental label in the module


dcwhite:
```
/Users/runner/work/SCIRun/SCIRun/src/Interface/Modules/Visualization/ShowUncertaintyGlyphsDialog.cc:37:77: error: unused parameter 'name' [-Werror,-Wunused-parameter]
ShowUncertaintyGlyphsDialog::ShowUncertaintyGlyphsDialog(const std::string& name,
```

src/Graphics/Glyphs/TensorGlyphBuilder.cc
```diff
@@ -37,6 +37,213 @@ using namespace Core::Geometry;
...
+  const Eigen::Matrix<double, 6, 6>& covarianceMatrix)
+    : TensorGlyphBuilder(t, center), emphasis_(emphasis), covarianceMatrix_(covarianceMatrix)
+{
+  h_ = 0.000001;// * t.frobeniusNorm();
+  hHalf_ = h_ * 0.5;
+}
+
+void UncertaintyTensorOffsetSurfaceBuilder::generateOffsetSurface(GlyphConstructor& constructor)
+{
+  Vector centerVector = Vector(center_);
```
dcwhite, line 51:
```suggestion
  Vector centerVector(center_);
```

src/Modules/Visualization/ShowUncertaintyGlyphs.cc
```diff
@@ -0,0 +1,320 @@
...
+
+ShowUncertaintyGlyphsImpl& ShowUncertaintyGlyphsImpl::setTensorsScale(double scale)
+{
+  scale_ = scale;
+  return *this;
+}
+
+
+// TODO move
+enum FieldDataType
```
dcwhite, line 149: There's already an enum for this, can you use that one?

src/Graphics/Glyphs/TensorGlyphBuilder.h
```diff
@@ -97,6 +110,49 @@ class SCISHARE TensorGlyphBuilder
   int nv_ = 0;
   int nu_ = 0;
 };
+
+  using MandelVector = Eigen::Matrix<double, 6, 1>;
+
+  struct DifftValues
```
dcwhite, line 116: Not a fan of a big struct of public arrays, with a private magic number size. I'll look at the usage and try to suggest something.

dcwhite (changes requested):

src/Modules/Visualization/ShowUncertaintyGlyphs.cc
RubioJr9: It's in 2 other modules. I was planning on moving this, just don't know where to put it

src/Graphics/Datatypes/RenderFieldState.h
```diff
@@ -26,12 +26,11 @@
...
 
-#ifndef CORE_ALGORITHMS_VISUALIZATION_RENDER_FIELD_STATE_H
-#define CORE_ALGORITHMS_VISUALIZATION_RENDER_FIELD_STATE_H
+#ifndef GRAPHICS_DATATYPES_RENDER_FIELD_STATE_H
+#define GRAPHICS_DATATYPES_RENDER_FIELD_STATE_H
 
 #include <Core/Datatypes/Color.h>
-#include <array>
-#include <Core/Algorithms/Visualization/share.h>
+#include <Graphics/Datatypes/share.h>
```
dcwhite, line 33:
```suggestion
#include <array>
#include <Graphics/Datatypes/share.h>
```

Also needed for Windows

src/Modules/Visualization/ShowUncertaintyGlyphs.cc
dcwhite: You can reuse `Mesh::NODES_E, Mesh::FACES_E`, and the like instead.

RubioJr9:
I'm going to add comments to explain the uncertainty glyph's variable naming soon. I wanted to get the builds going first.

jessdtate (commented):
It seems that the major "un-named" variables are qn and P.  qn seems to be a discretization of q or Q from the paper.    Does that match what see @RubioJr9?  if so, maybe it would be better as q_i to be more consistent with the paper?  p is a point/angle on the glyph, right?  I think it's fine to leave it as p in that case.  

I didn't see others that weren't part of loops.

RubioJr9:
> It seems that the major "un-named" variables are qn and P. qn seems to be a discretization of q or Q from the paper. Does that match what see @RubioJr9? if so, maybe it would be better as q_i to be more consistent with the paper? p is a point/angle on the glyph, right? I think it's fine to leave it as p in that case.
> 
> I didn't see others that weren't part of loops.

I don't think it is a discretization. I think the qn variable is a measure of finite differences in each of the 6 degrees of freedom of the mean tensor.

dcwhite:
@jessdtate Just waiting on your approval on this one too

dcwhite (approved):

jessdtate (approved):

PR #2390, opened 2022-07-11
ViewScene pop-ups improvement

dcwhite:
ViewScene toolbars are now movable and floatable, and the pop-up widgets respect position of the toolbar as well as fullscreen state. If you want popups to pop a different direction (like when the toolbar is near the bottom of the screen and the popups are going off the screen), there is a button that will make the popups pop the opposite way (the green arrow button). I believe I also fixed the bug where multiple popups would appear when not requested. This should solve most users' issues.


jessdtate (approved):

Sumientra:
Works much better now. Windows don't pop up unless you hover for a second. They show up in a usable location. Just a few things not completely working:

1. "Look down axis" menu didn't do anything when I tested it.
Screen Shot 2022-07-27 at 23 01 25
2. It would be helpful for the lighting menu to have a reset button. If you screw up the values now it's hard to get them back to something usable.
Screen Shot 2022-07-27 at 23 03 59
3. It would be helpful if the scale bar menu had a color option. The bar is currently white, which is not visible on a white background. 
Screen Shot 2022-07-27 at 23 04 51

dcwhite:
Thanks for testing. I can definitely work on all those for the upcoming ZZZ release.

PR #2414, opened 2023-01-31
Mac build fixes

dcwhite:
Deletes/updates a few old usages of sprintf et al.


src/Core/Utils/Legacy/FileUtils.cc
```diff
@@ -64,123 +64,6 @@ namespace bsys=boost::system;
 
 namespace SCIRun {
 
-////////////////////////////////////////////////////////////
```
jessdtate, line 67: Are these handled by Qt now?

jessdtate (approved):

GPayne (approved):

src/Core/Utils/Legacy/FileUtils.cc
dcwhite: Nothing was calling it

PR #2428, opened 2023-06-23
OSPRay Upgrade to 2.10

tarkpate:
Links to git tag from the [PR](https://github.com/CIBC-Internal/ospray/pull/7) branch for testing purposes. After that PR is merged, I'll revert the git repo back to cibc-internal.

Also, I added a include string statement in Mutex. I was getting errors on both of my machines that it was missing when I tried to build with OSPRay.

### DIFF
--- a/Superbuild/OsprayExternal.cmake
+++ b/Superbuild/OsprayExternal.cmake
@@ -25,7 +25,8 @@
 #  DEALINGS IN THE SOFTWARE.
 
 SET_PROPERTY(DIRECTORY PROPERTY "EP_BASE" ${ep_base})
-SET(ospray_GIT_TAG "origin/scirun-build-2.4")
+# SET(ospray_GIT_TAG "origin/scirun-build-2.4")
+SET(ospray_GIT_TAG "origin/cibc-2.10")
 
 set(ospray_DEPENDENCIES)
 LIST(APPEND ospray_DEPENDENCIES GLM_external)
@@ -34,7 +35,8 @@ LIST(APPEND ospray_DEPENDENCIES GLM_external)
 # git checkout -q will silence message about detached head (harmless).
 ExternalProject_Add(Ospray_external
   DEPENDS ${ospray_DEPENDENCIES}
-  GIT_REPOSITORY "https://github.com/CIBC-Internal/ospray.git"
+  # GIT_REPOSITORY "https://github.com/CIBC-Internal/ospray.git"
+  GIT_REPOSITORY "https://github.com/tarkpate/ospray.git"
   GIT_TAG ${ospray_GIT_TAG}
   PATCH_COMMAND ""
   INSTALL_DIR ""
--- a/Superbuild/OsprayExternal.cmake
+++ b/Superbuild/OsprayExternal.cmake
@@ -25,7 +25,6 @@
 #  DEALINGS IN THE SOFTWARE.
 
 SET_PROPERTY(DIRECTORY PROPERTY "EP_BASE" ${ep_base})
-# SET(ospray_GIT_TAG "origin/scirun-build-2.4")
 SET(ospray_GIT_TAG "origin/cibc-2.10")
 
 set(ospray_DEPENDENCIES)
--- a/src/Core/Thread/Mutex.h
+++ b/src/Core/Thread/Mutex.h
@@ -32,6 +32,7 @@
 #include <boost/noncopyable.hpp>
 #include <thread>
 #include <mutex>
+#include <string>
 #include <Core/Thread/share.h>
 
 namespace SCIRun


Superbuild/OsprayExternal.cmake
```diff
@@ -34,7 +35,8 @@ LIST(APPEND ospray_DEPENDENCIES GLM_external)
 # git checkout -q will silence message about detached head (harmless).
 ExternalProject_Add(Ospray_external
   DEPENDS ${ospray_DEPENDENCIES}
-  GIT_REPOSITORY "https://github.com/CIBC-Internal/ospray.git"
+  # GIT_REPOSITORY "https://github.com/CIBC-Internal/ospray.git"
+  GIT_REPOSITORY "https://github.com/tarkpate/ospray.git"
```
jcfr, line 38: I suggest to maintain official changes associated with `SCIRun` into a fork of `ospray` added to the https://github.com/SCIInstitute organization.

For guidance on naming convention for the branches et alike.

See these examples:
* https://github.com/search?q=topic%3Akitware-fork+org%3AKitwareMedical+fork%3Atrue&type=repositories

tarkpate: The CIBC-Internal repo is a part of the SCI Institute. There is a [PR](https://github.com/CIBC-Internal/ospray/pull/7) to put this branch in the OSPRay for there. Using my repo's branch is temporary for testing purposes.

jcfr: Sounds great. Thanks for the follow-up and clarification.

Superbuild/OsprayExternal.cmake
```diff
@@ -25,7 +25,7 @@
 #  DEALINGS IN THE SOFTWARE.
 
 SET_PROPERTY(DIRECTORY PROPERTY "EP_BASE" ${ep_base})
-SET(ospray_GIT_TAG "origin/scirun-build-2.4")
+SET(ospray_GIT_TAG "origin/cibc-2.10")
```
dcwhite, line 28:
```suggestion
SET(ospray_GIT_TAG "origin/scirun-build-2.10")
```

Superbuild/OsprayExternal.cmake
```diff
@@ -34,7 +34,8 @@ LIST(APPEND ospray_DEPENDENCIES GLM_external)
 # git checkout -q will silence message about detached head (harmless).
 ExternalProject_Add(Ospray_external
   DEPENDS ${ospray_DEPENDENCIES}
-  GIT_REPOSITORY "https://github.com/CIBC-Internal/ospray.git"
+  # GIT_REPOSITORY "https://github.com/CIBC-Internal/ospray.git"
+  GIT_REPOSITORY "https://github.com/tarkpate/ospray.git"
```
dcwhite, line 38:
```suggestion
  GIT_REPOSITORY "https://github.com/CIBC-Internal/ospray.git"
```

dcwhite (changes requested):
Now that the CIBC-internal repo is updated, let's rerun this one

jcfr:
Since `origin/scirun-build-2.10` is a branch, I would instead suggest to do the following:

```
SET(ospray_GIT_TAG "29a6d246370fd3ec208d3668c84240403ab45ba9") # scirun-build-2.10
```

That say, different build of scirun will be ensured to build against the same version of ospray. That approach should likely be used for all dependencies.

Also, if possible, I suggest to submit pull request to the upstream ospray project and then backport the corresponding commit into the `CIBC-Internal` fork prefixing it with `[Backport MR-NNNN]` where `NNNN` is the pull request number.

dcwhite:
I can make a named tag for our 2.10 version. There was no other activity on those repo copies, so no "danger" in using branches instead of tags. I don't think Ospray wants any of our changes (like deleting all the documentation), why would we send them a PR?

dcwhite (approved):

jcfr:
> I don't think Ospray wants any of our changes

The following could be done:

1. Create an issue upstream explaining that having a sub-module to include the documentation leads to additional work when integrating in client project like SCIRun. You could suggest the following:
  * Remove the [gitmodules](https://github.com/ospray/ospray/blob/master/.gitmodules) as it doesn't seems to exist. 
  * Use the [FetchContent](https://cmake.org/cmake/help/v3.11/module/FetchContent.html#examples) introduced in CMake 3.11. Since the CMake minimum required version in OSPRay is CMake 3.1, a fallback can easily be added to support older version.

2. Submit a pull-request to rename include `build_ospray` instead of `build_ospray.cmake` as well as using `CMAKE_CURRENT_LIST_DIR` instead of `CMAKE_CURRENT_SOURCE_DIR`

3. `OSPRAY_INSTALL_DEPENDENCIES` should probably be set before including `OSPRay` in `SCIRun` instead of being hard-coded in the project itself. This will ensure that building against another version of `OSPRay` will work as expected.

4. Instead of modifying `OSPRay` to define the project `SCIOspray`, it may be worth adding a folder called `SCIOspray` in `SCIRun`. The idea would be to minimize updates to the dependency and streamline future updates.

jessdtate (approved):

PR #2430, opened 2023-07-14
Build with Local Ospray

tpat95:
I added a CMake flag that searches for a prebuilt version of OSPRay. You just download and add the directory to the path. Also added instructions in the doc.


tarkpate:
I made this change so I can use a local build of OSPRay so I can edit and recompile separately, but it looks like it works with prebuilt binaries. We may be able to use this to replace the default action of compiling OSPRay from scratch with just a simple binary download and link. We also wouldn't need to fork and edit for each version release.

dcwhite:
This would be very useful.

src/Interface/Modules/Render/Ospray/OSPRayRenderer.cpp
```diff
@@ -46,7 +46,7 @@ using namespace Render;
 using namespace Core::Datatypes;
 using namespace Core::Geometry;
 
-#ifdef WITH_OSPRAY
+#if defined(WITH_OSPRAY) || defined(WITH_OSPRAY_LOCAL)
```
dcwhite, line 49: Can you define this disjunction in CMake? Then no C++ files need to change.

WITH_OSPRAY --> BUILD_OSPRAY
WITH_OSPRAY_LOCAL --> PREBUILT_OSPRAY
Then add cmake for WITH_OSPRAY = BUILD_OSPRAY || PREBUILT_OSPRAY
A few cmake files will need to change, but no C++

dcwhite (changes requested):

src/Interface/Modules/Render/Ospray/OSPRayRenderer.cpp
tarkpate: Good idea, I'll get this changed

dcwhite: @tarkpate I could handle this change if you are too busy.

tarkpate: Just updated with these new flags

dcwhite: Ha, I suggested invalid cmake (it was just pseudocode-cmake after all), so that failed the configure stage. The update is in

dcwhite (approved):

PR #2431, opened 2023-08-03
Python List to 1D Matrix

tarkpate:
Currently, a 1D python list is ignored as Matrix output. With this change, a 1D array converts to a Matrix of Nx1


dcwhite:
@tarkpate Could you write a unit test for this? Either C++ or Python would work

jessdtate (approved):

dcwhite:
@tarkpate Still waiting on some test code. Even an example network would work

Issue #1720, opened 2017-10-20
Docker


dcwhite:
Details later...

dcwhite:
Jon found a Maya Docker image with NVidia examples.

dcwhite:
@jessdtate Beta or post-beta?

jessdtate:
Post-beta

dcwhite:
**Analysis (2026-06-12)**

Docker image for SCIRun was flagged as post-beta work. No Dockerfile or CI Docker job exists yet.

For a headless/CI image the main challenge is bundling Qt (or building headless) + the Superbuild externals in a reproducible layer. A multi-stage build (Superbuild externals in stage 1, SCIRun build in stage 2) would keep image size reasonable.

No blockers identified beyond prioritization. Could leverage the same sccache/Superbuild caching approach added for #1487.

PR #2438, opened 2024-12-02
Only initialize the CLI for headless mode and add Dockerfile

georgiastuart:
I've been helping a group use SCIRun on our local cluster. As part of that work, I made a couple changes to SCIRun that may be useful to others. Namely...

1. Containerization was easier than keeping SCIRun happy with Cluster QT offerings, particularly after a cluster-wide OS update. I created a Dockerfile, which should be broadly useful.
2. SCIRun refused to start within SLURM batch jobs or CLI interactive jobs and would only run if it was either (a) compiled completely headless or (b) run within our GUI desktop, even when `-x` was specified, since QT always tried to identify a display. To solve this, I added a check for `-x` or `-h` to the Main function to initialize the command line program instead. 

Seems to work fine locally. Tested on an Ubuntu 24.04 system using the Apptainer runtime from a docker image.

### DIFF
--- a/src/Main/scirunMain.cc
+++ b/src/Main/scirunMain.cc
@@ -65,7 +65,11 @@ int mainImpl(int argc, const char* argv[], char **environment)
   //TODO: must read --headless flag here, or try pushing command queue building all the way up here
   //TODO: https://doc.qt.io/qt-5/qapplication.html#details
 #ifndef BUILD_HEADLESS
-  return GuiApplication::run(argc, argv);
+  if (!Application::Instance().parameters()->disableGui()) {
+    return GuiApplication::run(argc, argv);
+  } else {
+    return ConsoleApplication::run(argc, argv);
+  }
 #else
   return ConsoleApplication::run(argc, argv);
 #endif
--- a/src/Main/scirunMain.cc
+++ b/src/Main/scirunMain.cc
@@ -65,7 +65,7 @@ int mainImpl(int argc, const char* argv[], char **environment)
   //TODO: must read --headless flag here, or try pushing command queue building all the way up here
   //TODO: https://doc.qt.io/qt-5/qapplication.html#details
 #ifndef BUILD_HEADLESS
-  if (!Application::Instance().parameters()->disableGui()) {
+  if (!Application::Instance().parameters()->disableGui() && !Application::Instance().parameters()->help()) {
     return GuiApplication::run(argc, argv);
   } else {
     return ConsoleApplication::run(argc, argv);
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+.github
+scripts
+.clang-format
+.git 
+.gitattributes
+.gitignore
+.travis.yml
+build.sh
+compile_commands.json
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+#syntax=docker/dockerfile:1-labs
+FROM ubuntu:24.04
+
+ARG SCIRUN_VERSION="v5.0-beta.2023"
+
+RUN apt-get -y update \
+    && DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get -y install qtbase5-dev libqt5svg5-dev \
+    cmake git openssl tzdata build-essential g++ zlib1g-dev libssl-dev libncurses-dev libsqlite3-dev \
+    libreadline-dev libtk-img-dev libgdm-dev libdb-dev libpcap-dev \
+    && rm -rf /var/cache/apt/archives /var/lib/apt/lists/*
+
+WORKDIR /scirun-src
+COPY . /scirun-src/
+# RUN git clone --depth 1 --branch ${SCIRUN_VERSION} https://github.com/georgiastuart/SCIRun.git .
+
+WORKDIR /opt/scirun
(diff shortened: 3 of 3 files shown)


georgiastuart:
I also just pushed a tweak to the docs. Ubuntu 22.04 and later needs different QT5 package names, and I found that I needed to use `-DQt_PATH` for QT 5 instead of `-DQt5_PATH` or `-DQt5_DIR`. I'm happy to modify or revert if this isn't correct across the board, though.

dcwhite:
Thank you for the Docker file, that was a long-needed feature. I'll check the Windows and Mac builds and then we can probably merge this.

dcwhite:
Here's the issue, only 7 years old https://github.com/SCIInstitute/SCIRun/issues/1720

georgiastuart:
I think the mac build failed because of the `install-qt-action` version being out of date. Bumped to v4.

dcwhite:
> I think the mac build failed because of the `install-qt-action` version being out of date. Bumped to v4.

We might need 4.1 to fix the failure https://github.com/jurplel/install-qt-action/releases/tag/v4.1.0

georgiastuart:
Hmm, `@v4` should grab the latest in that major version series, so I think it's something else. I'm going to let the rest of the checks finish and then investigate further.

dcwhite:
According to https://github.com/orgs/community/discussions/48058 and https://github.com/jurplel/install-qt-action/tags, I think in this case it's using the older 4.0.0 tag, and the error looks exactly like 4.1.0's fix. I thought it would grab the latest too, but it seems to be repo-specific. Good to see SCIRun's rusty wheels getting some exercise, at least.

dcwhite:
While you're updating action versions, would you mind bumping upload-artifact too? https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

georgiastuart:
I think it is pulling 4.1, that MacOS update refers to this removed brew install, not the pip install: https://github.com/jurplel/install-qt-action/commit/cb07ed2575f3d0f7ea333707339b2edac669fed1

Looks like it's this and the fix is to let the action manage python, which seems OK since SCIRun builds its own Python: https://github.com/jurplel/install-qt-action/issues/239

Updating that and the request above.

dcwhite:
Hah I hope you're OK with using this PR as a github action cleanup opportunity. The new Mac error is this old line that should not be needed anymore: https://github.com/SCIInstitute/SCIRun/pull/2438/files#diff-38c2966d1144b82978ca2cfe7650879fb15bd6da5845a3c616e8d152b29317f1R90

It was added in 2019 probably to fix some other problem (the commit message gave no details), so it can be deleted. It's in four places in the mac yml file.

src/Main/scirunMain.cc
```diff
@@ -65,7 +65,11 @@ int mainImpl(int argc, const char* argv[], char **environment)
   //TODO: must read --headless flag here, or try pushing command queue building all the way up here
   //TODO: https://doc.qt.io/qt-5/qapplication.html#details
 #ifndef BUILD_HEADLESS
-  return GuiApplication::run(argc, argv);
+  if (!Application::Instance().parameters()->disableGui() && !Application::Instance().parameters()->help()) {
+    return GuiApplication::run(argc, argv);
+  } else {
+    return ConsoleApplication::run(argc, argv);
+  }
```
dcwhite, line 72:
```suggestion
  if (Application::Instance().parameters()->disableGui() || Application::Instance().parameters()->help()) 
  {
    return ConsoleApplication::run(argc, argv);
  }

  return GuiApplication::run(argc, argv);

```

Could you flip the if statement, it's clearer without the negations.

georgiastuart:
Np on the cicd cleanup opportunity! I removed the symlink lines and (I think) fixed the Windows Qt6 error (the Qt6_PATH env var is no longer set by the install action it looks like). 

If y'all have a place to park a Docker image, I can throw in a Docker build workflow.

.github/workflows/mac.yml
```diff
@@ -17,23 +17,19 @@ jobs:
...
 
     - name: Print out Qt directory
       run: echo $Qt5_DIR; echo $Qt5_Dir
 
-    - name: Prepare
-      run: |
-        ln -s /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/System/Library/Frameworks/Tk.framework/Versions/8.5/Headers/X11/ /usr/local/include/X11
-
     - name: make
       run: ./build.sh -DQt_PATH="${Qt5_Dir}"
```
dcwhite, line 34: Progress! Now the configure is failing, this should be changed to `${Qt_dir}` or path or whichever it is now...

dcwhite:
For now you could try putting something here, I just created it: https://hub.docker.com/repository/docker/functoire/scirun/general
If something is available there, we can get some SCI person to take over the publishing part (I'm no longer employed there).

dcwhite:
Meanwhile the headless Mac build is failing in Boost code, which would require an upgrade to Boost 1.86. Now that I don't expect you to do, unless you really want to take over as a maintainer...

dcwhite:
I got it started by updating the Python repo. Boost will be in 2025, yeah. Happy holidays!

PR #2564, opened 2026-07-16
Fix dyld crash on macOS < 15 from unguarded quick_exit reference

dcwhite:
## Problem

`SCIRun_test` aborts on load on macOS 14:

```
dyld[48953]: Symbol not found: _quick_exit
  Referenced from: .../bin26/SCIRun/lib/libInterface_Application.dylib
  Expected in:     /usr/lib/libSystem.B.dylib
```

## Cause

`quick_exit` only entered libSystem in **macOS 15.0**, and the SDK declares it with **no availability annotation**:

```c
void	quick_exit(int) __dead2;        // _stdlib.h:160
```

It's gated on `__DARWIN_C_LEVEL`/C11 — a *compile-time* feature gate, not an OS-version gate. So even with a correct deployment target (`minos 14.5`, confirmed via `otool -l`), the compiler has no version information to act on: it can't emit `-Wunguarded-availability`, and it can't weak-link. The result is a **hard** reference that links cleanly against the SDK's libSystem stub and fails only at load time.

libc++ forwards the same un-annotated declaration (`using ::quick_exit _LIBCPP_USING_IF_EXISTS;`), so `std::quick_exit` inherits it.

Xcode 16 ships only the macOS 15 SDK, so **on macOS 14 this silently produces a binary that cannot load on the machine that built it** — with zero warnings. This reached us through the two regression-mode exits added in a0e09547e.

## Fix

Extracts the call behind `Core::quickExit` ([`src/Core/Utils/QuickExit.h`](src/Core/Utils/QuickExit.h)), falling back to `_Exit` when the deployment target is below macOS 15:

```cpp
#if defined(__APPLE__) && (__MAC_OS_X_VERSION_MIN_REQUIRED < 150000)
  #define SCIRUN_NO_QUICK_EXIT 1
#endif
```

`_Exit` is C99 and always present. Since nothing in the tree registers `at_quick_exit` handlers, the two are behaviorally equivalent here — that constraint is documented in the header, as it's the one thing that could make the branches diverge.

The literal `150000` is deliberate rather than the nicer-reading `__MAC_15_0`: that constant is undefined on pre-15 SDKs and would expand to `0`, inverting the test and reintroducing the crash on exactly the configurations the guard protects.

## Verification

On macOS 14.5 / Xcode 16.2:

- `_quick_exit` references remaining in the build: **0** (full-build `nm` rescan)
- **Both** preprocessor branches compile — `min 14.5` → `__Exit`, `min 15.0` → `_quick_exit`. The `quick_exit` branch is dead on current hardware and would otherwise not compile until someone bumps the deployment target, so this guards against bit-rot.
- `SCIRun_test` loads and runs; regression network executes and exits 0
- lldb confirms the live path rather than just the symbol table:
  ```
  frame #0: libsystem_c.dylib`_Exit
  frame #1: libInterface_Application.dylib`SCIRun::Core::quickExit(int) + 12
  Process exited with status = 0
  ```

## Note

Worth a reviewer's judgment: `_Exit` is available on every platform SCIRun targets, so the conditional buys intent-signaling and auto-healing when the minimum target rises — not different behavior. Dropping to an unconditional `_Exit` would be simpler and equally correct today.

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -90,6 +90,7 @@ jobs:
       runner: macOS-latest
       variant: headless
       build-testing: true
+      run-smoke-test: true
 
   mac-headless-14-arm:
     uses: ./.github/workflows/reusable-build.yml
@@ -97,6 +98,7 @@ jobs:
       runner: macOS-14
       variant: headless
       build-testing: true
+      run-smoke-test: true
 
   mac-headless-14-arm-slim:
     uses: ./.github/workflows/reusable-build.yml
@@ -106,6 +108,7 @@ jobs:
       build-testing: true
       build-with-python: false
       run-unit-tests: true
+      run-smoke-test: true
 
   mac-headless-15-arm:
     uses: ./.github/workflows/reusable-build.yml
@@ -113,6 +116,7 @@ jobs:
       runner: macOS-15
       variant: headless
       build-testing: true
+      run-smoke-test: true
 
   mac-headless-15-arm-slim:
     uses: ./.github/workflows/reusable-build.yml
@@ -122,6 +126,7 @@ jobs:
       build-testing: true
       build-with-python: false
       run-unit-tests: true
+      run-smoke-test: true
 
   mac-headless-15-intel:
     uses: ./.github/workflows/reusable-build.yml
@@ -129,6 +134,7 @@ jobs:
       runner: macOS-15-intel
       variant: headless
       build-testing: true
+      run-smoke-test: true
 
   mac-headless-15-intel-slim:
     uses: ./.github/workflows/reusable-build.yml
@@ -138,6 +144,7 @@ jobs:
       build-testing: true
       build-with-python: false
       run-unit-tests: true
+      run-smoke-test: true
 
   mac-headless-26:
     uses: ./.github/workflows/reusable-build.yml
@@ -145,3 +152,4 @@ jobs:
       runner: macOS-26
       variant: headless
       build-testing: true
+      run-smoke-test: true
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -39,6 +39,27 @@ jobs:
       build-with-python: true
       artifact-name: SCIRunMacInstaller-26-intel
 
+  # Oldest available macOS image, built with the GUI. Every other mac-gui job runs
+  # on macOS-latest (currently the macos-26 image), and the only macOS 14 jobs are
+  # headless -- but src/CMakeLists.txt skips ADD_SUBDIRECTORY(Interface) when
+  # BUILD_HEADLESS is on, so without this job nothing compiles src/Interface for an
+  # older macOS at all. That gap let a macOS-15-only libSystem symbol (quick_exit)
+  # link cleanly on the 26 runners and abort at load on macOS 14 (PR #2564).
+  #
+  # Pinned deliberately to the oldest image rather than macOS-latest: running the
+  # newest macOS is already covered several times over, and only the oldest one
+  # exercises the lower deployment target. macOS-14 is deprecated by GitHub, so when
+  # it is retired this should be bumped to the next-oldest image, not deleted.
+  mac-gui-14-arm:
+    uses: ./.github/workflows/reusable-build.yml
+    with:
+      runner: macOS-14
+      qt-version: '6.10.2'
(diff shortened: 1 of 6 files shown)


dcwhite:
## Note for reviewers: CI cannot validate this fix

Worth flagging explicitly — **green CI on this PR does not exercise the code path it fixes.** I traced the runner images rather than going by job name:

| job | runs on | builds `src/Interface`? |
|---|---|---|
| `mac-gui`, `mac-gui-nonpython`, `mac-gui-regression` | macOS 26 (`macOS-latest` → `macos-26-arm64`) | yes |
| `mac-gui-26`, `mac-gui-26-intel` | macOS 26 | yes |
| `mac-headless-14-arm`, `mac-headless-14-arm-slim` | macOS 14 | **no** |

There is no "GUI build on macOS < 15" job, for two independent reasons:

1. Every GUI job runs on macOS 26, where `quick_exit` **is** present in libSystem — so the crash cannot manifest.
2. The only macOS 14 jobs are `variant: headless`, and `src/CMakeLists.txt:632` guards `ADD_SUBDIRECTORY(Interface)` behind `IF(NOT BUILD_HEADLESS)` — so the code referencing `quick_exit` is never compiled there.

This is exactly how a0e09547e landed with fully green checks while breaking local macOS 14 builds. It also means the `_Exit` fallback added here is **not** covered by any job on this PR; the evidence for it is local (`nm` showing zero `_quick_exit` refs, and lldb confirming `SCIRun::Core::quickExit(int)` → `libsystem_c.dylib\`_Exit` with the exit code propagated on macOS 14.5 / Xcode 16.2).

### What CI *does* cover

One useful accident: the deployment target is not pinned anywhere, so it auto-detects from the runner. The macOS 26 build log confirms:

```
checking which MACOSX_DEPLOYMENT_TARGET to use... 26.4
```

`260400 >= 150000`, so **CI compiles the `std::quick_exit` branch** of the new conditional, while the local macOS 14.5 build compiles the `_Exit` branch. Between the two, both branches are exercised on real hardware — but only by coincidence, not by design.

### Follow-up

The gap is worth closing separately: a GUI build on the oldest supported macOS runner. Note that a *build* alone is insufficient — this class of bug is a **load-time** dyld failure, so the job has to actually execute the binary to catch it.

dcwhite:
**#2537 depends on this.**

#2537 adds a second `quick_exit` call site in `Core/ConsoleApplication/ConsoleCommands.cc` and is currently failing `mac-headless-14-arm` on it (`use of undeclared identifier 'quick_exit'`). Its workaround (`fe316899e`, dropping the `std::` qualifier) doesn't hold up: older SDKs don't declare `quick_exit` in either namespace, so both spellings fail to compile — `std::quick_exit` as an unresolved using declaration, bare `quick_exit` as an undeclared identifier.

Once this merges, #2537 should rebase, revert `fe316899e`, and use `quickExit(1)` from `Core/Utils/QuickExit.h`. That clears its macOS-14 builds and keeps the console path from emitting the hard `_quick_exit` reference into `libCore_ConsoleApplication`.

dcwhite:
**Note on the smoke-test coverage here — and why #2565 is still worth doing.**

The smoke test builds and runs on the same runner (`reusable-build.yml` runs the just-built `bin/SCIRun/SCIRun_test --help` as a step in the build job). So there's no skew between the build SDK and the runtime OS: a symbol newer than the floor resolves fine at load, because the machine that emitted the reference is the machine executing it. No single job can reproduce the dyld form of this bug, which by construction needs a 15-SDK build to meet a pre-15 machine.

`mac-gui-14-arm` does catch *this* bug, but through a different door — as a compile error rather than a load failure. The macOS 14 SDK doesn't declare `quick_exit` at all, so it dies at the compile step and never reaches the smoke test. Still a real net, and worth having for the `src/Interface` coverage gap it closes.

The limit is that it makes the effective coverage floor "the oldest available GitHub image", which drifts upward on GitHub's schedule rather than ours — macOS-14 is already deprecated, as the comment in `mac.yml` notes. Anything below that floor is never exercised, and users on older Macs sit exactly there.

#2565 proposes pinning `CMAKE_OSX_DEPLOYMENT_TARGET` so the declared floor is ours and stops moving when an image is retired. Pinned, `-Wunguarded-availability` fires at compile time on *any* runner, so catching this class stops depending on having an old enough machine to run the binary on.

jessdtate (approved):

PR #2566, opened 2026-07-20
Fix Python_external Windows build when aka.ms/nugetclidl redirects to Bing

dcwhite:
## Summary
- The CPython build scripts (`PCbuild/find_python.bat`) fetch `nuget.exe` from the `aka.ms/nugetclidl` shortlink. That shortlink is currently redirecting to a Bing search results page instead of the NuGet installer, so PowerShell saves the HTML response as `nuget.exe`, which then fails to execute ("The system cannot execute the specified program").
- Because the script only downloads `nuget.exe` if it doesn't already exist, the corrupted file gets reused on every retry, and the failure cascades: `Python_external` fails, `Boost_external`'s configure step fails (it looks for `python.exe` for Boost.JSON's pretty-printer support), and finally `SCIRun_external`'s configure fails with `Boost 'atomic' not found for architecture x64` since Boost never actually built.
- Fix: override `NUGET_URL` in the `Python_external` `CONFIGURE_COMMAND` to point directly at `https://dist.nuget.org/win-x86-commandline/latest/nuget.exe`, bypassing the broken shortlink. Verified this URL returns a valid PE executable (`MZ` header, `application/x-msdownload`).

## Test plan
- [ ] Fresh Superbuild configure + build on Windows with an empty build directory, confirm `Python_external` downloads a valid `nuget.exe` and completes its configure step
- [ ] Confirm `Boost_external` and `SCIRun_external` proceed past configure once Python is available


jessdtate (approved):

jessdtate:
A lot of windows builds are red already.  Looks like all the ones with python

dcwhite:
Oops this one broke a lot of the non-required builds--I will fix it

Issue #2567, opened 2026-07-20
editMeshBBox_trans_check regression network crashes (use-after-free) when ViewScene dock is shown during load


dcwhite:
## Summary

The regression network `editMeshBBox_trans_check.srn5` crashes (segfault) reliably when run in regression mode on Windows. The crash is a **use-after-free / heap corruption**, and it detonates when the ViewScene dock widget is shown during network load. It is **not** caused by the ViewScene show path itself — that path is the victim; the corruption is introduced earlier while loading this specific network.

This was previously invisible because the local Windows regression tests could not run at all (they died with `0xc0000135` DLL-not-found before reaching any network). After fixing the ctest DLL/Qt environment so regression tests run locally (branch `windows-unit-test-fixes-phase2`), 308/309 pass — this network is the only remaining failure.

## Reproduction

```
ctest -C Release -R "\.Test\.ExampleNetwork\.editMeshBBox_trans_check_srn"
```

or directly:

```
SCIRun.exe -E src/ExampleNets/regression/Modules/editMeshBBox_trans_check.srn5 --no_splash --regression 60 -d <SCIRunTestData>
```

Crashes 4/4 times in isolation. Deterministic for this network; other ViewScene networks and other `EditMeshBoundingBox` networks (`editMeshBBox1`, `embb2`, `embb3`, `embbSimple`, `transformMesh`, …) all pass.

## Crash location

Symbolized call stack (debug build of `Interface_Application`):

```
NetworkEditorController::loadNetwork
 → NetworkEditorControllerGuiProxy::moduleAdded
 → NetworkEditor::addModuleWidget
 → ModuleWidget::ModuleWidget
 → ModuleWidget::makeOptionsDialog            (inline)
 → ModuleOptionsDialogConfiguration::config
 → ModuleOptionsDialogConfiguration::configDockable   // ModuleOptionsDialogConfiguration.cc:111
      dockable->show();   // ViewScene && isRegressionMode() branch
 → QWidget::setVisible → QWidgetPrivate::create → QWindowPrivate::create
 → QGuiApplicationPrivate::processGeometryChangeEvent → QWindow::xChanged (signal)
 → QMetaObject::activate  →  reads wild pointer  →  ACCESS VIOLATION
```

Faulting instruction: `test byte ptr [rdi+20h], 2` while `QMetaObject::activate` iterates a connection list.

## Why this is heap corruption (not a logic bug)

The faulting address **varies between runs**:
[...]

PR #2572, opened 2026-07-22
Windows test fixes phase 2: unit 50→54/56, regression 0→308/309, and fix a 2014 data race (#645)

### LINKED ISSUE #645: Random memory bug in ParallelLinearAlgebra
This shows up sometimes in the unit tests:

```
[ RUN      ] SolveLinearSystemModuleTest.CanSolveSimple
2014-09-24 15:24:42.830  [DEBUG] Module created: SolveLinearSystem with id: SolveLinearSystem:22
2014-09-24 15:24:42.830  [DEBUG] Module algorithm initialized: SolveLinearSystem
Remark: Running algorithm Parallel cg Solver with tolerance 1e-05 and maximum iterations 0
Remark: Using preconditioner: Jacobi
libc++abi.dylib: libc++abi.dylib: libc++abi.dylib: libc++abi.dylib: libc++abi.dylib: libc++abi.dylib: terminate called throwing an exceptionlibc++abi.dylib: libc++abi.dylib: terminate called throwing an exceptionterminate called throwing an exceptionterminate called throwing an exceptionterminate called throwing an exceptionterminate called throwing an exception
terminate called throwing an exceptionAbort trap: 6
```

dcwhite:
Follow-on to #2535. Gets the Windows test suites actually running and green, and fixes a long-standing data race found along the way.

Measured on a local Windows/MSVC/Qt5.15 Release build with `BUILD_TESTING=ON`:

| | before | after |
|---|---|---|
| Unit tests | 50/56 | **54/56** |
| Regression tests | 0/309 (none could run) | **308/309** |

The two remaining unit failures are `Engine_Python_Tests` and `Core_Python_Tests`, both fixed by the still-open #2551. The one remaining regression failure is #2567, tracked separately.

---

## 1. Mesh facade tests asserted a hash-dependent edge order

`TriSurfMeshFacadeTests` hardcoded a specific edge numbering. Edge indices are assigned while iterating a `std::unordered_map` in `TriSurfMesh::compute_edges_bugfix`, so the numbering is implementation-defined and shifted when that map moved from boost to std.

Rather than re-pin the numbers to whatever MSVC produces (which would just break libstdc++/libc++ next), the tests now compare the *set* of edges as canonical sorted node pairs and sort each face/node's incident-edge list.

## 2. Legacy importer test could not find its XML

`LegacyNetworkFileImporterTests` located `LegacyModuleImporter.xml` via `rootDir()/"../../src/..."`, which assumes the test-data directory sits two levels below the source tree. When `SCIRUN_TEST_RESOURCE_DIR` points outside the checkout, `SetUpTestCase` threw `"read error"` and **all 15 tests silently skipped**.

Adds a `TEST_SOURCE_ROOT_DIR` compile definition and `TestResources::sourceRootDir()`. All 15 now run and pass.

## 3. Streamline tests asserted exact point counts

`ManySeedsMultithreaded` compared exact node/elem/value counts. The adaptive (RungeKuttaFehlberg) and CellWalk integrators produce counts that vary with floating-point rounding, so they drifted per toolchain.

Confirmed this is **not** a race: single-threaded and multi-threaded runs produce identical counts (RKF 214099 == 214099, CellWalk 98120 == 98120), stable across runs. Switched to `EXPECT_NEAR` at 2% (observed drift < 0.05%). Geometric correctness stays tightly guarded by `ManySeedsMultithreadedStreamlineLength`.

## 4. Regression and import tests had no ctest environment

The unit tests got an `ENVIRONMENT` property with the Qt/Python/Qwt DLL paths, but the regression and import tests were added with a bare `ADD_TEST`. Launched from a plain shell they died instantly with `0xc0000135` (STATUS_DLL_NOT_FOUND) — so **no local regression test could run at all**, even though they pass on the build farm (whose job environment already has those directories on PATH).

Hoists the environment so all three groups share it, and adds the Qwt install dirs.

The Qt platform is deliberately split: unit tests keep `QT_QPA_PLATFORM=offscreen` (headless, must not pop a window), while the regression/import tests that launch the full app leave it unset. On a developer workstation ViewScene networks need the real `windows` platform for a usable GL context — under `offscreen` they exit 3 — and headless CI can still force it via the job environment.

## 5. Python unit test had stopped running entirely

Two separate problems:

**Discovery.** `list_tests.py` built each ctest name by splitting `str(test)`, assuming the pre-3.11 format `method (module.Class)`. Python 3.11 changed it to `method (module.Class.method)`, so the method name was appended twice:

```
Unit.testSCIRunAddModule.TestSCIRunAddModule.test_scirun_module_ids.test_scirun_module_ids
```

ctest reported a failure, but the test had in fact stopped executing on any modern Python. Now uses `TestCase.id()`, which is `module.Class.method` on every version.

**Environment.** These tests spawn `SCIRun.exe`, which then died with `0xc0000135`. They now get the DLL paths — deliberately *without* `PYTHONHOME`, since they are themselves run by a python interpreter that would otherwise look for its stdlib under the Superbuild prefix.

Also fixes a latent bug introduced earlier in this branch: `SCIRUN_TEST_ENVIRONMENT` was built by expanding another list variable unquoted, which strips a level of escaping from the semicolons in PATH. ctest then split PATH into separate entries and every unit test ran with a PATH truncated at the first separator. It stayed invisible until a rebuild gave `Interface_Modules_Render_Tests` and `Modules_Dialog_Factory_Tests` Qt/Qwt dependencies they could no longer resolve.

## 6. Data race on per-thread success flags — fixes #645

Parallel algorithms recorded per-worker results in a `std::vector<bool>`, written as `success_[proc] = ...` from each worker. That reads as obviously safe — every thread touches only its own index — but `std::vector<bool>` is the bit-packed specialization: elements share underlying words, so those writes are **not** writes to distinct memory locations and race with each other. A worker's result could be lost.

In `ParallelLinearAlgebra` a corrupted flag makes `start_parallel()` return false, surfacing as an intermittent `"Encountered an error while running parallel linear algebra"`. `Algorithms_Math_Tests` failed about one run in three on the Jacobi cases.

Adds `Core::Thread::ParallelSuccessFlags` next to `Parallel::RunTasks` — this is a property of the RunTasks pattern, not of any one algorithm — with the reasoning documented in that single place. `vector<char>` elements are distinct memory locations, so the writes are race-free and published to the reader by the thread join.

The pattern had been copied into **eight** files, all with the same race: `ParallelLinearAlgebra.h`, `GenerateStreamLines.cc`, `BuildFEMatrix.cc`, `BuildFEVolRHS.cc`, `MapFieldDataOntoElems.cc`, `MapFieldDataOntoNodes.cc`, `BiotSavartSolverAlgorithm.cc`, `ArrayMathInterpreter.h`.

Verified **12/12** consecutive passes, against the prior ~1-in-3 failure rate.

This is #645, reported in 2014. `fa0996e5d` fixed a *different* race in the same file (`makeCompressed` on the shared matrix) and the flag race survived; see the issue for the full analysis.

---

## Test plan

- [x] Unit suite: 54/56 (remaining two are #2551)
- [x] Regression suite: 308/309 (remaining one is #2567)
- [x] `Algorithms_Math_Tests` 12/12 consecutive after the race fix
- [x] All 15 legacy importer tests confirmed running, not skipping
- [x] Verified ViewScene regression networks render under the real Qt platform
- [ ] CI on this PR


dcwhite:
@jessdtate This one will clean up the overnight test numbers in Slack.

jessdtate (approved):

dcwhite:
@basisunus wait for this one to merge, then merge master to your PR branch.

PR #2574, opened 2026-07-23
CI: fix nightly/push concurrency collision; add Slack nightly summary; drop dead Travis config

dcwhite:
Follow-up to #2573. Three related CI cleanups.

## 1. Nightly and master-push builds no longer cancel each other
The five full-matrix workflows (`ccpp`, `mac`, `windows`, `windows-build-script`, `regression-tests`) keyed concurrency on `${{ github.workflow }}-${{ github.ref }}`. A scheduled run and a push to master share `github.ref` (`refs/heads/master`), so a nightly firing while a master-push build was still running cancelled it — or vice versa.

Observed on 2026-07-23: the mac nightly cancelled 11 of 14 jobs in the master-push run of the same commit. Adding `github.event_name` to the group puts `push`, `schedule`, and `workflow_dispatch` in separate groups.

## 2. Slack nightly summary
New `nightly-slack.yml`: a `workflow_run`-triggered notifier that posts a per-workflow summary of each **scheduled** run to Slack. One file instead of five copy-pasted notify jobs, and `reusable-build.yml` (the tested build path) is untouched.

It downloads the run's `test-results-*` artifacts and reports the **real ctest pass/fail counts**, so the regression job's `continue-on-error` "success" can't hide failing tests — a green job with failing tests is flagged yellow.

- Schedule-only (master pushes still get GitHub's default failure email).
- Requires a `SLACK_WEBHOOK_URL` repo secret; skips the post with a warning if unset.
- `workflow_run` only fires from the default branch, so this goes live on the first nightly **after** merge to master.

## 3. Remove dead `.travis.yml`
Travis no longer builds this repo. Dropped the stale config and its dead `.dockerignore` entry.

> Note: the Slack integration token that was in the old `notifications:` block remains in git history and must be revoked Slack-side (done separately — removing the `#scirun5build` Travis CI configuration).


dcwhite:
@jessdtate Next up

jessdtate (approved):

PR #2513, opened 2026-06-13
Add keyboard shortcuts to ViewScene (issue #694)

dcwhite:
## Summary

Implements a keyboard shortcut system for the ViewScene 3D viewer, addressing issue #694.

### Shortcuts added

| Key | Action |
|-----|--------|
| `0` | Autoview |
| `Ctrl+0` | Autoview (no scale) |
| `1`–`6` | Axis-aligned views (±X, ±Y, ±Z) |
| `X` | Snap to nearest axis |
| `H` | Go to home view |
| `Alt+H` | Set home view |
| `C` | Toggle active clipping plane visibility |
| `D` | Toggle fog |
| `K` | Toggle all lights on/off |
| `L` | Lock/unlock all view axes |
| `O` | Toggle corner orientation triad |
| `I` | Open shortcuts help dialog |

### Help dialog
- `I` key (or toolbar `?` button) opens a reference dialog listing all shortcuts
- Dialog positions at top-right of ViewScene and remembers position within a session
- Double-clicking a row in the dialog invokes that action
- Unimplemented shortcuts are hidden; wired shortcuts flash a tooltip on activation
- Key presses are forwarded to ViewScene even when the help dialog has focus

### Refactor
- Shortcuts defined via `ShortcutDef` enum class with string annotations and action mechanics, replacing a hand-coded `.ui` table

### Bug fixes
- Orientation glyph aspect ratio wrong at initialization in `CoreBootstrap` (stale 640×480 default size)
- Orientation glyph squished on first show

### OffscreenGLRenderer
- Prototype files (`OffscreenGLRenderer.cc/.h`) added but not wired into `ViewScene` — kept for future regression-image work

## Test plan
- [ ] Open ViewScene and verify each shortcut key triggers the correct action
- [ ] Press `I` — confirm help dialog opens at top-right, closes cleanly
- [ ] Double-click a shortcut row — confirm the action fires
- [ ] Verify `Alt+H` / `H` round-trip (set home, move camera, go home)
- [ ] Verify orientation glyph has correct aspect ratio on first show
- [ ] Confirm no regressions in existing ViewScene mouse/toolbar controls

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2575, opened 2026-07-25
Pin CMAKE_OSX_DEPLOYMENT_TARGET so the macOS floor is deliberate

### LINKED ISSUE #2565: Pin CMAKE_OSX_DEPLOYMENT_TARGET so one build runs on older macOS
Nothing in the build sets `CMAKE_OSX_DEPLOYMENT_TARGET` — the only matches in the tree are vendored libxml2's autoconf and an unrelated HTML file. When it is unset, CMake defaults the deployment target to the version of macOS the build happened to run on.

Two consequences:

**1. The macOS floor of a release is an accident of the build machine.** Whichever runner or laptop cut the build silently becomes the minimum supported version. Nothing declares that floor and nothing checks it.

**2. Symbols newer than the floor link cleanly and fail at load time on users' machines.** With a pinned deployment target the compiler can diagnose these (`-Wunguarded-availability`) and the linker can weak-link them. Unpinned, it has no version information to act on, so there is no build-time signal at all — the first sign is a user on an older Mac getting `dyld: Symbol not found`.

### Concrete instance: #2564

`quick_exit` entered libSystem in macOS 15.0, and the 15 SDK declares it with no availability annotation — gated only on `__DARWIN_C_LEVEL`/C11, a compile-time feature gate. On a **macOS 14.5 host with the 15.2 SDK**:

```console
$ cat t.cc
#include <cstdlib>
void boom() { std::quick_exit(1); }

$ clang++ -std=c++17 -c t.cc -o t.o && nm -u t.o
_quick_exit
```

Compiles clean, no warning, hard reference — and that binary will not load on macOS < 15. Worth noting the trigger is the **SDK, not the host**: a macOS 14 machine with Xcode 16 installed builds the broken binary, so "it works on my 14 box" proves nothing.

#2564 fixes that one symbol by always taking `_Exit` on Apple. That is a per-symbol workaround. The next post-floor API that drifts into the codebase reintroduces the same failure with the same absence of any build-time signal. Gating on `__MAC_OS_X_VERSION_MIN_REQUIRED` does not help either — it reads the deployment target, which follows the build host when unpinned, so it reads `150000` on exactly the builders that emit the bad reference.

### Proposal

Set `CMAKE_OSX_DEPLOYMENT_TARGET` explicitly in the Superbuild and propagate it to every external project, so SCIRun and all of its dependencies build against one declared floor.

Blast radius is wide — it touches every Superbuild dependency — so this is its own PR rather than a rider on #2564.

### Open question

What should the floor be? We have users on older Macs, so this wants a deliberate decision rather than whatever the newest runner happens to be. Whatever we pick becomes an enforced, checkable contract instead of a property of whoever ran the build.

Related: #2564, #2537

ZayanKhan-12:
Fixes #2565.

Nothing in the build set `CMAKE_OSX_DEPLOYMENT_TARGET`, so the macOS floor of every binary was an accident of the build machine's OS/SDK, and post-floor symbols linked cleanly but failed at load time on older Macs — #2564 (`quick_exit`) was one instance.

## What this does

- Defaults the floor to **14.0** (the oldest macOS in the CI matrix), as a user-overridable cache variable set before the first `PROJECT()` call in both `Superbuild/CMakeLists.txt` and `src/CMakeLists.txt`. Guarded on `CMAKE_HOST_APPLE` (the target `APPLE` var doesn't exist before `project()`); ignored off-Apple.
- Forwards it explicitly to every compiled ExternalProject, since `ExternalProject` sub-builds do not inherit it:
  - CMake-configured externals (Zlib, SQLite, Teem, Freetype, Cleaver2, LodePng, Tny, Qwt, Tetgen, Ospray) get a cache/args entry mirroring the existing `CMAKE_BUILD_TYPE` forwarding.
  - **Boost**: the real compile happens in the `build_libs` b2 step, so b2 additionally gets `cxxflags=/linkflags=-mmacosx-version-min=…`.
  - **CPython**: `MACOSX_DEPLOYMENT_TARGET` is set in the configure env; configure bakes it into its Makefile so the build steps inherit it.
  - Header-only externals (Eigen, GLM, SpdLog) and WIN32-only Glew unchanged. `GoogleTestExternal.cmake` is not referenced by the Superbuild, so it was left alone.
- Adds `-Wunguarded-availability-new` to the APPLE warning set in `SCIRUN_ADD_LIBRARY`, so availability-annotated post-floor API use is a compile-time warning instead of a runtime dyld failure. (Honest caveat: it would *not* have caught `quick_exit`, which the 15 SDK declares without availability annotations — but with the pin, `__MAC_OS_X_VERSION_MIN_REQUIRED` now reads 140000 deterministically, which is what makes #2564's guard reliable.)
- Adds a macOS CI step asserting the built binary's `LC_BUILD_VERSION` `minos` equals the pin, so a dropped flag shows up as a red build rather than a user crash report.

## Verification

- Superbuild configure run locally (Apple Silicon, CMake 4.3): `CMAKE_OSX_DEPLOYMENT_TARGET:STRING=14.0` lands in the top-level cache and in every generated `Externals/tmp/<name>_external-cache-Release.cmake` plus Tetgen's configure command.
- The new CI otool step verifies the end-to-end result on the macOS PR job.

Note: touches the same `ExternalProject_Add` blocks as the superbuild-caching residue on #2513 — happy to rebase whichever lands second.


dcwhite:
I'll have to examine this one carefully (not just with Claude 😉 )

PR #2576, opened 2026-07-25
CI: cache Superbuild externals between runs

### LINKED ISSUE #2520: CI: cache Superbuild externals build output (actions/cache)
> Tracked under #1487 ("External dependency caching"); this is the detailed plan for the caching item in #2028.

## Summary

SCIRun's CI rebuilds the Superbuild externals (the heavy third-party C++ dependencies) from scratch on every run — none of the nine workflows (`ccpp.yml`, `mac.yml`, `windows.yml`, `reusable-build.yml`, etc.) use any caching. We can cut a large fraction of CI time by caching the externals build output, keyed on the files that define how those externals are built.

This mirrors how our sibling project [SCIInstitute/shapeworks](https://github.com/SCIInstitute/shapeworks) handles it: its biggest CI win is a GitHub-native `actions/cache` of the compiled dependencies (ITK, VTK, OpenVDB, …), keyed on the dependency *recipe* rather than the project source, so the multi-hour external build is reused across every PR and commit.

## The pattern (from ShapeWorks)

Split restore / save, with the build and the save both gated on a cache miss:

```yaml
- name: Restore Externals Cache
  id: cache-externals
  uses: actions/cache/restore@v4
  with:
    path: <superbuild externals install/build dir>
    key: ${{ runner.os }}-externals-${{ hashFiles(
      'Superbuild/CMakeLists.txt',
      'Superbuild/**/*.cmake') }}

- name: Build Externals
  if: steps.cache-externals.outputs.cache-hit != 'true'
  run: <superbuild configure+build>

- name: Save Externals Cache
  if: steps.cache-externals.outputs.cache-hit != 'true'
  uses: actions/cache/save@v4
  with:
    path: <same path>
    key: <same key>
```

Why it works:

- **Key hashes the externals recipe, not the source.** Keyed on the Superbuild CMake files (+ toolchain/compiler), *not* on SCIRun's own `.cpp`/`.h`. The cache only invalidates when the externals definition changes, so it survives across normal PRs.
- **Split `restore` + `save`** (not the combined action) so we can skip both the build and the re-save on a hit.
- **Per-variant keys** to avoid collisions: include `runner.os`, build type (Debug/Release), and on macOS the deployment target / arch.

## Scope (this issue = "Step 1" only)

- [ ] Identify the exact Superbuild externals output path(s) to cache, per platform
- [ ] Decide what goes into the cache key (Superbuild CMake files + toolchain + build type + arch)
- [ ] Wire `actions/cache/restore` + gated build + `actions/cache/save` into `reusable-build.yml` (and any workflow that builds externals directly)
- [ ] Verify a cold run populates the cache and a warm run restores it and skips the externals build
- [ ] Confirm cache size stays within GitHub's per-repo limits

Out of scope (separate follow-up): adding `ccache`/`sccache` as a compiler launcher for SCIRun's own sources.

## Context

Faster CI directly helps unblock the in-progress C++ modernization work, which is currently gated on new CI testing jobs.

Reference: [shapeworks/.github/workflows](https://github.com/SCIInstitute/shapeworks/tree/master/.github/workflows) — `build-linux.yml`, `gha_deps.sh`, `common.sh`.

ZayanKhan-12:
Fixes #2520.

The externals (Boost, CPython, Teem, …) change rarely but dominate the wall time of every CI build. This caches the `Externals` tree keyed on the exact build flavor plus `hashFiles('Superbuild/**')`, so a run whose externals are unchanged only compiles SCIRun itself.

## Why it needs more than an actions/cache step

`ExternalProject` re-runs any step whose stamp is older than its inputs, and a cache restore produces exactly that — the runner's CMake modules and freshly written `Externals/tmp` scripts are newer than every restored stamp, so make would redo all external steps, including *deleting and re-cloning* `Source/`. Two supporting changes make a restored tree stick:

- `build.sh` future-dates the restored external stamps between configure and make, gated on `SCIRUN_TOUCH_EXTERNAL_STAMPS`, which CI sets only on a cache hit. `SCIRun_external`'s stamps are excluded (and not cached), so SCIRun itself always reconfigures and rebuilds. Local builds are untouched.
- The Superbuild sets `EP_UPDATE_DISCONNECTED TRUE` so reconfigure doesn't run `git fetch` inside restored `Source/` trees, which are cached **without** `.git` to keep entries small (~1–2 GB compressed). Developers keep the `<name>_external-update` targets for deliberate updates.

## Design notes

- **Exact-key restore only, no `restore-keys`**: with future-dated stamps, a prefix-matched stale cache would be silently used with the wrong externals. A key change means a clean cold build.
- Key includes runner, variant, python/qt/ospray/coverage flavor, and the Superbuild hash. The `superbuild-v1` prefix is the manual bust knob — e.g. if a branch-tracking external (LodePng `origin/master`, Tny) moves upstream without any `Superbuild/` edit, bump it to `v2`.
- Save runs only after a successful build on a cache miss.
- **Windows is scoped out** for now (different `build/` dir, per-config stamps under the VS generator) — can follow up if this proves out.

## Verification

- Locally: superbuild configure passes with `EP_UPDATE_DISCONNECTED`; the stamp-touch was exercised against a real configured `Externals/Stamp` tree — all external stamps future-date to 2100, `SCIRun_external`'s stay untouched.
- On this PR: the first CI run is cold and should show the cache save; a follow-up push will demonstrate the warm path (restore + externals skipped). I'll post both timings here.

Note: overlaps the same `reusable-build.yml` region and Superbuild files as #2513's caching groundwork (`.github/superbuild-caching.md`) — this implements the actions/cache half described there; happy to rebase on #2513 if that's landing first.

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -74,6 +74,27 @@ jobs:
       - name: Print Qt dir
         run: echo "QT_ROOT_DIR=${QT_ROOT_DIR:-(unset)}"
 
+      # Superbuild externals change rarely but cost the bulk of every build.
+      # Cache the whole Externals tree: the inner SCIRun build consumes headers
+      # and libraries from Source/ (Eigen, GLM, SpdLog, Boost stage), Build/
+      # (Zlib, Teem, ...), and Install/ (Python, Tetgen, Qwt) alike. Exact-key
+      # match only -- a prefix-matched stale tree would be silently used with
+      # the wrong externals because the restored stamps defeat ExternalProject's
+      # own staleness detection (see touch_external_stamps in build.sh).
+      # Windows is scoped out for now: different build dir, per-config stamp
+      # layout under the VS generator.
+      - name: Restore Superbuild externals cache
+        id: cache-externals
+        if: ${{ runner.os != 'Windows' }}
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            bin/Externals
+            !bin/Externals/Download
+            !bin/Externals/Source/**/.git
+            !bin/Externals/Stamp/SCIRun_external
+          key: superbuild-v1-${{ inputs.runner }}-${{ inputs.variant }}-py${{ inputs.build-with-python }}-qt${{ inputs.qt-version }}-osp${{ inputs.with-ospray }}-cov${{ inputs.coverage }}-${{ hashFiles('Superbuild/**') }}
+
       - name: Checkout SCIRunTestData
         if: ${{ inputs.run-regression-tests || inputs.coverage }}
         uses: actions/checkout@v6
@@ -90,6 +111,8 @@ jobs:
       - name: Build (Linux/macOS)
         if: ${{ runner.os != 'Windows' }}
         shell: bash
+        env:
+          SCIRUN_TOUCH_EXTERNAL_STAMPS: ${{ steps.cache-externals.outputs.cache-hit == 'true' && '1' || '0' }}
         run: |
           set -eux
           CMD=(./build.sh)
@@ -206,6 +229,19 @@ jobs:
           & cmake --build . --config Release --parallel
 
           Pop-Location
+
+      # Save only after a successful build on a cache miss; a hit means this
+      # exact key is already stored (immutable), so saving again would fail.
+      - name: Save Superbuild externals cache
+        if: ${{ runner.os != 'Windows' && steps.cache-externals.outputs.cache-hit != 'true' }}
+        uses: actions/cache/save@v4
+        with:
+          path: |
+            bin/Externals
+            !bin/Externals/Download
+            !bin/Externals/Source/**/.git
+            !bin/Externals/Stamp/SCIRun_external
+          key: ${{ steps.cache-externals.outputs.cache-primary-key }}
 
       - name: Smoke test - executable loads and runs (Unix)
         if: ${{ inputs.run-smoke-test && runner.os != 'Windows' }}
--- a/Superbuild/Superbuild.cmake
+++ b/Superbuild/Superbuild.cmake
@@ -70,6 +70,13 @@ ENDIF()
 
 INCLUDE( ExternalProject )
 
+# Skip the per-external git fetch on reconfigure. Every external pins a tag or
+# tracking branch that changes rarely; the fetch mostly costs network time and,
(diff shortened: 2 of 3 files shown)


dcwhite:
I tried this one already and got a lot of failures. This will need to run the full CI set to run, not just the PR subset, and will need to be tested on multiple runs. Maybe we can merge it to a side branch for more extensive testing first.

dcwhite:
And yes, this is blocked by #2513, that needs to merge first.

dcwhite:
@ZayanKhan-12 Claude merged this to a staging branch and tested it out, and there are some issues. I'll have it post a full summary for "digestion"

dcwhite:
Validated this on a staging branch (`superbuild-cache-staging`) running the full CI matrix cold, then a warm rerun. The save/restore machinery works on a single build, but the cache is fundamentally too large for the matrix and delivers ~zero warm hits.

**Blocker 1 — the cache footprint exceeds GitHub's 10 GB/repo limit by several times, within a single workflow.** `mac-build` alone has 13 cache-saving jobs at ~2.4 GB each ≈ ~31 GB of caches per run, against a hard 10 GB ceiling. Jobs LRU-evict each other mid-run; after the full matrix, only 3 mac caches (~7 GB) survived and every linux/regression cache was evicted. A warm rerun then confirmed the consequence directly — every completed job reported `Cache not found for input keys: …`. No steady-state speedup exists; the cache thrashes itself every run.

**Blocker 2 — cross-workflow key collision.** `linux-headless` in `ccpp.yml` and the matching job in `regression-tests.yml` compute an identical key (no workflow component in it) and raced:
`Failed to save: Unable to reserve cache … another job may be creating this cache.`

**To make this viable:**
- Drastically shrink per-job footprint — cache only `Externals/Install` (the artifacts the inner build actually consumes), not `Source` + `Build` + stamps — and confirm the *whole matrix* fits under 10 GB, not just one job.
- Or scope caching to a small subset of jobs rather than every matrix entry.
- Add the workflow name to the cache key regardless, to eliminate the collision.

The single-build mechanism (stamp future-dating, `EP_UPDATE_DISCONNECTED`) is sound — the problem is scale, and it needs a footprint redesign before the warm path can pay off. Happy to share the staging run logs.

dcwhite (changes requested):
See claude comment

PR #2577, opened 2026-07-25
Add delimiter option to CreateMatrix text entry

### LINKED ISSUE #706: Specify delimiter for parsing entry in CreateMatrix
Follow-up from #122, requested by Josh.

ZayanKhan-12:
Fixes #706.

Pasting comma- or semicolon-separated data (the common CSV case) into CreateMatrix silently mis-parsed: the text goes straight to the whitespace-splitting stream parser, so `1,2` read as the single value `1` and the rest of the line was dropped.

## What this does

- Adds a `MatrixTextDelimiter` state option — **Space** (default), **Comma**, **Semicolon**, **Tab** — with a combo box in the dialog (wired via `addComboBoxManager`, following e.g. `BooleanCompareDialog`).
- Non-space delimiters are normalized to spaces before parsing, so `1, 2` with both a comma and a space still parses, and the parser in `Core/Datatypes/MatrixIO.h` is untouched.
- Back-compat: networks saved before this option carry no delimiter state and keep the historical space-delimited behavior; a unit test pins that down.

One correction for the issue thread: the 2026-06 triage comment says the current parser hardcodes comma — it actually hardcodes whitespace (stream extraction in `MatrixIO.h`). This change makes comma input *work* rather than altering an existing comma path.

## Tests

New `CreateMatrixTests.cc` (registered in `Modules_Math_Tests`): space/comma/semicolon/tab parsing, comma+space input, unknown-delimiter fallback to legacy behavior, and empty entry. Runs in the `linux-unit-tests` PR job via `ctest -R Modules_Math_Tests`.


dcwhite:
Thanks for including the unit test. The unit test will run as part of the revamped CI system but won't report a failure yet (that's part of another PR).

 I'll get this reviewed within the next few days. Then, you can do the merge.

Issue #2518, opened 2026-06-18
Migrate branch protection to GitHub rulesets


dcwhite:
## Summary

SCIRun currently protects `master` with a **classic branch protection rule** and has **no rulesets** defined. GitHub rulesets are the modern replacement for classic protection and would let us reproduce everything we do today, gain visibility/auditability for contributors, and add protections that classic rules can't (tag protection, push-time file restrictions, stackable/org-wide policies).

This issue proposes migrating to rulesets and adding a couple of new protections.

## Current state (`master`)

Confirmed via the GitHub API:

- Required status checks (strict / must be up to date): `linux-gui / build`, `mac-gui / build`, `windows-gui-qt6 / build`
- Require PR before merge: 1 approval, dismiss stale reviews on push, no CODEOWNERS requirement
- Required conversation resolution: on
- Block force pushes: on · Block deletions: on
- **Include administrators: on** (admins are not exempt)
- Linear history: off · Required signatures: off
- Rulesets: none

## Why rulesets

| | Classic protection (today) | Rulesets |
|---|---|---|
| How many apply to a branch | One rule per pattern | Many stack/aggregate, most-restrictive wins |
| Visibility | Admin-only | Anyone with read can see the constraints |
| Disable temporarily | Delete it | Flip to Disabled (config retained) |
| Dry-run a change | No | Evaluate mode (Enterprise) |
| Tags | Limited | First-class tag targeting |
| Push-time controls (file size/path/type) | No | Push rulesets (cover the fork network) |
| Bypass granularity | "include admins" toggle | Per actor (roles/teams/apps), always vs PRs-only |

### 1. Migrate `master` to a branch ruleset (1:1 with today)

Empty bypass list to preserve "admins included".

```jsonc
// gh api repos/SCIInstitute/SCIRun/rulesets --method POST --input ruleset.json
{
  "name": "master protection",
  "target": "branch",
  "enforcement": "active",
  "conditions": { "ref_name": { "include": ["refs/heads/master"], "exclude": [] } },
  "bypass_actors": [],
  "rules": [
    { "type": "deletion" },
    { "type": "non_fast_forward" },
    { "type": "pull_request", "parameters": {
        "required_approving_review_count": 1,
        "dismiss_stale_reviews_on_push": true,
        "require_code_owner_review": false,
        "require_last_push_approval": false,
        "required_review_thread_resolution": true
    }},
    { "type": "required_status_checks", "parameters": {
[...]

PR #2586, opened 2026-07-27
ci: add CODEOWNERS to auto-request maintainer review on every PR

dcwhite:
Adds `.github/CODEOWNERS` with a catch-all rule so GitHub auto-requests review
from the core maintainers on **every** pull request:

```
* @dcwhite @jessdtate
```

### Behavior
- Any PR touching any path auto-requests review from both maintainers.
- GitHub never requests review from the PR **author**, even when they're an
  owner — so a PR opened by @dcwhite auto-requests @jessdtate, and vice versa.
- Both accounts need write access to the repo for the request to stick (both are
  maintainers, so this is fine).

### To make it *required*, not just requested
CODEOWNERS only **requests** review. To block merge until a code owner approves,
enable **Settings → Branches → master → Require review from Code Owners** (branch
protection). That toggle is repo-admin UI and isn't something a PR can set.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,12 +1,27 @@
 # Code owners for SCIRun.
 #
-# GitHub automatically requests review from the owners listed here whenever a
-# pull request touches a matching path. The catch-all rule below therefore
-# requests the core maintainers on every PR.
+# GitHub auto-requests review from the owners listed here whenever a PR touches
+# a matching path. It never requests the PR author, so when a maintainer opens a
+# PR only the *other* listed owner is pulled in.
 #
-# Note: GitHub never requests review from the PR author, even if they are an
-# owner. So when one maintainer opens a PR, only the other is auto-requested.
+# This file is deliberately SCOPED rather than a blanket "* @owner": we only gate
+# the paths where a bad merge is expensive or hard to reverse. Everything not
+# matched below (docs, README/badges, general CI tweaks, most module code) has no
+# required owner and can be self-merged once branch checks pass. Add paths here as
+# the reviewer pool grows.
 #
 # See: https://docs.github.com/articles/about-code-owners
 
-* @dcwhite @jessdtate
+# --- Build & dependency configuration (high blast radius) ---
+/Superbuild/                             @jessdtate @dcwhite
+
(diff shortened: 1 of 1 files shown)


dcwhite:
@jessdtate Merge this one first (adds CODEOWNERS + the review gate), then the rest of today's batch in this order:

1. #2586 — CODEOWNERS (this PR)
2. #2585 — README badges (docs-only, trivial)
3. #2584 — arm64 Boost fix (the only bug fix)
4. #2587 — VERSIONS.cmake dep pins (foundational; land before the CI PRs so they can reference the centralized pins)
5. #2589 & #2588 — both edit `windows.yml` and will conflict at the `jobs:` tail; merge either first, then rebase the other keeping both job blocks
6. #2576 — Superbuild cache (community PR, needs a full CI run first)

The only real mechanical conflict in the batch is #2588 ↔ #2589 — a 30-second "keep both" resolution. Everything else branches independently off master, and CI is green across the board (only #2587 still has a couple jobs running).

dcwhite:
Update: reworked this from the blanket `* @owner` catch-all into a **scoped** CODEOWNERS. Now only high-risk paths auto-request review:

- `/Superbuild/**` (dependency/build config)
- `/.github/workflows/publish-release.yml` (release + installer publishing)
- `/src/Dataflow/Serialization/`, `/src/Dataflow/Network/` (engine + `.srn5` format)
- `/src/Core/Datatypes/` (core datatypes)
- `/.github/CODEOWNERS` (the policy itself)

Everything else — docs, README badges, general CI-matrix tweaks, most module code — is intentionally left ungated, so it only pings a maintainer when it actually matters.

**Correction on the enforcement side** (my earlier note suggested `required_approving_review_count: 0` — scratch that). Lowering the count to 0 would let *any* author self-merge ungated paths without review, which we do not want. The right approach keeps the count at **1** and adds a **per-actor bypass** so specific maintainers can self-merge while everyone else still needs an approval. That is a GitHub **ruleset** feature (classic branch protection can only do all-or-nothing "include admins"), which ties into #2518 (migrate branch protection to rulesets). Proposing we settle the policy there / at Friday's meeting before this merges. Holding the PR until then.

jessdtate (approved):

Issue #2480, opened 2026-05-30
Improve dependency update flow — Modernize external dependency management and CI


dcwhite:
Improve dependency update flow — Modernize external dependency management and CI

Goal:
Simplify and modernize how SCIRun tracks, updates, and documents external dependencies (VTK, Python libs, third-party tools).

Acceptance criteria:
- Documented process for updating dependencies.
- Automated dependency checks or periodic CI job.
- Pinning policies for reproducible builds.

Suggested steps:
1. Add a dependency manifest and update instructions.
2. Add CI job that checks for outdated pinned versions and runs smoke builds.

dcwhite:
**Analysis (2026-06-13)**

Improve dependency update flow. A dependency manifest (pinned versions + sources) doesn't exist — version info is scattered across Superbuild/*.cmake files. A central VERSIONS.cmake or a lockfile approach would simplify updates. Adding a CI job that checks if pinned versions have newer releases would enable proactive updates.

dcwhite:
**Status (2026-08-08): implementation merged, issue deliberately held open**

The manifest work is on `master`: `Superbuild/VERSIONS.cmake`, `Superbuild/scripts/check_dependency_versions.py`, `.github/workflows/dependency-check.yml`, `docs/dev_doc/DependencyUpdates.md` — landed in 5937825be plus follow-ups through bd36ee3d4. The branch has no unmerged commits.

Holding this open until a real dependency upgrade has been driven through the manifest, rather than closing on the file existing. Boost (#2548) is the natural first exercise. The one acceptance-criteria bullet not implemented is the "runs smoke builds" half of the CI job — that is intentionally not being chased here; a weekly job that rebuilds drifted externals from scratch is not affordable until the externals install/caching program (#2627) lands.

## Things to watch

1. **First scheduled sweep is Mon 2026-08-10 06:00 UTC.** Every run in the workflow's history so far is `pull_request` on the feature branch. The scheduled-only path — `steps.check.outputs.actionable == 'true'` opening/updating a tracking issue — has never actually executed. If it misbehaves it will do so by filing a junk issue, which is loud but harmless.
2. **The branch-head assertion added in bd36ee3d4** fails the job if any `*_GIT_TAG` is edited back to a mutable branch name. This is the guardrail most likely to fire during a Boost or Python bump — that is working as intended, not a bug.
[...]

PR #2587, opened 2026-07-27
Centralize dependency pins in VERSIONS.cmake + add freshness CI (#2480)

dcwhite:
Addresses #2480 (Improve dependency update flow).

## What this does

Previously, dependency versions were scattered across ~20 `Superbuild/*External.cmake` files, and ~12 of them pinned to a floating branch ref (`origin/master`, `origin/scirun-5.0.0-beta`, …), so clean builds days apart could silently differ — the "reproducible builds" acceptance criterion was not met.

This PR:

1. **Adds a central manifest** — [`Superbuild/VERSIONS.cmake`](Superbuild/VERSIONS.cmake), the single source of truth for every built dependency's source URL + pinned version, via a `sci_dep_version()` macro so each pin is a CLI-overridable cache var (`-DSPDLOG_GIT_TAG=…`). Included early in `Superbuild.cmake`.
2. **Pins for reproducibility** — resolves all 12 floating `origin/<branch>` refs to immutable **commit SHAs** (captured 2026-07-27), keeping the tracked branch name in each doc string. Release tags kept where they existed (Eigen, GLM, spdlog, Boost, TetGen, Qwt).
3. **Wires 16 `*External.cmake` files** to consume the manifest.
4. **Adds a report-only CI job** — [`dependency-check.yml`](.github/workflows/dependency-check.yml) + [`check_dependency_versions.py`](Superbuild/scripts/check_dependency_versions.py). Weekly + on-demand; flags **DRIFT** (tracked branch advanced past our SHA) and **UPDATE** (newer upstream release), writes a run summary, and opens/refreshes one tracking issue.
5. **Docs** — new [`DependencyUpdates.md`](docs/dev_doc/DependencyUpdates.md) (process + pinning policy); `dependencies.md` updated to point at the manifest.
6. **Moves 4 dead files → `Superbuild/deprecated/`** (git-renamed, recoverable): `LibPNGExternal`, `GoogleTestExternal`, `DataExternal`, `CtkExternal` — none are `ADD_EXTERNAL`'d. GoogleTest is the git submodule; CTK is an in-tree code copy at `src/Interface/Modules/Base/CustomWidgets/CTK/`.

## Not manifest-managed (documented)
GoogleTest (submodule), CTK (in-tree copy), Qt (`find_package`), and test-data fixtures.

## Testing
- `check_dependency_versions.py` self-tested (`--offline-upstream`): parses the manifest and confirms all pins match their tracked branch tips.
- ⚠️ **Not yet run through a full CMake Superbuild configure** — the variable wiring should be confirmed with a clean configure before merge.


dcwhite:
Came across this while investigating Boost's clone cost for #2617 (details in #2630). **Two of the pins are still floating**, and the drift checker won't catch either.

I verified every ref in `VERSIONS.cmake` with `git ls-remote`. The bulk of the PR holds up — all nine `scirun-pin-2026.07.27` refs are real tags, and GLM, spdlog, TetGen and the Qwt wrapper are genuine upstream tags:

| dep | ref | actual |
|---|---|---|
| zlib, lodepng, freetype, teem, sqlite, glew, cleaver2, tny, ospray | `scirun-pin-2026.07.27` | TAG ✅ |
| GLM | `0.9.9.8` | TAG ✅ |
| spdlog | `v1.10.0` | TAG ✅ |
| TetGen | `v1.6.1` | TAG ✅ |
| Qwt wrapper | `v0.1.1` | TAG ✅ |
| **Boost** | `v1.90.0` | **BRANCH** ❌ |
| **Python** | `origin/3.13.1` | **BRANCH** ❌ |

## Boost

```cmake
sci_dep_version(BOOST_GIT_TAG "v1.90.0" "Boost pinned tag")
```

```
$ git ls-remote https://github.com/CIBC-Internal/boost.git | grep v1.90.0
f596b5d3f7dc7195f20591aca3b5d1e33d534c0c	refs/heads/v1.90.0
```

`refs/heads/`, not `refs/tags/`. The only actual tag in that repo is `python1`. The PR description lists Boost under "Release tags kept where they existed", but there is no release tag here — it's a branch head that can move, and the repo was last pushed 2026-02-13.

Suggested: `"f596b5d3f7dc7195f20591aca3b5d1e33d534c0c"` with doc string `"Boost pin -> commit f596b5d3 (branch v1.90.0)"`, matching the convention used for the other nine.

## Python

`PythonExternal.cmake:59` builds the ref as `origin/${USER_PYTHON_VERSION}`:

```
$ git ls-remote https://github.com/CIBC-Internal/python.git | grep 3.13.1
b54e57f244c928d7f2b90106a74603462d1662a2	refs/heads/3.13.1
```

Also a branch. The doc string does say `origin/<version>`, so this one is at least visible — but it is one of the floating refs the PR set out to eliminate, and it survived.

## The drift checker won't flag either

`check_dependency_versions.py` derives the tracked branch from the doc string:

```python
BRANCH_DOC_RE = re.compile(r"\(branch\s+(?P<branch>[^)]+)\)")
```

Neither `"Boost pinned tag"` nor the Python doc string contains a `(branch …)` marker, so `_check_drift()` never runs for them. Both fall through to the release-tag path and get compared against `boostorg/boost` and `python/cpython` — upstream repos that are **not** the repos actually being built. `CIBC-Internal/boost` isn't even a GitHub fork (`fork: false`, no parent) and its `master` ends in 2016 carrying SCIRun patches, so an upstream-release comparison there is not meaningful.

Net effect: the two deps that are still mutable are also the two the checker is blind to.

## Suggested fix

- Resolve `BOOST_GIT_TAG` to the SHA above and give it a `(branch v1.90.0)` doc string.
- Same treatment for Python, if the `origin/<version>` indirection can accommodate a SHA.
- Consider having `check_dependency_versions.py` fail loudly when a ref documented as a "pinned tag" resolves to `refs/heads/` — that assertion would have caught both of these automatically, and would keep catching them.

Happy to push these onto the branch if useful.

dcwhite:
Pushed `bd36ee3d4` with the fixes.

**Boost** → `f596b5d3f7dc7195f20591aca3b5d1e33d534c0c`, doc string `"Boost pin -> commit f596b5d3 (branch v1.90.0)"`.

**Python** → new `PYTHON_GIT_TAG` = `b54e57f244c928d7f2b90106a74603462d1662a2`, doc `"Python pin -> commit b54e57f2 (branch 3.13.1)"`. `PythonExternal.cmake` uses it when `USER_PYTHON_VERSION` matches the manifest default and otherwise falls back to `origin/<version>` with a `MESSAGE(STATUS ...)` saying the build is not reproducible — so switching Python versions still works, it just stops being silently unpinned.

Raw SHAs rather than `scirun-pin-2026.07.27` tags, since creating tags on those two repos isn't something I should do from here. The manifest's own "HOW TO UPDATE" already blesses "a full 40-char commit SHA", and I noted the reasoning in the reproducibility section.

## Checker changes

- **New `INVALID` status**: any `*_GIT_TAG` documented as a release tag that resolves to `refs/heads/` is flagged. It's actionable and affects the exit code, whereas transient `ERROR` rows (ls-remote / API failures) deliberately are not — a network blip still can't fail the job.
- Dropped Boost's `boostorg/boost` upstream mapping. `CIBC-Internal/boost` isn't a fork of it, so the comparison was meaningless.
- Split Python into two rows — a commit-pin row checked for drift, and a `PYTHON_VERSION` row still compared against CPython releases — so the "newer Python exists" signal survives.
- The ref assertion only applies to `*_GIT_TAG` vars, so `EIGEN_VERSION` (a tarball) and `PYTHON_VERSION` (a version string that happens to also name a branch) don't false-positive.

## Verification

Ran the checker both ways:

```
# as committed
| Boost  | ✅ OK | f596b5d3f7dc7195f20591ac | branch v1.90.0 | up to date with branch tip f596b5d3f7dc |
| Python | ✅ OK | b54e57f244c928d7f2b90106 | branch 3.13.1  | up to date with branch tip b54e57f244c9 |
EXIT=0

# with BOOST_GIT_TAG restored to "v1.90.0"
| Boost | ❌ INVALID | v1.90.0 | release tag (claimed) | pin 'v1.90.0' is a BRANCH head, not a tag -- it can move and breaks reproducible builds. …
EXIT=1
```

`VERSIONS.cmake` parses under `cmake -P`, and the Python selection resolves to the SHA for `3.13.1` and `origin/3.12.8` when the version is switched.

Also updated the **INVALID** row in `docs/dev_doc/DependencyUpdates.md` and corrected the reproducibility note, which claimed every pin was "either an upstream release tag or an annotated `scirun-pin-<date>` tag" — now also covers raw SHAs.

Not verified: an actual Superbuild run. The Boost SHA is the current tip of `v1.90.0`, so the clone resolves it the same way it did before; worth a clean build before merge regardless.

basisunus (approved):
I was prepared to update each cmake file when versions change. But having them in a centralized place is convenient.

PR #2594, opened 2026-07-30
Fix ReadColorMapXml crash on empty/non-standard colormap files (#2579)

### LINKED ISSUE #2579: ReadColorMapXml: empty output and downstream crash
Sub-issue of #2578.

`ReadColorMapXml` reads an XML file but emits an empty colormap, and SCIRun crashes when that output is piped into any downstream module.

### Root cause

Two distinct problems in the parse/convert path:

1. **Rigid parser.** [`readColorMapXml`](../blob/master/src/Core/ImportExport/ColorMap/ColorMapIEPlugin.cc) only accepts the old ParaView/IDL layout (`<ColorMaps><ColorMap name= space=><Point x= o= r= g= b=/>…`). If the outer `<ColorMap>` element matches but the inner `<Point>` elements don't (different dialect, `RGBPoints`, missing `o` attribute, etc.), `points` ends up empty and **no error is raised** — hence "reads the file but output is empty."

2. **Missing empty-data guard → crash.** An empty-points colormap flows through `createColorMapFromXmlData` into a `ColorMap` with empty `colorData_`. The first downstream color lookup hits `readColorFromArray` in `src/Core/Datatypes/ColorMap.cc`:
   ```cpp
   uint32_t segments = v.size() - 1;   // 0 - 1 → 4294967295 (underflow)
   ColorRGB c0 = v[index];             // out-of-bounds → crash
   ```

### Suggested fix

- Guard `readColorFromArray` / `createColorMapFromXmlData` against empty color data so the module errors cleanly instead of producing a broken colormap.
- Broaden the parser to accept the colormap XML dialect(s) users actually export. **Need a sample file from the reporter to confirm which format** — asked on #2578.

dcwhite:
## Summary

Fixes #2579. `ReadColorMapXml` could crash SCIRun when loading a colormap XML file that parsed to zero color points (a non-standard/unsupported dialect). An empty color array made `v.size() - 1` underflow (unsigned) in `readColorFromArray`, producing a huge segment count and an out-of-bounds access when a downstream module performed a color lookup.

## Changes

- **`Core/Datatypes/ColorMap.cc`** — guard `readColorFromArray` against empty and single-element color arrays, so no colormap can ever crash a lookup (the critical defensive fix).
- **`Core/ImportExport/ColorMap/ColorMapIEPlugin.cc`** — also accept files that omit the `<ColorMaps>` wrapper (a single `<ColorMap>` at the document root, as some ParaView single-preset exports produce); filter explicitly on `<ColorMap>` element names.
- **`Modules/DataIO/ReadColorMapXml.cc`** — reject colormaps that parsed to zero points with a clear error instead of silently emitting an invalid colormap.

## Tests

- 4 new regression unit tests in `ColorMapXmlTests.cc` (empty/single-color no-crash, no-points parses to empty, no-wrapper parses). All pass.
- New example net `src/ExampleNets/regression/ImportExport/colorMapXmlNonStandard.srn5` (auto-discovered as `.Test.ExampleNetwork.colorMapXmlNonStandard_srn`), loading a no-wrapper colormap end to end through `RescaleColorMap` → `ShowField` → `ViewScene`.

Verified in the real app via `SCIRun_test` in regression mode:

| Build | Exit | Behavior |
|-------|------|----------|
| With fix | 0 | net runs clean; colormap flows downstream |
| Without parser fix | 3 | `No colormaps found` → downstream modules starved |

## Companion change (separate repo)

The net's data file `Other/colormaps/purple-8-nonstandard.xml` was pushed to [CIBC-Internal/SCIRunTestData](https://github.com/CIBC-Internal/SCIRunTestData) `master` (commit `f65b3c6`).


dcwhite:
Added support for the reporter's actual colormap dialect.

@Sumientra's `TERcolormap.xml` (attached on #2579) uses a different layout than the modern SCIRun/ParaView one:

```xml
<colormap name="TERcolormap">
  <color a="1" b="0.50" g="0.00" r="0.00" x="0.00"/>
  ...
</colormap>
```

Differences: lowercase `<colormap>` root, `<color>` point elements instead of `<Point>`, and opacity spelled `a` (alpha) instead of `o`. The parser now accepts `<color>` elements, reads opacity from `o` **or** `a`, and recognizes the lowercase root.

Verified end to end via `SCIRun_test` in regression mode:

| Build | Exit | Result |
|-------|------|--------|
| Before this commit | 9 | `No colormaps found in xml file: TERcolormap.xml` → downstream starved |
| After | 0 | colormap loads and flows through the net |

Added `AlternateColorDialect` unit test and a `colorMapXmlTERdialect.srn5` example net. The data file `Other/colormaps/TERcolormap.xml` is on SCIRunTestData `master` (commit `61352f2`).

jessdtate (dismissed):

jessdtate (approved):

Issue #2617, opened 2026-08-04
Investigate self hosted Mac runner to speed up CI


dcwhite:
I've got a M2 mini 2023 running 14.x. Check that out.

@jessdtate has one too.

dcwhite:
## Investigation: where the CI time actually goes

Pulled timing data from the nightly `mac-build` ([run 30903244214](https://github.com/SCIInstitute/SCIRun/actions/runs/30903244214)) plus the raw job logs. Short version: **the CI is slow for reasons a faster Mac doesn't fix.** The mini is still worth doing, but it's third in line behind two nearly-free changes.

### 1. The macOS and Linux builds are entirely single-threaded

`build.sh` defaults `makeflags=` to empty (line 169), and `reusable-build.yml` calls `./build.sh` with no `-j` (line 95). The log confirms it:

```
Make Flags:
```

...followed by 1508 `Building CXX object` lines in strict sequence, ~3s apart, one at a time. On a `macos-14-arm64` runner (3 vCPU / 7 GB M1), two of three cores sit idle for 62 minutes. Same story on Linux. Windows is the only leg that parallelizes (`cmake --build --parallel`).

Breakdown of the 63-minute `mac-gui-14-arm` job -- 62 min of it is a single step:

| phase | time |
|---|---|
| externals (Eigen -> Boost, strictly serial) | ~18 min |
| Boost alone (download + bootstrap + build) | ~12 min |
| inner SCIRun build | ~44 min |
| everything else (checkout, Qt, smoke test) | ~1 min |

Boost is also built `variant=release,debug` (`BoostExternal.cmake:322`) -- both configurations, in every job, though CI only ever consumes Release.

### 2. There is no caching of any kind

No `actions/cache`, no ccache/sccache anywhere in the workflows. Every job re-downloads and rebuilds Boost, Python, Qwt, Teem, Freetype, SQLite and Cleaver2 from scratch.

### 3. The nightly is queue-bound at 5 concurrent macOS jobs

13 mac jobs, **21 hours of macOS compute**, 5h31m wall clock. Exactly 5 jobs start at 11:03:5x and each subsequent start lines up with a prior completion -- GitHub's public-repo cap of 5 concurrent macOS runners.

### 4. Nearly half that compute is Intel, which an M2 can't touch

[...]

PR #2619, opened 2026-08-05
CI: stop building serially; auto-detect make job count (#2617)

dcwhite:
Step 1 of the #2617 investigation ([comment](https://github.com/SCIInstitute/SCIRun/issues/2617#issuecomment-5186114800)).

## The bug

`build.sh` defaults `makeflags` to empty (line 169) and `reusable-build.yml` calls `./build.sh` with no `-j` (line 95). Every macOS and Linux build — CI *and* local — has been fully serial. From the nightly `mac-gui-14-arm` log:

```
Make Flags:
```

followed by 1508 `Building CXX object` lines in strict sequence, ~3s apart. Two of the runner's three cores idle for 62 minutes. Windows was unaffected because it uses `cmake --build --parallel`.

## The fix

Default to a detected job count when the caller passes no `-j`:

```
min(cores, RAM_GB / 2), floor 1
```

Capped by memory as well as cores because several Field/Mesh translation units instantiate enough template code to want well over a gigabyte apiece, and core count alone would OOM a 7 GB runner. Measured against the runner profiles:

| runner | cores | RAM | jobs |
|---|---|---|---|
| `macos-*-arm64` | 3 | 7 GB | `-j3` |
| `macos-*-intel` | 4 | 14 GB | `-j4` |
| `ubuntu-latest` / `-arm` | 4 | 16 GB | `-j4` |
| M2 mini (16 GB) | 8 | 16 GB | `-j8` |
| M2 mini (8 GB) | 8 | 8 GB | `-j4` |

An explicit `-j` still wins, and `-j1` restores the old serial behaviour.

## Two steps that escape the jobserver

Externals that recurse via `$(MAKE)` inherit the top-level jobserver and are bounded by it automatically — they must *not* get a second independent budget. Two do not recurse:

- **Qwt** shells out to `cmake --build`, which honours the now-exported `CMAKE_BUILD_PARALLEL_LEVEL`.
- **Boost's b2** is not a make at all and had no `-j`, so it built every library serially no matter what the outer make was told. `Superbuild.cmake` now derives `SUPERBUILD_PARALLEL_JOBS` from the same env var, falling back to `ProcessorCount` for people who configure the Superbuild with `cmake` directly.

## Expected effect

The mac build step is 62 of the job's 63 minutes: ~18 min externals (Boost alone ~12) and ~44 min inner SCIRun build. Both parts are compile-bound and serial today. Linux jobs are worse — 140 min each.

Queue time should fall with it: the nightly's 13 mac jobs contend for GitHub's 5-concurrent-macOS cap, and across 79 recent mac jobs the mean queue was 68 min (median 47, p90 177). That queue is our own matrix saturating the slots, so cutting run time cuts waiting proportionally.

## Verification

`detect_parallel_jobs` was unit-tested against the profiles in the table above; `bash -n` clean. Real numbers will come from this PR's own CI run versus the ~100 min `mac-gui` and ~140 min `linux-gui` baselines on master.

Deliberately **not** in this PR: Boost's `variant=release,debug` (builds both configs though CI only consumes Release) and Superbuild external caching — separate PRs.


basisunus (approved):
I think I had the issue that parallel build failed initially. I guess that fixed it.

Issue #2627, opened 2026-08-05
Superbuild: give externals a real install stage so there is something compact to cache


dcwhite:
Structural prerequisite for #2520. Surfaced by the #2617 timing investigation, after both caching attempts (#2513, #2576) failed for **opposite** reasons that trace to this one cause.

## The problem

**20 of 23 externals use `INSTALL_COMMAND ""`.** Only `PythonExternal.cmake`, `QwtExternal.cmake` and `TetgenExternal.cmake` actually install anything.

Everything else leaves its artifacts wherever they were built:

- `Externals/Build/<name>/` for the CMake-based externals
- `Externals/Source/Boost_external/stage/lib/` for Boost

`Externals/Install/` therefore holds almost nothing — generated `*Config.cmake` files and the three real installs. The Superbuild compensates by exporting `SCI_<NAME>_LIBRARY_DIR` / `SCI_<NAME>_INCLUDE` variables pointing straight into `Build/` and `Source/`.

## Why it blocks caching

There is no directory that means "the built externals", so a cache has to be either empty or enormous:

| attempt | cached | result |
|---|---|---|
| #2513 | `Install/` + `Stamp/` | **17–66 KB**. Reverted in 431dd8246: *"compiled artifacts live in `Externals/Build/`, not `Externals/Install/`"* |
| #2576 | whole `Externals/` tree | **~2.4 GB/job**; 13 mac jobs ≈ 31 GB against GitHub's 10 GB/repo cap → jobs LRU-evict each other mid-run → zero warm hits |

Note for anyone reading the #2576 thread: the review comment there recommending *"cache only `Externals/Install`"* is wrong — that is precisely what #2513 already disproved.

## The change

1. Give each external a real install (`INSTALL_COMMAND` / `CMAKE_INSTALL_PREFIX` into `Externals/Install/<name>`).
2. Repoint the exported `SCI_*_LIBRARY_DIR` / `SCI_*_INCLUDE` variables at `Install/` instead of `Build/` and `Source/`.
3. Boost needs care — b2 stages into `Source/boost/stage/lib`; see also the double-build problem in the Boost issue filed alongside this one.

Estimated result: a self-contained `Install/` tree of roughly 150–300 MB per job instead of 2.4 GB, which fits the whole matrix comfortably under 10 GB and makes #2520 straightforward.

## Also worth having independently of caching

- Packaging currently reaches into build trees for its dependencies.
[...]

PR #2643, opened 2026-08-05
CI: make pr.yml answer for the merge queue (merge_group support)

dcwhite:
Workflow-side prerequisite for enabling a GitHub merge queue on `master`. **This does not turn the queue on** — that is a repository setting. Without this change the queue cannot be enabled at all, because `ci-gate` would never report against a queued batch and every entry would time out.

## Why

Branch protection currently has `strict: true` ("require branches to be up to date"), with `ci-gate` as the only required context. Every merge to master invalidates every other open PR's `ci-gate`, and `ci-gate` is a full linux + mac + windows Superbuild.

Recent successful `pr.yml` runs took **210, 301, 311, 348, 409, 410 and 411 minutes** — 5–7 hours, mostly macOS queueing against the 5-slot cap. With 8 approved PRs open, `strict` turns that into ~50 hours of strictly serialized CI: merge one, six others go `BEHIND`, update each branch, wait 5–7h, merge, repeat.

A merge queue keeps the guarantee `strict` provides — nothing lands untested against the real tip — but tests a *batch* of PRs as one candidate, and never requires a PR branch to be updated. Cost per batch instead of cost per PR.

## What changed

A merge group has no pull request behind it, which breaks three things:

- **`on:` gains `merge_group:`** so `ci-gate` reports against the queued batch.
- **Concurrency group.** It keyed off `github.event.pull_request.number`, empty in a merge group — every queue entry would collide in a single group and cancel its predecessor. Now keyed off the entry's `head_ref`, with `cancel-in-progress` limited to pull requests. Cancelling a queue entry reports a failed check and ejects the PR from the queue, so it must not happen here.
- **`changes` job.** It called `gh pr view` for the file list. In a merge group it diffs `base_sha..head_sha`, classifying the batch by everything it proposes to land — one code PR anywhere in a batch pulls the whole batch through a real build. Adds a `fetch-depth: 0` checkout, on the `merge_group` path only.

`reusable-build.yml` needs no change — plain `actions/checkout@v6` resolves to the triggering ref, which in a merge group is the speculative merge commit.

## Testing

YAML parses; the `docs_only` classifier was exercised on both paths (docs-only → skip, mixed batch → build, empty → build). The `merge_group` path cannot be exercised until the queue is enabled — see below.

## Suggested queue settings, if we enable it

The 5–7h builds invert the usual advice:

- **Max entries to build concurrently: 1.** The default speculates (testing `PR1`, `PR1+2`, `PR1+2+3` in parallel). Correct for fast builds; here it would consume the entire macOS quota on candidates that get discarded.
- **Minimum group size: 5, wait time ~60 min.** This is the setting that does the work — it batches, so ~8 approved PRs become roughly one build instead of eight.
- **Maximum group size: 10.**
- **Turn `strict` off once the queue is on.** The queue supersedes it; both together reintroduce the branch-update cycle.

Two caveats worth weighing: `enforce_admins: true` means the queue becomes the only way in, with no manual merge as an escape hatch. And since this would be our first `merge_group` run, worth queueing one low-stakes PR to confirm `ci-gate` reports correctly before trusting a batch to it.

The 5–7h build remains the underlying multiplier — #2627 is what makes it cheap. This makes it cheap *per batch* in the meantime.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -21,13 +21,31 @@ name: pr
 # regression-tests.yml / windows-build-script.yml and runs on master pushes and
 # nightly -- not on every pull request. Use the `workflow_dispatch` trigger on
 # those workflows to run the full matrix against a risky branch on demand.
+#
+# This workflow also answers for the merge queue. A queued entry is a synthetic
+# commit -- master's tip plus every PR ahead of it in the queue -- pushed to a
+# `gh-readonly-queue/master/...` ref. GitHub then waits for the same required
+# context (`ci-gate`) against that ref before landing the batch. So a queue
+# entry is a normal run of everything below, just on a ref that no PR owns.
+#
+# Two things do NOT survive the switch from `pull_request` to `merge_group`,
+# because a merge group has no pull request behind it:
+#
+#   * `github.event.pull_request.number` is empty. It appeared in the old
+#     concurrency group and in the `changes` job; both are handled below.
+#   * `cancel-in-progress` must be off here. Cancelling a queue entry reports a
+#     failed check, which ejects the PR from the queue -- the exact stall the
+#     queue is meant to remove. Pull requests still cancel as before.
 
 on:
   pull_request:
+  merge_group:
 
 concurrency:
-  group: pr-${{ github.event.pull_request.number }}
-  cancel-in-progress: true
+  # One group per PR, or per queue entry. `head_ref` is unique per entry, so
+  # speculative batches never cancel each other.
+  group: ${{ github.event_name }}-${{ github.event.pull_request.number || github.event.merge_group.head_ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   changes:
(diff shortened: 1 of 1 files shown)


dcwhite:
## Status: parked pending data — do not merge-and-enable yet

Recording the reasoning from the discussion that produced this PR, so the decision is here rather than in someone's head.

### Decision

**Hold.** Let the overnight build + Slack reporting (#2574) merge and report for a few days first, then decide between enabling a merge queue and simply dropping `strict`. This PR is inert until a queue actually exists, so there is no cost to it sitting open, and no need to rush it in.

### Why hold rather than enable

The queue-vs-`strict: false` choice turns on merge rate and macOS capacity, and we don't have either number yet.

Merge queues are designed for repos where the tip moves faster than CI can keep up — steady, high merge rate. SCIRun's traffic is bursty: a batch of ~9 approvals lands after a review meeting, then it's quiet for weeks. That's the pattern queues fit *worst*. Batching genuinely helps a burst, but the minimum-group-size timer that makes batching work is also what makes a PR sit visibly idle for an hour during the quiet stretches.

It's entirely possible the right answer is just `strict: false`, relying on master-push and nightly builds to catch the semantic conflicts `strict` protects against. No deploys, and a broken master here is annoying rather than costly. That would cost zero new concepts for anyone. Worth letting the data decide instead of adopting the heavier mechanism preemptively.

### Numbers worth collecting once #2574 is reporting

1. **Uncontended nightly duration per platform.** The 210–411 min figures in the PR description are mostly queue wait, not build time, and the two can't be separated without a clean run.
2. **Merges per day, averaged.** This is the one that settles it, and the easy one to forget. Merge rate x CI-hours-per-merge, compared against macOS-hours-per-day: under capacity means `strict: false` suffices and a queue is overhead; at the ceiling means batching is the only lever that helps.
3. **Unit test pass/fail stability across nights.** Separate track — it's what tells us when `continue-on-error` can come off the ctest steps in `reusable-build.yml`, at which point ci-gate becomes a real test gate with no change needed here.

### Measurement confound

Draining the ~8 currently-approved PRs saturates the same 5 macOS slots the nightly builds need. If the drain overlaps the measurement window, those nights are contention numbers, not baselines. Either drain first and start the clock after, or record which nights overlapped and read them as worst case.

### One correction to the record

Branch protection has `dismiss_stale_reviews: true`, but this does **not** cost the approval when a PR is brought up to date. "Update branch" creates a base-into-head merge commit, which doesn't change the PR's diff, so the approval survives. The bottleneck is therefore pure machine time — it does not send anything back through review.

jessdtate (approved):

Issue #2657, opened 2026-08-05
Boost's bundled b2 cannot bootstrap under Visual Studio 2026 (unknown toolset 'vcunk')


dcwhite:
The `windows-2025-gui-qt6-vs2026` nightly job fails while bootstrapping Boost.Build:

```
Bootstrapping b2
Building Boost.Build engine
Found with vswhere C:\Program Files\Microsoft Visual Studio\18\Enterprise
###
### "Unknown toolset: vcunk"
###
### Toolsets supported by this script are: borland, como, gcc,
###     clang, clang-win, gcc-nocygwin, intel-win32, mingw,
###     vc12, vc14, vc141, vc142, vc143
###
Failed to build Boost.Build engine.
```

First seen in [nightly run 30998726422](https://github.com/SCIInstitute/SCIRun/actions/runs/30998726422).

## Root cause

`Superbuild/BoostExternal.cmake` pins `CIBC-Internal/boost` at `v1.90.0`, whose
`tools/build` submodule sits at b2 commit `be69c44857e34fae250aa1321808a05c1d88bfde`.
That b2 predates Visual Studio 2026 in two places:

1. `src/engine/guess_toolset.bat` has no `VS180COMNTOOLS` case, so vswhere's
   detection of VS 18 falls through to the catch-all and sets `B2_TOOLSET=vcunk`.
2. `src/engine/config_toolset.bat` never dispatches `vcunk` — a `:Config_VCUNK`
   label exists in the file but nothing in the `:Start` block calls it, so
   `_known_` is never set and bootstrap reports "Unknown toolset". This is an
   upstream bug at that commit, not something our invocation can work around.

Upstream b2 `main` has since fixed both: `guess_toolset.bat` maps
`VS180COMNTOOLS` to `vc145`, and `config_toolset.bat` dispatches `vc145` to a
new `:Config_VC145`.

## Second-order problem

Forcing the bootstrap through is not sufficient. `src/tools/msvc.jam` at the
pinned commit has `.known-versions = 14.3 14.2 14.1 14.0 ...` — no 14.5 — and
its fallback path-sniffing regex is `MSVC\\14.[34]`. VS 2026's compiler lives in
`VC\Tools\MSVC\14.51.36231`, which matches none of the version patterns, so even
a successfully bootstrapped b2 would misconfigure the msvc toolset. The fix has
to be a newer b2, not a flag passed from the Superbuild.

## Proposed fix

[...]

Issue #2548, opened 2026-07-02
Upgrade Boost to 1.91


dcwhite:
## Context

Noted during review of [PR #2463](https://github.com/SCIInstitute/SCIRun/pull/2463) (cmake/Boost/Qt modernization), which brought Boost onto the imported-target (`Boost::*`) config-based setup:

- [`Superbuild/BoostExternal.cmake:28`](https://github.com/SCIInstitute/SCIRun/pull/2463#discussion_r3191077779) — "Seems pretty well commented. The test will be to see if we can upgrade easily to 1.91, which just came out :)"

Related prior (closed) attempts: #2442 (1.89, old cmake system), #2453 (general "update boost to something").

## Task

- Bump the Superbuild Boost version to 1.91.
- Verify the config-based `find_package(Boost CONFIG)` / imported-target setup from #2463 still resolves cleanly against 1.91 on Linux, macOS, and Windows.
- Watch for any removed/deprecated Boost components used by SCIRun that changed between the current version and 1.91.

dcwhite:
Extra motivation: the VS 2026 nightly jobs cannot build Boost at all on 1.90 — b2's bootstrap dies with `"Unknown toolset: vcunk"`. 1.91's bundled b2 adds full `vc145`/14.5 support (engine bootstrap *and* `msvc.jam`), so this upgrade is what unblocks those builds. Details and the 1.90-vs-1.91 comparison in #2657.

PR #2658, opened 2026-08-05
CI: unbreak the VS 2026 nightly builds (CPython toolset, Qwt logging, non-blocking jobs)

dcwhite:
The first nightly to run on the `windows-2025-vs2026` runner image
([run 30998726422](https://github.com/SCIInstitute/SCIRun/actions/runs/30998726422))
failed in three unrelated places, all in Superbuild externals — nothing in
SCIRun's own code. This fixes two of them and stops the third from failing the
nightly.

## 1. CPython picked the VS 2015 toolset

`windows-2025-headless-vs2026` died three minutes in:

```
error MSB8020: The build tools for Visual Studio 2015 (Platform Toolset = 'v140')
cannot be found  [...\Python_external\PCbuild\_freeze_module.vcxproj]
```

CPython 3.13.1's `PCbuild/python.props` maps `VisualStudioVersion` 15/16/17 to
v141/v142/v143 and falls back to **v140** for anything it doesn't recognize. VS
2026 reports `18.0`, so the fallback fired. Upstream CPython `main` has since
added the `'18.0' → v145` line; our pinned fork predates it.

`Superbuild/PythonExternal.cmake` now derives `/property:PlatformToolset` from
`CMAKE_VS_PLATFORM_TOOLSET` and passes it to `PCbuild\build.bat` and both
`msbuild` invocations. `build.bat` forwards any argument it doesn't recognize
straight to MSBuild (verified against CPython 3.13.1's `build.bat`, which ends
its MSBuild command line with `%1 %2 ... %9`). The variable is empty for non-VS
generators, and on windows-2022 it resolves to the v143 that `python.props`
would have chosen anyway — so this is a no-op everywhere except VS 2026.

## 2. Qwt failed invisibly

The same run's GUI job lost Qwt, but `LOG_BUILD 1` meant CI printed only
`Command failed: 1` plus a path to a stamp log that never leaves the runner.
`LOG_OUTPUT_ON_FAILURE 1` (CMake >= 3.14; the Superbuild requires 3.20) echoes
the captured log when a step fails, so the next nightly will actually say what
broke. **The Qwt failure itself is still undiagnosed** — this change is what
makes diagnosing it possible.

## 3. Boost: not fixable here, so the jobs stop blocking

The b2 bundled with Boost 1.90 cannot bootstrap under VS 2026 at all
(`"Unknown toolset: vcunk"`) — see #2657 for the analysis. The upgrade to Boost
1.91 (#2548) fixes it outright; there is no useful Superbuild-side workaround,
since that b2's `msvc.jam` doesn't know 14.5 either.

Until then, both VS 2026 jobs are marked experimental through a new
`allow-failure` input on `reusable-build.yml`. `continue-on-error` is rejected on
a job that calls a reusable workflow — GitHub's workflow parser allows only
`concurrency, if, name, needs, permissions, secrets, strategy, uses, with`
there — so the flag is plumbed through as an input and applied to the job inside
the reusable workflow. The jobs still run and still report red on their own;
they just no longer fail the run.

That alone would have made the nightly report a clean green, because
`nightly-slack.yml` keys off `workflow_run.conclusion`. It now also queries the
run's jobs and, on a "successful" run that contains failed jobs, goes yellow and
names them.

## Also

Corrects the comment claiming windows-2025 ships a side-by-side VS 2022. The
image carries a single VS Enterprise 2026 install; v143 is present as a
component of it (`Microsoft.VisualStudio.Component.VC.14.44.17.14.x86.x64`), so
`-T v143` remains available as an escape hatch.

## Testing

Not verifiable locally — these paths are Windows-only and the externals take
hours to build. The Python and Qwt changes get their first real exercise on the
next nightly. The workflow changes were validated against GitHub's own workflow
parser schema (`actions/languageservices`), which is what caught the invalid
`continue-on-error` placement.

Fixes nothing on its own for #2657 (tracked separately, resolved by #2548).


basisunus (commented):
i haven't got a copy of vs2026 yet, but the fix seems no problem to me.

dcwhite:
## Local VS 2026 repro notes

Notes for validating this PR by hand on a machine with VS 2026 installed, since
CI can't verify the Python and Qwt changes and the Qwt failure is still
undiagnosed.

### Expect a full build to fail anyway — at Boost

A complete Superbuild under VS 2026 will **still** fail with this PR applied,
because the Boost `vcunk` bootstrap failure (#2657) is deliberately not fixed
here. That is the expected outcome, not a regression.

Worth knowing why the nightly doesn't already show this: in
[run 31249565048](https://github.com/SCIInstitute/SCIRun/actions/runs/31249565048)
the headless job died at `Python_external` about 90 seconds in and MSBuild tore
down the remaining targets, so `Boost_external` never reached its bootstrap step
— it appears in that log only as a mention, with no build steps and no
`Completed`. These externals did complete: Zlib, Eigen, Teem, Tetgen, Cleaver2,
Freetype, GLM, Glew, LodePng, SQLite, SpdLog, Tny, BrainStimulator,
FwdInvToolkit, SCIRunTestData.

So fixing CPython's toolset just advances the headless failure to Boost. Build
the suspect targets individually instead of chasing a green full build.

### Configure: headless (mirrors `windows-2025-headless-vs2026`)

This is the one that exercises fix 1 — `BUILD_WITH_PYTHON` defaults to `ON`, so
`Python_external` is in play.

```powershell
git fetch origin claude/vs2026-nightly-build-failure-f02f17
git switch claude/vs2026-nightly-build-failure-f02f17
mkdir bin-vs2026-headless
cd bin-vs2026-headless
cmake -G "Visual Studio 18 2026" -A x64 -DBUILD_HEADLESS:BOOL=ON -DBUILD_TESTING:BOOL=ON -DREGENERATE_MODULE_FACTORY_CODE:BOOL=OFF -S ..\Superbuild -B .
```

```powershell
cmake --build . --target Python_external --config Release
```

Success here means `/property:PlatformToolset` reached MSBuild and
`_freeze_module.vcxproj` stopped asking for v140. If it still fails, the thing
to check is whether CPython 3.13.1's `PCbuild\build.bat` really forwards the
unrecognized argument through to MSBuild — that is the one assumption in fix 1
that was verified by reading the script rather than by running it.

### Configure: GUI (mirrors `windows-2025-gui-qt6-vs2026`)

Python is off in this job, which is why it fails at Boost and Qwt rather than at
CPython. Point `Qt_PATH` at a local Qt 6.3.x.

```powershell
mkdir bin-vs2026-gui
cd bin-vs2026-gui
cmake -G "Visual Studio 18 2026" -A x64 -DBUILD_WITH_PYTHON=OFF -DSCIRUN_QT_MIN_VERSION=6.3.1 -DQt_PATH="C:\Qt\6.3.2\msvc2019_64" -DREGENERATE_MODULE_FACTORY_CODE:BOOL=OFF -S ..\Superbuild -B .
```

```powershell
cmake --build . --target Qwt_external --config Release
```

Qwt does not depend on Boost — both failed independently in the same GUI job —
so this runs without Boost in the way. **This is the highest-value thing to run
locally.** Nobody has seen the actual Qwt error yet: `LOG_BUILD 1` sent it to
`Externals/Stamp/Qwt_external/Qwt_external-build-*.log`, which never leaves the
runner, and all CI printed was `Command failed: 1`. Building the target locally
puts the real message on the console with no dependency on the
`LOG_OUTPUT_ON_FAILURE` change in this PR.

Note the CI configure lines also pass `-DCMAKE_C_COMPILER_LAUNCHER=sccache` and
`-DCMAKE_CXX_COMPILER_LAUNCHER=sccache`; those are omitted above since they're
CI-only.

### Bonus: confirm Boost 1.91's b2 empirically (2 minutes, no Superbuild)

#2657 argues from upstream changelogs that Boost 1.91's b2 fixes `vcunk`. With
VS 2026 on hand that can be checked directly, which de-risks #2548. Fetching
just the `tools/build` submodule at the 1.91 tag avoids cloning all of Boost:

```powershell
git clone https://github.com/boostorg/boost.git --branch boost-1.91.0 --depth 1 --no-recurse-submodules C:\tmp\b191
cd C:\tmp\b191
git submodule update --init --depth 1 tools/build
.\tools\build\src\engine\build.bat
```

The question is whether it detects VS 18 / toolset 14.5 instead of printing
`### "Unknown toolset: vcunk"` and listing `vc12 … vc143`.

### Useful to report back

- The Qwt error text — that's the open unknown in this PR.
- Whether `Python_external` builds clean.
- Whether 1.91's `build.bat` recognizes VS 18.

dcwhite:

dcwhite:
Ran the VS 2026 GUI path locally. **The Qwt failure is Qt, not Qwt** — here is the error text this PR was meant to surface, plus a new follow-on failure for `Python_external`.

### Environment

- VS 2026 Professional 18.8.12023.21, MSVC 14.51.36231 (`cl` 19.51.36252)
- CMake 4.3.1 — the copy bundled inside VS 18; 3.31 has no `Visual Studio 18 2026` generator
- Qt 6.6.0 msvc2019_64, `-G "Visual Studio 18 2026" -A x64`, Release
- master @ b950285ed, so #2694 is in

14 externals built clean (Zlib, SQLite, Eigen, GLM, SpdLog, Tny, LodePng, Glew, Freetype, Teem, Tetgen, Cleaver2, BrainStimulator, FwdInvToolkit). Qwt and Python failed; Boost and SCIRun never started.

### 1. The Qwt error — Qt 6.x vs the VS 2026 STL

```
qvarlengtharray.h(963,39): error C2065: 'stdext': undeclared identifier
qvarlengtharray.h(963,39): error C3861: 'stdext': identifier not found
```

Nothing to do with Qwt's own code. From `QtCore/qcompilerdetection.h`, in the MSVC branch:

```cpp
#  define QT_MAKE_UNCHECKED_ARRAY_ITERATOR(x) stdext::make_unchecked_array_iterator(x)
#  define QT_MAKE_CHECKED_ARRAY_ITERATOR(x, N) stdext::make_checked_array_iterator(x, size_t(N))
```

MSVC 14.51 removed `stdext::make_checked_array_iterator` — the checked-iterator extensions, deprecated since VS2017. `QVarLengthArray::erase` calls the macro, so any translation unit instantiating it fails.

Qt does carry a `(x)` fallback further down the same header, but it is `#ifndef`-guarded and the MSVC branch defines the macro unconditionally first — so **`-DQT_MAKE_CHECKED_ARRAY_ITERATOR=...` on the command line does not work** (C4005 redefinition, the header wins). There is no Superbuild-side workaround; it needs Qt >= 6.7, where the `stdext` usage is gone.

The Qt versions on hand are affected identically:

| Qt | `qcompilerdetection.h` |
|---|---|
| 6.2.4 | lines 103–104, unconditional `stdext::` |
| 6.3.0 | lines 103–104, unconditional `stdext::` |
| 6.6.0 | lines 78–79, unconditional `stdext::` |

So the GUI job's Qt 6.3.x hits this as well — the repro above (`-DQt_PATH="C:\Qt\6.3.2\msvc2019_64"`) should reproduce it directly. And it is not confined to Qwt: SCIRun's own sources pull in `QVarLengthArray` too, so a Qt bump is a prerequisite for any GUI build on VS 2026, not just for clearing this one external.

### 2. `Python_external` — #2694 works, with a new failure behind it

The toolset fix does its job: no `MSB8020`, no v140, and CPython 3.13 compiles. It now fails later, and only on the UWP targets:

```
include\experimental\coroutine(37,1): error C2338: static assertion failed:
  'error STL1011: The /await compiler option, <experimental/coroutine>, <experimental/generator>,
   and <experimental/resumable> are deprecated by Microsoft and will be REMOVED SOON ...
   define _SILENCE_EXPERIMENTAL_COROUTINE_DEPRECATION_WARNINGS to suppress this error for now.'
[ ...\PCbuild\python_uwp.vcxproj ]
[ ...\PCbuild\pythonw_uwp.vcxproj ]
```

`pythoncore`, `python`, `pythonw`, the launchers and the stdlib extension modules all build. `PythonExternal.cmake` builds all of `PCbuild/pcbuild.sln`, which drags in two UWP artifacts SCIRun never uses. Either define `_SILENCE_EXPERIMENTAL_COROUTINE_DEPRECATION_WARNINGS` for that build, or build only the targets we need.

### 3. Boost — no data

The build aborted at Qwt/Python before `Boost_external` started, so the 1.91 b2 / `vcunk` question is still unanswered from here.

### Also: Qt component discovery breaks under the documented GUI repro

Independent of VS 2026, but anyone following the repro above will hit it first. `find_package(Qt6 ... HINTS ${Qt_PATH})` does not propagate into the `find_dependency()` calls inside Qt's own config files, which search `CMAKE_PREFIX_PATH` — so `Qt6DBus` -> `Qt6DBusTools` -> `Qt6CoreTools` fails at configure time even with all three present under `Qt_PATH`:

```
Could not find a package configuration file provided by "Qt6CoreTools"
(requested version 6.6.0) with any of the following names: Qt6CoreToolsConfig.cmake ...
Failed to find required Qt component "DBus".
```

`Superbuild.cmake` forwards 8 `Qt*_DIR` cache vars to the inner build but not `DBusTools`, `WidgetsTools`, `Svg`, `PrintSupport` or `OpenGLWidgets`, which is why this surfaces intermittently rather than always. Prepending `Qt_PATH` to `CMAKE_PREFIX_PATH` at both find sites (`Superbuild/Superbuild.cmake` and `src/CMakeLists.txt`) fixes it; that is on `claude/srbuilds26-vs2026-config-ae3733` if useful, happy to split it into its own PR.

dcwhite:
Follow-up to the above, with Qt 6.8.3 installed and `BUILD_WITH_PYTHON=OFF` to get past CPython and reach Boost. Same VS 2026 Professional 18.8.12023.21 / MSVC 14.51.36231 / CMake 4.3.1 setup, Release x64, GUI (not headless).

### Qwt is fixed by the Qt bump — confirmed

`Qwt_external` now **builds clean**. 15 externals complete (the previous 14 plus Qwt). Qt 6.8.3's `qcompilerdetection.h` carries the guard 6.6 lacked:

```cpp
#  if _MSC_VER < 1938 // stdext is deprecated since VS 2022 17.8
#    define QT_MAKE_CHECKED_ARRAY_ITERATOR(x, N) stdext::make_checked_array_iterator(x, size_t(N))
```

`cl` 19.51 is `_MSC_VER` 1951, so the branch is skipped and the `(x)` fallback applies. Verified on 6.8.3 specifically; I did not test 6.7.x.

### Boost: `vcunk` confirmed empirically

#2657 argued this from changelogs. Direct confirmation, Boost 1.90.0 (`project(Boost VERSION 1.90.0 ...)`), running `bootstrap.bat` by hand in the cloned tree:

```
Building Boost.Build engine
Found with vswhere C:\Program Files\Microsoft Visual Studio\18\Professional
###
### "Unknown toolset: vcunk"
###
### Toolsets supported by this script are: borland, como, gcc,
###     clang, clang-win, gcc-nocygwin, intel-win32, mingw,
###     vc12, vc14, vc141, vc142, vc143
###
Failed to build Boost.Build engine.
```

vswhere resolves VS 18 correctly; it is `guess_toolset` that has no mapping for it. The supported list ends at `vc143`, so toolset 14.5 is simply unrepresentable in this b2. No Superbuild-side workaround — consistent with #2657, and #2548 (Boost 1.91) remains the fix.

`SCIRun_external` never ran, since it depends on Boost.

### New, separate bug: `bootstrap.bat` is invoked as a bare relative name

Before reaching `vcunk` at all, the Superbuild step fails outright:

```
Bootstrapping b2
'bootstrap.bat' is not recognized as an internal or external command,
operable program or batch file.
```

`Superbuild/BoostExternal.cmake`:

```cmake
IF(WIN32)
  SET(_B2_CMD ${SOURCE_DIR}/b2.exe)        # absolute
  SET(_B2_BOOTSTRAP_CMD bootstrap.bat)     # bare relative name
ELSE()
  SET(_B2_CMD ${SOURCE_DIR}/b2)
  SET(_B2_BOOTSTRAP_CMD ./bootstrap.sh)
ENDIF()
```

The step does set `WORKING_DIRECTORY ${SOURCE_DIR}`, and `bootstrap.bat` is present there — but resolving a bare command name from the current directory depends on the environment. This machine has **`NoDefaultCurrentDirectoryInExePath=1`** set, which tells cmd not to search the working directory, so the step dies before b2 is ever built. Clearing that variable for a single process is what let the `vcunk` result above be observed at all.

GitHub's runner images do not set that variable, which is why CI reaches `vcunk` directly and this never appeared there. It is still worth fixing: `${SOURCE_DIR}/bootstrap.bat` matches both the `_B2_CMD` line directly above it and the Unix branch's `./bootstrap.sh`.

Note this is not sufficient on its own for such an environment — Boost's own `build.bat` invokes `guess_toolset.bat` the same bare way, and fails identically:

```
'guess_toolset.bat' is not recognized as an internal or external command
### "Could not find a suitable toolset."
```

So with that variable set, Boost cannot bootstrap regardless of what the Superbuild passes. The one-line Superbuild fix is still correct; it just moves the failure to the upstream one.

### Summary of the VS 2026 GUI path as it stands

| External | Status |
|---|---|
| 15 externals incl. Qwt | build clean with Qt 6.8.3 |
| `Python_external` | UWP targets fail (`STL1011`), see previous comment |
| `Boost_external` | blocked — `vcunk` on 1.90, needs #2548 |
| `SCIRun_external` | not reached |

dcwhite:
Ran the bonus experiment from the review comment — **Boost 1.91's b2 does recognize VS 2026.** Same machine as above (VS 2026 Professional 18.8.12023.21, MSVC 14.51.36231).

```powershell
git clone https://github.com/boostorg/boost.git --branch boost-1.91.0 --depth 1 --no-recurse-submodules C:\temp\b191
cd C:\temp\b191
git submodule update --init --depth 1 tools/build
cd tools\build\src\engine
.\build.bat
```

Exit 0, and `b2.exe` (B2 5.4.2) is produced. Contrast with the pinned 1.90, which fails `"Unknown toolset: vcunk"` on the identical machine:

| | Boost 1.90.0 | Boost 1.91.0 |
|---|---|---|
| `build.bat` bootstrap | `"Unknown toolset: vcunk"`, exit 1 | exit 0, `b2.exe` built |
| Toolsets listed by `build.bat` | `... vc141, vc142, vc143` | `... vc141, vc142, vc143, **vc145**` |

The `vc145` entry is the entire difference. 1.90's table ends at `vc143`, so MSVC 14.5 has no mapping and `guess_toolset` falls through to `vcunk`; 1.91 maps it and the engine compiles with the 14.51 toolset.

So #2548 is confirmed as the fix for #2657 on real VS 2026 hardware, rather than inferred from changelogs.

(One caveat for anyone reproducing: if `NoDefaultCurrentDirectoryInExePath=1` is set in the environment, `build.bat` cannot find its own `guess_toolset.bat` and reports `"Could not find a suitable toolset."` instead of `vcunk` — that is the environment issue described in the previous comment, not a Boost version difference. Clear the variable for the shell before running.)

PR #2659, opened 2026-08-05
Qt6: replace Q_FOREACH over non-shared containers with range-based for

dcwhite:
Qt 6.10 deprecates `Q_FOREACH` on containers that are not implicitly shared, because the macro silently copies the container:

```
qforeach.h:51:5: warning: 'warnIfContainerIsNotShared<std::set<SCIRun::Gui::ConnectionLine *>>' is deprecated
Port.cc:668:3: note: in instantiation of ... requested here
```

Two sites in the codebase trigger it:

- `PortWidget::deleteConnections/deleteConnectionsLater/trackConnections` — `std::set<ConnectionLine*> connections_`
- `ModuleProxyWidget::createPortPositionProviders` — `boost::join` range over the two port deques

`deleteConnections()` was quietly depending on the copy: `~ConnectionLine` calls back into `PortWidget::removeConnection`, which erases from `connections_` and would invalidate a plain range-for iterator. That loop now copies the set explicitly, with a comment saying why. The others iterate in place.

The `QList<QAction*>` loop in `Port.cc` was converted for consistency; QLists are implicitly shared so it never warned.

The remaining ~64 `Q_FOREACH` uses (mostly `scene_->items()` in `NetworkEditor.cc`, plus the vendored CTK widgets) all iterate implicitly-shared Qt containers and compile clean — left alone.

## Testing

Behavior-preserving; no functional change intended. Not separately built here — please let CI cover it.


jessdtate (approved):

dcwhite:
I wanted to test this one locally before merging, actually.

PR #2660, opened 2026-08-05
Replace deprecated QCheckBox::stateChanged with toggled

dcwhite:
Fixes the Qt 6.9 deprecation warning:

```
ViewSceneControlsDock.cc:595:56: warning: 'stateChanged' is deprecated: Use checkStateChanged() instead [-Wdeprecated-declarations]
```

`checkStateChanged()` is Qt 6.7+ only and SCIRun still supports Qt 5.15, so this uses `QAbstractButton::toggled` instead — available and non-deprecated in both. All 33 `&QCheckBox::stateChanged` connects in `src/` are converted, not just the one that produced the warning.

## What changed beyond the connects

`toggled` carries a `bool` rather than a `Qt::CheckState`, so every receiving slot that took `int state` is converted to `bool state` and its body simplified (`state != 0` → `state`, `0 == state` → `!state`). Direct callers that passed `mode ? 1 : 0` now pass the bool. Zero-arg targets (`toggleSpinBoxes`, `CheckBoxSlotManager::push`, `NetworkEditor::snapToModules`, …) needed no change.

Converted signatures: 8 slots in `PreferencesWindow`; `SCIRunMainWindow::setSaveBeforeExecute` / `adjustModuleDock`; both `setShowPrefs` overloads in `MainWindowCollaborators`; `TriggeredEventsWindow::enableStateChanged`; the `NetworkEditor::highlightPorts` signal plus its `ModuleProxyWidget::highlightPorts` slot; 2 in `ElectrodeCoilSetupDialog`; both `pushMatrixToState`; and 6 across the Fields dialogs.

None of the checkboxes involved are tristate, so emission behavior is unchanged.

## One real correctness item

`TriggeredEventsWindow::enableStateChanged` compared `state == Qt::Checked` (i.e. `== 2`). Leaving the `int` signature and relying on the implicit `bool`→`int` conversion would have compiled fine and silently made that comparison always false. Converting the signature to `bool` is what makes this safe, which is why the sweep changes signatures rather than only the connect expressions.

## Also included

The last two string-based `SIGNAL()`/`SLOT()` connects in the tree ([ShowUncertaintyGlyphsDialog.cc](https://github.com/SCIInstitute/SCIRun/blob/master/src/Interface/Modules/Visualization/ShowUncertaintyGlyphsDialog.cc)) are converted to the function-pointer form. `src/` now has zero of them. These were already correct at runtime; the change buys compile-time checking. Widget types were confirmed from the `.ui` (`QSlider`, `QDoubleSpinBox`), so no `qOverload` is needed and Qt 5/6 both resolve unambiguously.

## Verification

All 23 touched translation units pass `-fsyntax-only` against a real Qt 6.10.2 build configuration (reusing an existing `compile_commands.json`). No `SIGNAL(`/`SLOT(` or `on_<widget>_stateChanged` auto-connect slots reference any of the renamed members.

This is a syntax check, not a full build or run — moc regeneration, linking, and GUI behavior are not covered here and should come from CI.


jessdtate:
This one touches so many layers of the code and I'd want to see more successful tests before approving.  since it is related to the Qt/UI layer, it may be helpful to see cross platform tests too.   

Is this fix causing discrepancies between qt5 and qt6?

Issue #694, opened 2014-11-17
ViewScene: Keyboard shortcuts


dcwhite:
Mimic v4, make a helper window with list.

<img width="474" alt="screenshot 2015-07-08 12 04 05" src="https://cloud.githubusercontent.com/assets/1399383/8578044/86321d32-2569-11e5-98ca-43fb33d4d917.png">

- [ ] For existing actions, add keyboard shortcut in Designer.
- [ ] Figure out where new actions can live, then add keyboard shortcuts in code.
- [ ] For unimplemented features, add grayed out help line on the help widget (don't start implementing said features).

dcwhite:
@jessdtate IBBM

dcwhite:
I'll make the GUI for this, but the hookup will need to wait until after the SRInterface refactoring.

dcwhite:
This one should be upcoming.

dcwhite:
@RubioJr9 A good one to do, but wait until current PRs are merged before starting. There's just too much to worry about conflicting with right now.

dcwhite:
Fine, I'll just do this one for free

PR #2661, opened 2026-08-06
Fix ViewScene keyboard shortcut follow-up bugs from Mac testing

### LINKED ISSUE #2618: ViewScene keyboard shortcuts: follow-up bugs from Mac testing
Follow-up to #694 / #2513. @jessdtate found these while testing the ViewScene keyboard shortcuts locally on macOS — original report: https://github.com/SCIInstitute/SCIRun/pull/2513#issuecomment-5184573265

I traced each one to its cause on `viewscene-shortcuts`; details below.

---

## 1. Autoview (no scale) does the scaling

`Alt+0` is wired to the same action as plain `0`. From the shortcut table in `ViewScene.cc`:

```cpp
{ Id::AutoviewNoScale, Qt::Key_0, Qt::AltModifier,
  "Autoview (no scale)", "Alt+0", "Reset the eye so the data is centered",
  // TODO: add a spire->centerView() that moves lookAt to bbox center
  // without changing zoom; for now doAutoView is a reasonable stand-in.
  [](ViewSceneDialog* d) { d->autoViewClicked(); } },
```

It's a knowing stand-in, but the help dialog advertises "Reset the eye so the data is centered" and lists it as implemented, so it reads as a bug to anyone using it.

**Fix:** add a `SRInterface`/spire `centerView()` that moves `lookAt` to the scene bbox center while leaving camera distance alone, and point the lambda at it. Until that exists, the row should probably be `nullptr` so it grays out in the help dialog like the other unimplemented entries.

## 2. `Alt` instead of `Cmd` on macOS

The table really does say `Qt::AltModifier` (Option on macOS), and the PR description's `Ctrl+0` is stale. It moved in 14548ced / 6c4d2383: `Set Home` had to leave `Ctrl+H` because Qt maps `Qt::ControlModifier` to Cmd on macOS and **Cmd+H is hide-window**, then Autoview-no-scale was switched to Alt "for consistency".

But Cmd+0 has no macOS conflict — only `Set Home` needed to move. And the table is inconsistent anyway, since `Id::CopyView` still uses `Qt::ControlModifier` / `"Ctrl+1-9"`.

**Fix:** put Autoview (no scale) back on `Qt::ControlModifier` (= Cmd on macOS, Ctrl elsewhere, which Qt handles for free) and update `shortcutDisplay`. Worth deciding at the same time whether `Set Home` stays on Alt+H everywhere or goes platform-conditional, and whether the help dialog should render `Ctrl` as `⌘` on macOS.

## 3. Shortcuts change the state but not the UI

Confirmed for `C`, `D`, `K` — and **`O` (orientation icon) has the same defect**, it just wasn't hit during testing.

**Root cause:** these slots are the *downstream receivers* of the control-dock widget signals. When the user clicks the widget, the widget already holds the new value and the slot only propagates it onward. Driving the slot directly from a shortcut skips the widget entirely — and in three of the four cases it also skips extra work that lives in the connect lambda:

| Key | Shortcut calls | Widget (signal) | Also skipped |
|---|---|---|---|
| `C` | `setClippingPlaneVisible` | `planeVisibleCheckBox_` (`toggled`) | `updateToolbarButton("lightGray")` |
| `D` | `setFogOn` | `fogGroupBox_` (`clicked`) | — |
| `K` | `toggleLight` | `lightCheckBox_` (`toggled`) | `updateLightColor()` |
| `O` | `showOrientationChecked` | `orientationCheckableGroupBox_` (`toggled`) | `toggleButton()` |

So the toolbar buttons go stale along with the checkboxes.

`L` (view locking) is the one toggle that works, precisely because `lockAllTriggered()`/`unlockAllTriggered()` set the widgets themselves:

```cpp
impl_->lockRotation_->setChecked(true);
impl_->lockPan_->setChecked(true);
impl_->lockZoom_->setChecked(true);
```

**Fix:** have each shortcut drive the widget rather than the dialog slot. `C`, `K` and `O` are connected on `toggled`, so a plain `setChecked(!isChecked())` on the widget fires the existing lambda and everything stays in sync.

`D` needs care — `fogGroupBox_` is connected on `clicked`, which `setChecked()` does **not** emit. There's already a helper that does it correctly:

```cpp
void FogControls::toggleFog()
{
  auto toggle = !fogGroupBox_->isChecked();
  fogGroupBox_->setChecked(toggle);
  Q_EMIT setFogTo(toggle);
}
```

so `D` should call `FogControls::toggleFog()`. That's arguably the pattern for all four: one entry point per toggle that owns widget sync *and* downstream propagation, with both the shortcut and the widget going through it.

---

### Checklist

- [ ] Implement a real no-scale autoview (`centerView()`), or gray the row out until it exists
- [ ] Move Autoview (no scale) back to `Qt::ControlModifier`; settle the Set Home modifier story on macOS
- [ ] Consider rendering `Ctrl` as `⌘` in the help dialog on macOS
- [ ] `C` / `D` / `K` / `O` sync their checkbox **and** toolbar button
- [ ] Retest on macOS and Windows

dcwhite:
Fixes #2618. Follow-up to #694 / #2513, from @jessdtate's macOS testing.

## 1. Autoview (no scale) no longer scales

`Alt+0` was wired to the same lambda as plain `0`, with a TODO saying so. Added a real one:

- `SRCamera::centerView()` — moves `lookAt` to the scene bbox centre and recomputes the clip planes, leaving `mCamDistance` untouched
- exposed as `RendererInterface::centerView()` / `SRInterface::centerView()`
- `ViewSceneDialog::autoViewNoScaleClicked()` calls it and pushes camera state

## 2. Modifier

Autoview (no scale) is back on `Qt::ControlModifier` (`Ctrl+0`, `⌘0` on macOS). Only `Set Home` ever needed to leave Ctrl — `Cmd+H` is hide-window on macOS — and it stays on `Alt+H` on every platform, now with a comment explaining why. The help dialog renders `Ctrl+` as `⌘` and `Alt+` as `⌥` on macOS.

**Not in the issue:** `Ctrl+0` is already `actionResetNetworkZoom_` ("Reset Network Zoom") in `SCIRunMainWindow.ui`. A main-window `QAction` shortcut beats the dialog's key events whenever the ViewScene is *docked* — the default floating state hides this. Rather than retreat to Alt, `ViewSceneDialog::event()` now accepts `QEvent::ShortcutOverride` for any key the shortcut table actually handles, so the key arrives as an ordinary key press. Same mechanism will protect the future `Ctrl+1-9` Copy View from `Ctrl+9` Center Network View.

## 3. Toggles that changed state but not the UI

`C`, `D`, `K` and `O` called the dialog slots that are the *downstream receivers* of the control-dock widget signals, so the check box — and the toolbar button styled from it — went stale, and the extra work in the connect lambdas (`updateToolbarButton`, `updateLightColor`, `toggleButton`) was skipped entirely.

Each shortcut now drives the widget through a single named entry point per toggle, and the toolbar-button lambdas route through the same entry point:

| Key | Entry point |
|---|---|
| `C` | `ClippingPlaneControls::toggleVisible()` |
| `D` | `FogControls::toggleFog()` (existing; `fogGroupBox_` is connected on `clicked()`, which `setChecked()` does not emit) |
| `K` | `LightControls::setLightOn()` per light |
| `O` | `OrientationAxesControls::toggleOrientation()` |

## Checklist

- [x] Implement a real no-scale autoview (`centerView()`)
- [x] Move Autoview (no scale) back to `Qt::ControlModifier`; Set Home stays on Alt+H everywhere
- [x] Render `Ctrl`/`Alt` as `⌘`/`⌥` in the help dialog on macOS
- [x] `C` / `D` / `K` / `O` sync their checkbox **and** toolbar button
- [ ] Retest on macOS and Windows

## Testing

All four changed translation units pass `-fsyntax-only` against the build's real compile flags. **Not** built and exercised in the GUI — the shortcut behaviour still needs a manual pass on macOS and Windows.


jessdtate:
a couple more quirks in the toggle logic that doesn't seem consistent. For instance, the clipping plane hotkey (C) only toggles the clipping plane that is selected. This is at least consistent with the UI (clicking the clipping plane icon). However, the lighting hotkey (K) toggles all the lights on or off and it is not consistent with features in the UI. Each of the (4) lights are toggled independently through the UI, and only the headlight can be toggled by clicking the UI icons which changes said icon. The other three are toggled in a submenu and the state is not indicated in the UI icon (either with mouse or hotkey changes). Also, the only light that is on by default is the headlight, and when the lighting is toggled off and on, the other three lights are turned on. There is a similar, though less confusing, scenario with the camera lock hotkey (L), in that it only toggles between all off or all on, yet the UI allows for three independent camera locks (zoom, pan, rotate).

All of the three control problems mentioned above are not only hotkey problems, yet I think only the lighting toggle introduces confounding behaviors that are worth fixing.

I think the most complete fix would be for lighting state for all for lights to be saved (which ones are on or off), then toggle between all off and the saved state. This logic (or similar) could be applied to clipping planes and camera lock toggles for come complete behavior.

The most simple fix (I can think of right now) is to only apply the toggle (K) to the head lamp. This fix may not match SR4, but would match the current UI behavior.

dcwhite:
I think the "off = all off, on = restore what was on" should be pretty easy.

dcwhite:
Took the save/restore approach in 4d69206e6.

**Lighting (K)** — `ViewSceneDialog::toggleAllLights()` now saves the four per-light on/off states when it switches everything off, and restores exactly those on the next press. So headlight-only stays headlight-only across an off/on cycle instead of coming back with all four on. If there's nothing meaningful to restore (all-off state was saved, or it's the first press at startup) it falls back to the headlight alone, which matches the default.

It also reads the check boxes now rather than `Parameters::HeadLightOn`, so the hotkey can't drift from the widget state.

**Camera locks (L)** — `toggleAllLocks()` does the same across rotation/pan/zoom. Fallback from a fully-unlocked state is lock-all, i.e. unchanged from today.

Both follow the pattern `toggleAutoRotate()` already uses for the rotate vector.

**Clipping planes (C)** — left as-is, per your note that it's consistent with clicking the clipping-plane icon.

**Not fixed here:** the point about the secondary lights' UI icon not showing state. That one is real but it's toolbar wiring rather than hotkey behavior — lights 1-3 each get a `QPushButton` in `addLightButtons()` that is never added to a toolbar, and the shared secondary-light button has no linked checkable, so nothing styles it. Split out to #2696.

Still needs the manual pass on macOS and Windows — the changed files only have a syntax check against the real compile flags so far.

jessdtate (approved):

Issue #2642, opened 2026-08-05
VTK vendors zlib, freetype and glew that SCIRun also builds — decide share vs vendor once


dcwhite:
Against `refactor_add_vtk`. Related: #2476 (how VTK enters the build), #2627 (why it is currently hard to do anything about).

VTK 9.6 vendors its own copies of several libraries the SCIRun Superbuild already builds. Each `ThirdParty` module has a `VTK_MODULE_USE_EXTERNAL_VTK_<name>` switch, and on the branch today all of them are left at the default — VTK builds and links its own.

This has already been hit once. `Superbuild/VtkExternal.cmake` contains an abandoned attempt to share zlib:

```cmake
# ---- External dependency control ----
#-DVTK_MODULE_USE_EXTERNAL_VTK_zlib:BOOL=ON

# ---- Help VTK locate them (minimal first pass) ----
# headers
#-DZLIB_INCLUDE_DIR:STRING=${ZLIB_SOURCE_DIR}\;${ZLIB_BINARY_DIR}

# library (platform-aware)
#-DZLIB_LIBRARY:FILEPATH=$<$<PLATFORM_ID:Windows>:${ZLIB_BINARY_DIR}/$<CONFIG>/z.lib>$<$<NOT:$<PLATFORM_ID:Windows>>:${ZLIB_BINARY_DIR}/libz.a>
```

Sharing required naming a raw `libz.a` / `z.lib` inside another project's build tree, plus a generator expression to cover the Windows layout difference — and `ZlibExternal.cmake` had to grow `ZLIB_SOURCE_DIR` / `ZLIB_BINARY_DIR` exports purely to make the attempt possible. Reasonable call to abandon it; that is the only shape the current structure allows. See #2627 for why, and note that once externals install properly this becomes `VTK_MODULE_USE_EXTERNAL_VTK_zlib` plus one `ZLIB_ROOT`, platform-independent.

## The actual overlap

Checked against `Superbuild/Superbuild.cmake` rather than the file listing — several `*External.cmake` files in the tree are not registered:

| library | SCIRun builds it? | VTK bundles it? | status |
|---|---|---|---|
| **zlib** | yes, always (`Superbuild.cmake:212`) | yes | duplicated today; sharing attempted and abandoned |
| **freetype** | yes, always (`:215`) | yes | duplicated today, all platforms, unnoticed |
| **glew** | **Windows only** (`:223`, inside `IF(WIN32)`) | yes | duplicated on Windows only |
[...]

Issue #2606, opened 2026-08-01
VTK: flip WITH_VTK default to OFF before merging refactor_add_vtk


dcwhite:
On the `refactor_add_vtk` branch, `WITH_VTK` currently defaults to **ON**:

```cmake
# Superbuild/Superbuild.cmake:114
OPTION(WITH_VTK "build VTK" ON)
```

That makes VTK 9.6.1 (`Superbuild/VtkExternal.cmake`, cloned from `gitlab.kitware.com/vtk/vtk.git`) a mandatory external for every SCIRun build.

Almost certainly a deliberate development convenience while the branch is under active work — filing it so it doesn't ride along into `master` unnoticed.

### Why it matters

- The first build is already ~15 GB and 30–90 minutes. VTK is a heavyweight addition on top of that, paid by everyone including people who never open a VTK module.
- It works directly against #1487 (build time reduction super-issue) and adds to the CI cost that #2520 is trying to reduce.
- Every other optional backend is off by default — compare `OPTION(WITH_OSPRAY "Build Ospray." OFF)` at `Superbuild.cmake:106`.
- #2476 asks specifically for "a clean CMake option to enable/disable building the VTK renderer backend." A default of ON doesn't quite deliver that.

### Suggested resolution

Flip the default to `OFF` before merge:

```cmake
OPTION(WITH_VTK "Build VTK renderer backend." OFF)
```

…and add a CI matrix entry that builds with `-DWITH_VTK=ON` so the path stays covered (this is also an acceptance criterion on #2476).

If the intent is genuinely that VTK becomes a required dependency — which would be a reasonable end state given milestone #101, "Transition Main renderer to VTK" — then that's worth making an explicit, documented decision rather than something inherited from a development default. In that case the build-time impact should be measured and noted against #1487.

Found while analysing #2148 (ShowField text labels) — see [that discussion](https://github.com/SCIInstitute/SCIRun/issues/2148#issuecomment-5149711162) for the wider review of the branch.

Related: #2476, #1487, #2520

dcwhite:
This is done with #2667

PR #2667, opened 2026-08-07
Refactor add vtk

basisunus:
Adding VTK for rendering


Superbuild/Superbuild.cmake
```diff
@@ -138,6 +109,10 @@ OPTION(WITH_OSPRAY "Build Ospray." OFF)
 # Configure data
 OPTION(BUILD_WITH_SCIRUN_DATA "Svn checkout data" OFF)
 
+###########################################
+# Configure vtk
+OPTION(WITH_VTK "build VTK" ON)
```
dcwhite: Decided to keep WITH_X convention, just switch it off by default
```suggestion
OPTION(WITH_VTK "build VTK" OFF)
```

src/Core/Datatypes/CMakeLists.txt
```diff
@@ -85,6 +87,7 @@ TARGET_LINK_LIBRARIES(Core_Datatypes
   Core_Persistent
   Core_Datatypes_Legacy_Base
   Core_Geometry_Primitives
+  VTK::CommonCore
```
dcwhite: will need the WITH_VTK wrapper

dcwhite: Or maybe not, it seems like VtkGeometry is neutral for now

dcwhite (commented):
Reviewed at 5ba5513 (includes `vtk default off` and `added vtk flag for datatype cmake`). Both of those look right — `Core_Datatypes` now guards the sources, headers and the `VTK::CommonCore` link, which was the piece that would have turned CI red the moment the default flipped. There's one more instance of the same pattern, though, and it's the one currently failing CI.

Reviewed with the understanding that this is work-in-progress being landed to get the scaffolding in place, so it's triaged that way: what needs to be true for this to sit on `master` without affecting anyone who isn't working on VTK, versus what's clearly yours to finish on your own schedule.

Nice drive-by fix on the `OsprayViewer.cc` → `OsprayViewer.h` typo in `Modules/Render/CMakeLists.txt`, by the way — that's been wrong on master for a while.

## The one thing still blocking a clean `WITH_VTK=OFF` build

This is what's turning CI red right now — `windows-headless` failed at link, and it passed on master an hour later, so it's this PR rather than pre-existing.

The module factory doesn't know about the flag, so excluding the sources isn't enough. `GENERATE_MODULE_FACTORY_CODE` is ON by default (`Superbuild.cmake:216`), and `Generator::GetListOfModuleDescriptorFiles` walks `Modules/Factory/Config/` with a plain `directory_iterator`, picking up every `*.module` with no flag awareness. It emits `#include <Modules/Visualization/ShowFieldVtk.h>` and `addModuleDesc<ShowFieldVtk>(...)` regardless of the flag. The headers are on disk either way so it compiles, then:

```
error LNK2019: unresolved external symbol
  "public: static struct ...ModuleLookupInfo const
   SCIRun::Modules::Visualization::ShowFieldVtk::staticInfo_"
  referenced in function
  "...ModuleDescriptionLookup::addModuleDesc<class ...ShowFieldVtk>(...)"

error LNK2019: unresolved external symbol
  "public: __cdecl SCIRun::Modules::Render::ViewSceneVtk::ViewSceneVtk(void)"
  referenced in function "...boost::factory<class ...ViewSceneVtk *>..."

error LNK1120: 19 unresolved externals
```

Three separate groups are unresolved, all from the same root cause:

- **Both modules** — `staticInfo_`, constructors, destructors, `execute`, `setStateDefaults`, `asyncExecute`, `portRemovedSlotImpl`, `inputPort0Name` / `outputPort0Name`, and the `Has1InputPort`/`Has1OutputPort<GeometryVtkPortTag>` port-description instantiations.
- **`VtkDataAlgorithm`** — ctor, dtor and `run`, via `boost::factory` in the generated *algorithm* factory, since `ShowFieldVtk.module` names it.
- **The `ALGORITHM_PARAMETER_DECL` names** — `_init_DefaultColorR/G/B/A`, `_init_Radius`, `_init_ModuleID`. Including `VtkDataAlgorithm.h` emits dynamic initializers for those, but the matching `ALGORITHM_PARAMETER_DEF`s live in the excluded `.cc`. Worth knowing because it means the header can't safely be included in that configuration at all, not just that the class can't be constructed.

Worth flagging one Windows-specific wrinkle for the fix: `isImplementationDisabled` shows up unresolved too, even though `DISABLED_WITHOUT_ABOVE_COMPILE_FLAG` defines it inline. Because `SCISHARE` marks the whole class `__declspec(dllimport)` from the factory's point of view, MSVC imports every member including inline ones. So adding the macro alone won't rescue this — the `.cc` genuinely has to be compiled into the DLL.

There's a second consequence beyond the link error: the `#ifndef WITH_VTK` / `error("Must compile WITH_VTK...")` path in `ViewSceneVtk::execute()` and the `DISABLED_WITHOUT_ABOVE_COMPILE_FLAG` in `ViewSceneVtk.h` are both unreachable, because the file isn't compiled in the configuration where they'd fire. `ShowFieldVtk.h` doesn't have the macro at all.

`OsprayViewer` and `InterfaceWithPython` solve both problems the other way round: sources compiled unconditionally, only the *bodies* guarded (`OsprayViewer.cc:151`), plus `DISABLED_WITHOUT_ABOVE_COMPILE_FLAG` on the class. That's exactly what lets a saved network containing those modules load in a build without the dependency and report a clean "rebuild with X" rather than failing.

Suggest matching it: drop the `if (WITH_VTK)` guards around the two module sources and `VtkDataAlgorithm`, guard the VTK-dependent bodies with `#ifdef WITH_VTK`, and add `DISABLED_WITHOUT_ABOVE_COMPILE_FLAG` to `ShowFieldVtk.h`.

## A few unrelated changes to back out

- `src/Externals/submodules/googletest` is bumped `52eb810` → `8dfcda9`. Looks accidental.
- `Superbuild/ZlibExternal.cmake` — every zlib-for-VTK line in `VtkExternal.cmake` is commented out, and `ZLIB_SOURCE_DIR`/`ZLIB_BINARY_DIR` are set non-cache so they wouldn't reach the VTK project anyway. Safe to revert.
- `TextureMan::updateTexture` and `ES/shaders/ViewQuad.vs/.fs` have no callers or references anywhere in the tree — leftovers from the spire-texture approach before you switched to the `QImage` path. Worth stripping now since two of them are edits to vendored spire.

Also still needs a rebase — currently conflicting with master.

## Test network

Could you include the test network you demoed in the meeting? Two things worth checking in with it once the factory issue above is sorted: that it loads and runs with `WITH_VTK=ON`, and that in a default (`OFF`) build it still opens and the modules report themselves as disabled rather than the network failing to load.

## `.module` config fields

Both configs ship the placeholder `"status": "..."` and `"description": "..."`. Worth filling in, because those two strings are emitted verbatim as the arguments to `addModuleDesc<>` and surface in the module list — so right now a half-wired `ShowFieldVtk` shows up described as `...`. The repo already uses `"In progress"` and variants like `"In progress: needs more testing"` for exactly this case. Something like:

`ShowFieldVtk.module`:
```json
  "module": {
    "name": "ShowFieldVtk",
    "namespace": "Visualization",
    "status": "In progress: experimental VTK rendering path",
    "description": "Converts a field into VTK geometry for display in ViewSceneVtk. Requires WITH_VTK.",
    "header": "Modules/Visualization/ShowFieldVtk.h"
  },
```

`ViewSceneVtk.module`:
```json
  "module": {
    "name": "ViewSceneVtk",
    "namespace": "Render",
    "status": "In progress: experimental VTK renderer",
    "description": "VTK-based scene viewer. Requires WITH_VTK; see ViewScene for the production renderer.",
    "header": "Modules/Render/ViewSceneVtk.h"
  },
```

## Two things you'll want to know about (not asks)

Both are VTK behaviors that are easy to lose time to, so flagging them now rather than letting you rediscover them:

1. **`VtkRenderer::createLookupTable` never reads `tfn.colors`.** It sets the range, and only when `tfn.colors` is *empty* sets a hue range — the colors themselves are never used. So all seven `ColorMap_OSP_helper` calls in `VtkGeometryBuilder` are currently feeding something nothing consumes, and the ColorMap input has no visible effect yet.

2. **Scalar visibility overrides `prop->SetColor`.** The builders always attach a `"Values"` point array (zero-filled when the field has no data), so the mapper's LUT wins and the Default Color spinboxes don't affect surfaces or unstructured grids. Looks like a UI bug from the outside but isn't.

## Backlog, whenever you get back to it

Not blocking, and mostly unreachable while the flag is off — just so they're written down somewhere:

- `QImage::flipped()` in `VtkRenderer.cc` is Qt 6.9+. In practice we all build against later Qt so this won't bite you locally, but CI still installs `6.3.*` (`pr.yml` windows-gui-qt6, plus `windows.yml`) and `5.15.2` (the qt5 jobs), so those are the jobs that would break whenever VTK gets switched on in CI. `mirrored(false, true)` is the portable spelling if we don't want to raise the floor first — worth a separate conversation about what the real minimum is, given nobody's actually building at 6.3.
- `addStructVol` is the only builder missing the `vfield->num_values() > 0` guard (two sites), `dx = sizeX / (ni - 1)` divides by zero on a single-slice latvol, and `meshObj` never gets a `tfn`.
- `renderImageData` indexes `tfn.opacities[0]` / `colors[0..2]` guarded only by `range.empty()`.
- `updateGeometries` calls `ResetCamera()` unconditionally, so the camera snaps back on every re-execute — including each spinbox change.
- `sendCompositeGeometry` early-returns on empty input, so disconnecting the last port leaves the previous scene on screen.
- `VtkQWidget` repaints on a fixed 16 ms timer regardless of whether anything changed, and `renderFrame()` re-issues `SetSize` every frame — a permanent 60 fps offscreen render per open window. Cheaper to make this on-demand now than after more is built on top.
- Pre-ship rather than pre-merge: VTK version is pinned twice (`v9.6.1` in `VtkExternal.cmake`, `vtk-9.6` in the `VTK_DIR` path at `src/CMakeLists.txt:310` — candidate for the VERSIONS.cmake work in #2480); `find_package(VTK COMPONENTS ...)` omits `CommonDataModel`, `RenderingAnnotation` and `RenderingVolumeOpenGL2` although all three get linked; only Windows gets runtime-path handling in `Main/CMakeLists.txt`, so mac/linux packaging won't find the VTK dylibs; and the Rendering group brings VTK's vendored freetype alongside `SCI_FREETYPE`, which is the check tracked in #2642.
- `VtkExternal.cmake` has no `GIT_SHALLOW`, so it clones VTK's full history for one tag.

dcwhite:
@basisunus — flagging that the branch has been fully red since the `WITH_VTK` default flip, and it's probably worth pausing feature work to get it green before pushing more. Every push here costs ~15 h of runner time (measurement below), so red pushes are expensive, and after 18647d10b there's no configuration of this branch that builds.

## Current state

Both post-flip SHAs, every job, both platforms:

| SHA | linux-build | windows-build | windows-build-script |
|---|---|---|---|
| 5ba5513ee | ❌ | ❌ | ❌ |
| 18647d10b | ❌ | ❌ | ❌ |
| 897fa3d44 (pre-flip) | ❌ | ❌ | ✅ |

## What's breaking

Flipping the default to OFF was the right call (#2606), but the three code generators scan `src/Modules/Factory/Config/` unconditionally. `ShowFieldVtk.module` and `ViewSceneVtk.module` still get registered when `WITH_VTK=OFF`, while the CMake guards exclude the translation units that implement them. Three separate symptoms, one cause:

- `AlgorithmFactoryImpl_Generated` — 12 undefined: `VtkDataAlgorithm::VtkDataAlgorithm()`, `VtkVisualization::Parameters::_init_*`
- `ModuleFactoryImpl_Generated` — 12 undefined: `ShowFieldVtk` / `ViewSceneVtk` ctors, dtors, `staticInfo_`
- `ModuleDialogFactory_Generated` — fatal: `Interface/Modules/Visualization/ui_ShowFieldVtk.h` not found, because the `.ui` never reaches `uic` when the guard is on

OSPRAY never hit this only because `OsprayDataAlgorithm.cc` compiles unconditionally.

## Suggested fix

The intent is that every module is always built and always registered, with the internals compiled out — so a network containing `ShowFieldVtk` still *loads* in a non-VTK build, the module just doesn't work. That's the existing `DISABLED_WITHOUT_ABOVE_COMPILE_FLAG` pattern, and `InterfaceWithTetGen` is the reference implementation:

- `InterfaceWithTetGen.cc/.h` are in the **unconditional** source list
- only `InterfaceWithTetGenImpl.cc/.h` sit behind `IF(WITH_TETGEN)`
- the header carries `#ifndef WITH_TETGEN → DISABLED_WITHOUT_ABOVE_COMPILE_FLAG`
- `execute()` has an `#else error("This module needs the build flag WITH_TETGEN enabled in order to work.")`
- `ModuleProxyWidget` then greys the module out and warns the user at load time

You've already done exactly this for `ViewSceneVtk` — `ViewSceneVtk.h:98` has the macro and `ViewSceneVtk.cc:114` has the `execute()` guard. The problem is that the sources are *also* excluded in CMake (`src/Modules/Render/CMakeLists.txt:40`, `src/Modules/Visualization/CMakeLists.txt:66,93`, `src/Core/Algorithms/Visualization/CMakeLists.txt:37`), which defeats the pattern — the factory references a class that was never compiled.

So roughly:

1. Move `ViewSceneVtk.*`, `ShowFieldVtk.*`, `VtkDataAlgorithm.*` and the two dialogs back into the unconditional source lists, and keep only the genuinely VTK-touching code (`VtkGeometryBuilder`, `Interface/Modules/Render/Vtk/*`, the `VTK::` link lines) behind `IF(WITH_VTK)` — the Tetgen `*Impl` split.
2. Both `.ui` files always go through `uic`; that costs nothing and needs no VTK.
3. `ShowFieldVtk` needs the pattern added — it has no guard at all today.
4. `ALGORITHM_PARAMETER_DECL` definitions in `VtkDataAlgorithm.cc` must compile unconditionally; they're just `Name` objects and don't pull in VTK headers.

## Why it's worth doing now rather than later

Measured across every green run on this branch against a time-matched master baseline (both pre-#2619 serial, so it's apples to apples): VTK-enabled jobs run **1.90×** master, **+911 min ≈ 15.2 h of runner time per push**. Log parsing attributes essentially all of it to the `VTK_external` build at a flat ~111 min per job, near-identical on Linux and Windows; SCIRun's own compile is unchanged. Full write-up and the CI plan are in #2677.

Two related things from that issue, so they're on your radar: `VtkExternal.cmake` already has a real install step, which means VTK can be `actions/cache`'d today without waiting on #2627 — that turns the 111 min into a download. And this PR currently has **zero macOS coverage** (the branch has no `pr.yml` and inherits master's `mac.yml`, which no longer triggers on pull requests), so the VTK build has never been exercised on a Mac.

To be explicit about the rule this follows, since it's the difference between "un-gate the sources" and "teach the factory to skip modules": **every module is always built and always registered regardless of build flags.** Optional-dependency modules compile their internals out and report themselves disabled, so an existing network always loads and the affected modules simply don't run. That's already what Tetgen, OSPRay and the Python modules do.

---

@jessdtate — I didn't raise this at yesterday's meeting, so flagging it here. Can you confirm the priority call? My read is that this branch should go green before any more feature commits land on it, and that Yong should treat the build fix as the next thing rather than continuing on renderer work.

Arguing for that: CI coverage of this branch has been thin for a while and is now zero. `linux-build` and `windows-build` last went green on 2026-07-14; `windows-build-script` carried the branch on its own after that, and the flip took that one down too. So Yong's local `WITH_VTK=ON` build is currently the only thing compiling this code at all — which is real coverage, but it's one configuration on one machine. On top of that, each push burns ~15 h of runner time to produce red, and the longer the OFF path stays broken the more code gets written against an ON-only assumption, which is how it broke in the first place.

Arguing against, and the reason I'm asking rather than just saying it: it's a few hours of CMake work with no renderer value, milestone "VTK renderer" is due 2026-08-31, and if you'd rather Yong keep momentum on the renderer and have someone else take the build fix, that's an entirely reasonable call — I'm happy to pick it up.

dcwhite:
image
From today's meeting. 

@basisunus wants to have nodes and edges working with the colormap input and then we can checkpoint and merge this PR.

dcwhite:
Could you commit the `.srn5` test network you've been using to drive `ViewSceneVtk` / `ShowFieldVtk`? The branch doesn't add one, so reviewing means rebuilding it by hand.

dcwhite:
Two test networks for the VTK render path, in case they're useful alongside your own. Both use empty `<stateMap>` blocks so every module takes its `setStateDefaults()` values — no data files or external resources needed.

**`vtk-latvol-basic.srn5`** — smoke test:

```
CreateLatVol:0 ──OutputField──▶ ShowFieldVtk:0 ──SceneGraph──▶ ViewSceneVtk:0 (GeneralGeom)
```

**`vtk-vs-classic-compare.srn5`** — the same `CreateLatVol` fanned out to both pipelines, so the VTK renderer and the production one can be compared on identical geometry:

```
                    ┌─▶ ShowField:0    ──▶ ViewScene:0
CreateLatVol:0 ─────┤
                    └─▶ ShowFieldVtk:0 ──▶ ViewSceneVtk:0
```

Verified against a local Windows build (VS 2019, Qt 5.15.2, `WITH_VTK=ON`) on this branch: both archives parse and all modules instantiate with the expected IDs. Caveat worth stating plainly — I could not confirm the **connections** at runtime headlessly (`scirun_get_module_input_object` returns `None` for every port), so the wiring is structurally correct but unproven until someone opens these in the GUI. Port names and IDs were taken from the `INPUT_PORT`/`OUTPUT_PORT` macros and cross-checked against existing nets.

<details>
<summary><code>vtk-latvol-basic.srn5</code></summary>

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
<!DOCTYPE boost_serialization>
<boost_serialization signature="serialization::archive" version="12">
<networkFile class_id="0" tracking_level="0" version="3">
	<networkInfo class_id="1" tracking_level="0" version="0">
		<modules class_id="2" tracking_level="0" version="0">
			<count>3</count>
			<item_version>0</item_version>
			<item class_id="3" tracking_level="0" version="0">
				<first>CreateLatVol:0</first>
				<second class_id="4" tracking_level="0" version="0">
					<module class_id="5" tracking_level="0" version="0">
						<package_name_>SCIRun</package_name_>
						<category_name_>NewField</category_name_>
						<module_name_>CreateLatVol</module_name_>
					</module>
					<state class_id="6" tracking_level="0" version="0">
						<stateMap class_id="7" tracking_level="0" version="0">
							<count>0</count>
							<item_version>0</item_version>
						</stateMap>
					</state>
				</second>
			</item>
			<item>
				<first>ShowFieldVtk:0</first>
				<second>
					<module>
						<package_name_>SCIRun</package_name_>
						<category_name_>Visualization</category_name_>
						<module_name_>ShowFieldVtk</module_name_>
					</module>
					<state>
						<stateMap>
							<count>0</count>
							<item_version>0</item_version>
						</stateMap>
					</state>
				</second>
			</item>
			<item>
				<first>ViewSceneVtk:0</first>
				<second>
					<module>
						<package_name_>SCIRun</package_name_>
						<category_name_>Render</category_name_>
						<module_name_>ViewSceneVtk</module_name_>
					</module>
					<state>
						<stateMap>
							<count>0</count>
							<item_version>0</item_version>
						</stateMap>
					</state>
				</second>
			</item>
		</modules>
		<connections class_id="8" tracking_level="0" version="0">
			<count>2</count>
			<item_version>0</item_version>
			<item class_id="9" tracking_level="0" version="0">
				<moduleId1_>CreateLatVol:0</moduleId1_>
				<port1_ class_id="10" tracking_level="0" version="0">
					<name>OutputField</name>
					<id>0</id>
				</port1_>
				<moduleId2_>ShowFieldVtk:0</moduleId2_>
				<port2_>
					<name>Field</name>
					<id>0</id>
				</port2_>
			</item>
			<item>
				<moduleId1_>ShowFieldVtk:0</moduleId1_>
				<port1_>
					<name>SceneGraph</name>
					<id>0</id>
				</port1_>
				<moduleId2_>ViewSceneVtk:0</moduleId2_>
				<port2_>
					<name>GeneralGeom</name>
					<id>0</id>
				</port2_>
			</item>
		</connections>
	</networkInfo>
	<modulePositions class_id="11" tracking_level="0" version="0">
		<count>3</count>
		<item_version>0</item_version>
		<item class_id="12" tracking_level="0" version="0">
			<first>CreateLatVol:0</first>
			<second class_id="13" tracking_level="0" version="0">
				<first>-2.00000000000000000e+002</first>
				<second>-1.50000000000000000e+002</second>
			</second>
		</item>
		<item>
			<first>ShowFieldVtk:0</first>
			<second>
				<first>-2.00000000000000000e+002</first>
				<second>0.00000000000000000e+000</second>
			</second>
		</item>
		<item>
			<first>ViewSceneVtk:0</first>
			<second>
				<first>-2.00000000000000000e+002</first>
				<second>1.50000000000000000e+002</second>
			</second>
		</item>
	</modulePositions>
	<moduleNotes class_id="14" tracking_level="0" version="0">
		<count>0</count>
		<item_version>0</item_version>
	</moduleNotes>
	<connectionNotes>
		<count>0</count>
		<item_version>0</item_version>
	</connectionNotes>
	<moduleTags class_id="15" tracking_level="0" version="0">
		<count>3</count>
		<item_version>0</item_version>
		<item class_id="16" tracking_level="0" version="0">
			<first>CreateLatVol:0</first>
			<second>0</second>
		</item>
		<item>
			<first>ShowFieldVtk:0</first>
			<second>0</second>
		</item>
		<item>
			<first>ViewSceneVtk:0</first>
			<second>0</second>
		</item>
	</moduleTags>
</networkFile>
</boost_serialization>
```

</details>

<details>
<summary><code>vtk-vs-classic-compare.srn5</code></summary>

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
<!DOCTYPE boost_serialization>
<boost_serialization signature="serialization::archive" version="12">
<networkFile class_id="0" tracking_level="0" version="3">
	<networkInfo class_id="1" tracking_level="0" version="0">
		<modules class_id="2" tracking_level="0" version="0">
			<count>5</count>
			<item_version>0</item_version>
			<item class_id="3" tracking_level="0" version="0">
				<first>CreateLatVol:0</first>
				<second class_id="4" tracking_level="0" version="0">
					<module class_id="5" tracking_level="0" version="0">
						<package_name_>SCIRun</package_name_>
						<category_name_>NewField</category_name_>
						<module_name_>CreateLatVol</module_name_>
					</module>
					<state class_id="6" tracking_level="0" version="0">
						<stateMap class_id="7" tracking_level="0" version="0">
							<count>0</count>
							<item_version>0</item_version>
						</stateMap>
					</state>
				</second>
			</item>
			<item>
				<first>ShowField:0</first>
				<second>
					<module>
						<package_name_>SCIRun</package_name_>
						<category_name_>Visualization</category_name_>
						<module_name_>ShowField</module_name_>
					</module>
					<state>
						<stateMap>
							<count>0</count>
							<item_version>0</item_version>
						</stateMap>
					</state>
				</second>
			</item>
			<item>
				<first>ShowFieldVtk:0</first>
				<second>
					<module>
						<package_name_>SCIRun</package_name_>
						<category_name_>Visualization</category_name_>
						<module_name_>ShowFieldVtk</module_name_>
					</module>
					<state>
						<stateMap>
							<count>0</count>
							<item_version>0</item_version>
						</stateMap>
					</state>
				</second>
			</item>
			<item>
				<first>ViewScene:0</first>
				<second>
					<module>
						<package_name_>SCIRun</package_name_>
						<category_name_>Render</category_name_>
						<module_name_>ViewScene</module_name_>
					</module>
					<state>
						<stateMap>
							<count>0</count>
							<item_version>0</item_version>
						</stateMap>
					</state>
				</second>
			</item>
			<item>
				<first>ViewSceneVtk:0</first>
				<second>
					<module>
						<package_name_>SCIRun</package_name_>
						<category_name_>Render</category_name_>
						<module_name_>ViewSceneVtk</module_name_>
					</module>
					<state>
						<stateMap>
							<count>0</count>
							<item_version>0</item_version>
						</stateMap>
					</state>
				</second>
			</item>
		</modules>
		<connections class_id="8" tracking_level="0" version="0">
			<count>4</count>
			<item_version>0</item_version>
			<item class_id="9" tracking_level="0" version="0">
				<moduleId1_>CreateLatVol:0</moduleId1_>
				<port1_ class_id="10" tracking_level="0" version="0">
					<name>OutputField</name>
					<id>0</id>
				</port1_>
				<moduleId2_>ShowField:0</moduleId2_>
				<port2_>
					<name>Field</name>
					<id>0</id>
				</port2_>
			</item>
			<item>
				<moduleId1_>CreateLatVol:0</moduleId1_>
				<port1_>
					<name>OutputField</name>
					<id>0</id>
				</port1_>
				<moduleId2_>ShowFieldVtk:0</moduleId2_>
				<port2_>
					<name>Field</name>
					<id>0</id>
				</port2_>
			</item>
			<item>
				<moduleId1_>ShowField:0</moduleId1_>
				<port1_>
					<name>SceneGraph</name>
					<id>0</id>
				</port1_>
				<moduleId2_>ViewScene:0</moduleId2_>
				<port2_>
					<name>GeneralGeom</name>
					<id>0</id>
				</port2_>
			</item>
			<item>
				<moduleId1_>ShowFieldVtk:0</moduleId1_>
				<port1_>
					<name>SceneGraph</name>
					<id>0</id>
				</port1_>
				<moduleId2_>ViewSceneVtk:0</moduleId2_>
				<port2_>
					<name>GeneralGeom</name>
					<id>0</id>
				</port2_>
			</item>
		</connections>
	</networkInfo>
	<modulePositions class_id="11" tracking_level="0" version="0">
		<count>5</count>
		<item_version>0</item_version>
		<item class_id="12" tracking_level="0" version="0">
			<first>CreateLatVol:0</first>
			<second class_id="13" tracking_level="0" version="0">
				<first>0.00000000000000000e+000</first>
				<second>-2.00000000000000000e+002</second>
			</second>
		</item>
		<item>
			<first>ShowField:0</first>
			<second>
				<first>-2.50000000000000000e+002</first>
				<second>0.00000000000000000e+000</second>
			</second>
		</item>
		<item>
			<first>ShowFieldVtk:0</first>
			<second>
				<first>2.50000000000000000e+002</first>
				<second>0.00000000000000000e+000</second>
			</second>
		</item>
		<item>
			<first>ViewScene:0</first>
			<second>
				<first>-2.50000000000000000e+002</first>
				<second>2.00000000000000000e+002</second>
			</second>
		</item>
		<item>
			<first>ViewSceneVtk:0</first>
			<second>
				<first>2.50000000000000000e+002</first>
				<second>2.00000000000000000e+002</second>
			</second>
		</item>
	</modulePositions>
	<moduleNotes class_id="14" tracking_level="0" version="0">
		<count>0</count>
		<item_version>0</item_version>
	</moduleNotes>
	<connectionNotes>
		<count>0</count>
		<item_version>0</item_version>
	</connectionNotes>
	<moduleTags class_id="15" tracking_level="0" version="0">
		<count>5</count>
		<item_version>0</item_version>
		<item class_id="16" tracking_level="0" version="0">
			<first>CreateLatVol:0</first>
			<second>0</second>
		</item>
		<item>
			<first>ShowField:0</first>
			<second>0</second>
		</item>
		<item>
			<first>ShowFieldVtk:0</first>
			<second>0</second>
		</item>
		<item>
			<first>ViewScene:0</first>
			<second>0</second>
		</item>
		<item>
			<first>ViewSceneVtk:0</first>
			<second>0</second>
		</item>
	</moduleTags>
</networkFile>
</boost_serialization>
```

</details>

dcwhite:
Heads up on the test environment — have you built this branch with `-DBUILD_TESTING=ON`? On a Windows build (VS 2019, Qt 5.15.2, `WITH_VTK=ON`) **777 of 794 tests fail**, all with exit code `0xc0000135` (`STATUS_DLL_NOT_FOUND`) before reaching `main()`.

It's a one-line environment gap, not anything wrong with the renderer code:

1. This branch puts `VtkGeometryObject` in `Core/Datatypes`, so `Core_Datatypes.dll` now imports `vtkCommonCore-9.6.dll` and `vtksys-9.6.dll`.
2. `src/Testing/CMakeLists.txt` (~line 67) builds an explicit `_test_path` containing Python + Qt + Qwt + the configure-time `$ENV{PATH}`, and sets it as each test's `ENVIRONMENT` property.
3. That property **replaces** `PATH` rather than extending it, so adding the VTK bin directory to the ambient environment before running `ctest` has no effect.
4. Nearly every test links `Core_Datatypes`, so nearly every test dies at load time.

Confirmed by copying `Externals/Install/VTK_external/bin/*.dll` next to the test binaries in `SCIRun/Release/` — the suite then runs normally.

The proper fix looks like appending the VTK install `bin` to `_test_path`, guarded by `WITH_VTK`, next to the existing Qwt entries. Worth doing before `WITH_VTK` is turned on for Windows CI, or the whole suite goes red at once.

Two pre-existing oddities in that same block, neither related to this failure but visible while tracing it:

- `${Qwt_DIR}/src/Debug` and `${Qwt_DIR}/src/Release` expand to `/src/Debug` and `/src/Release` — `Qwt_DIR` is empty. Harmless today since Qwt builds as a static `qwt.lib`, but it's dead weight in every test's PATH.
- `EXE_DIRPATH` (line 130, `ADD_TEST("${test}" "../${EXE_DIRPATH}/${test}")`) is never defined anywhere in the repo. Also harmless, since CTest substitutes the config directory for relative test commands under `-C Release`.

dcwhite:
One structural request, and I'd like to see it land as part of this PR rather than as follow-up: **move `VtkGeometry.{h,cc}` out of `Core_Datatypes` into a separate, conditionally-built library.**

The reason is the link edge this currently creates:

```cmake
IF (WITH_VTK)
  TARGET_LINK_LIBRARIES(Core_Datatypes
    VTK::CommonCore
  )
ENDIF()
```

`Core_Datatypes` is the foundation nearly every target in the tree links, so this makes VTK a de facto universal dependency — for everything, whether or not it touches VTK. That is what produced the 777-test failure in the comment above, and it will show up again in deployment (the VTK runtime becomes required by anything that touches a SCIRun datatype) and in build times, since `vtkSmartPointer.h` and `vtkDataObject.h` now leak into every TU that includes the header.

Only one member actually needs VTK:

```cpp
#ifdef WITH_VTK
    vtkSmartPointer<vtkDataObject> dataObject;
#endif
```

Everything else in `VtkGeometryObject` — `Material`, `TransferFunc`, `GeometryType`, `BBox`, `id`, `version` — is SCIRun-native.

I'd keep the `vtkSmartPointer` rather than flattening the payload to plain arrays. Holding the `vtkDataObject` is the point: `ShowFieldVtk` builds it once and the renderer consumes it directly, with no conversion step. Flattening would reintroduce exactly the cost this path is well placed to avoid.

**The split looks cheap.** I traced the coupling on this branch:

- `src/Dataflow/` contains **zero** references to `VtkGeometry` — the engine only ever sees the forward declaration already present in `DatatypeFwd.h`, which is all the port machinery needs since ports traffic in the `Datatype` base and handles.
- No references in `src/Modules/Factory/` either, so no factory or serialization path needs the complete type.
- Only five files include `Core/Datatypes/VtkGeometry.h`: `VtkGeometryBuilder.h`, `VtkRenderer.h`, `ViewSceneVtk.cc`, `ShowFieldVtk.cc`, and its own `.cc` — all already VTK-aware and already linking VTK.

So it should be a new library target, a moved pair of files, and a handful of link lines.

A bonus worth having: with the library built only when `WITH_VTK`, the `#ifdef` inside the class can go away entirely. As written, `sizeof(VtkGeometryObject)` depends on a preprocessor flag. `add_definitions(-DWITH_VTK)` is global in `src/CMakeLists.txt` so it is consistent within a build today, but a conditionally-built library removes the class from existence rather than silently changing its layout — which is the safer property to rely on.

Unrelated but spotted while reading the class:

```cpp
VtkGeometryObject* VtkGeometryObject::clone() const
{
    return nullptr;  // TODO
}
```

A `Datatype` subclass whose `clone()` returns null will surface as a null dereference somewhere far from the cause if anything in the engine copies port data. Worth either implementing it or making it throw, so it fails where the mistake is.

dcwhite:
The two generated networks above work on my Win11 machine. I am generating a few more to test out all the SFV flags, colormapping, and different field types (i do not expect them all to pass)

dcwhite:
Interesting usablilty bug: with all options turned off in ShowFieldVTK, a latvol input errors with "Field type not supported". It should probably just output an empty geometry.

dcwhite:
[vtk-showfield-flag-matrix.zip](https://github.com/user-attachments/files/31847287/vtk-showfield-flag-matrix.zip)
Here's the test network for testing all flag combos

dcwhite:
There's also some custom code that turns off the module dialog's execute buttons for ViewScene. VSvtk should have this as well.

dcwhite:
[vtk-field-types.zip](https://github.com/user-attachments/files/31849327/vtk-field-types.zip)
more networks to commit

dcwhite (approved):

dcwhite:
Oops, the windows build break is from @jessdtate, his didn't run that job and I shouldve run it. I will fix ASAP

dcwhite:
I built it on Windows myself, and it was fine--weird

PR #2603, opened 2026-07-31
[2/3] C++20 module boilerplate: variadic ports + unified MODULE_TRAITS_AND_INFO

dcwhite:
## Summary

Modernizes the SCIRun module **boilerplate** with C++20, on top of the `.module`
registration standardization. Introduces variadic input/output port declarations
via C++20 string non-type template params, compile-time port lookup, and a single
4-arg `MODULE_TRAITS_AND_INFO` that replaces `MODULE_INFO_DEF`.

## What changes (the C++20 layer, on top of `module-config-cleanup`)

- Variadic `HasInputPorts` / `HasOutputPorts` via C++20 string NTTPs.
- Compile-time port lookup + `TagDataType` trait + `getRequiredInput_` macro;
  `sendOutput_` macro.
- Bumps the C++ standard to **20**; `fmt::runtime` in logs.
- 4-arg `MODULE_TRAITS_AND_INFO`, eliminating `MODULE_INFO_DEF`.
- Reference migration: ShowString fully converted to variadic ports + the new
  macros, with a `showStringCpp20.srn5` demo network.
- Delta over the base branch: **9 commits, 11 files (+1982/−221)**. (The rest of
  this branch's diff vs master is inherited from `module-config-cleanup`.)

## Why

Collapses the repetitive `INPUT_PORT`/`OUTPUT_PORT` + `MODULE_INFO_DEF` ceremony
into compile-time-checked variadic declarations — less per-module boilerplate,
earlier (compile-time) errors, and a uniform module shape that tooling can
generate.

## Relationship / merge order

- **Stacked on `module-config-cleanup`** — this branch *contains* all of that
  branch's commits. This PR targets `module-config-cleanup` as its base so the
  diff shows only the C++20 layer; GitHub will auto-retarget it to `master` once
  the base PR merges.
- **Merge order:** base branch first, then this one.
- **Blocks:** #2596, #2595 — the conversion automation targets these final macros.
- **Blocked by:** #101 / `module-config-cleanup`.

## Status

Draft for visibility. `module-config-cleanup` (which carries master) is merged in,
so this branch is up to date with master and the stacked diff is a clean 11-file
C++20 layer. CI runs on the merged state.

Refs #101, #2595

PR #2678, opened 2026-08-09
CI: narrow ci-gate to two Linux jobs on non-platform PRs

dcwhite:
## What

`mac-gui` and `windows-gui-qt6` now run on a PR only when it touches build-system, CI, or platform-specific paths. Ordinary PRs gate on `linux-gui-qt6` + `linux-unit-tests`.

```
ordinary PR:   linux-gui-qt6 (~63) + linux-unit-tests (~50)   -> ci-gate ~63 min
platform PR:   + mac-gui (~43) + windows-gui-qt6 (104-139)    -> ci-gate ~139 min, i.e. today
```

Both jobs stay in `ci-gate`'s `needs`. When they run, the gate waits on them exactly as it does now. When they're skipped, branch protection counts that as passing and the gate falls through to the two Linux jobs. Risky PRs keep the full guarantee automatically — no label to remember, no second advisory context.

`ci-gate` keeps its name and stays the sole required context, so **no branch-protection change is needed**, and this is one commit to revert.

## Why

`ci-gate` is a strict required check (`required_status_checks.strict: true`), so every merge to `master` invalidates every other open PR and re-runs everything the gate waits on. What it waits on is dominated by `windows-gui-qt6` at 104–139 min, with `macOS-latest` capped at 5 concurrent jobs repo-wide behind it — run 31078669153 queued 4h05m for a mac runner and cancelled the two runs before it.

This wasn't possible until #2666. There was no Qt6 build anywhere outside mac and windows, and `linux-unit-tests` is `variant: headless`, which skips `src/Interface` entirely — so no combination of the pre-existing Linux jobs could stand in for them.

`linux-gui-qt6` now has **8 green runs across 5 branches, 58–63 min**, including the Qt6 deprecation PRs (#2659, #2660) that are precisely what this gate has to protect.

## The path filter

Selects on `Superbuild/`, `.github/`, `CMake/`, `src/Externals/`, `src/Main/`, `*.cmake`, `build.sh|ps1`, and the two **top-level** `CMakeLists.txt`.

Nested `CMakeLists.txt` are deliberately excluded. An earlier draft matched all of them and classified 13 of 18 open PRs as platform-risky, which is useless — adding a module means editing a source list, and that says nothing about platform behaviour. The two top-level files are the ones holding the `APPLE`/`WIN32` blocks.

Against the 18 open PRs the narrowed filter gives **8 full-matrix / 10 Linux-only**, each selection with a visible cause:

| PR | trigger |
|---|---|
| #2667 | `Superbuild/Superbuild.cmake` |
| #2644, #2576, #2575 | `.github/workflows/reusable-build.yml` |
| #2603, #2602 | `src/CMakeLists.txt` |
| #2658 | `.github/workflows/nightly-slack.yml` |
| #2643 | `.github/workflows/pr.yml` |

The 10 skipped are pure C++/Qt/module work.

## What this accepts

**MSVC is the strictest of the three compilers** — missing includes and two-phase name lookup pass gcc and fail cl. That's the real exposure. Mitigations: Yong develops on Windows day to day, `windows.yml` runs the full matrix on every `master` push, and the nightly runs it again.

**macOS** is well covered — daily local builds plus `mac.yml` on master push.

Honest about the ceiling: most of the last 20 *merged* PRs would still take the full matrix, because recent merge history is mostly CI and build-system work. The savings land on module and GUI PRs, which is what the open queue currently is.

Any PR can still be given the full matrix on demand via `workflow_dispatch` on `mac.yml` / `windows.yml`, or by pushing a trivial touch to a filtered path.

Follow-up to #2666.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -16,6 +16,11 @@ name: pr
 #     contexts directly re-creates the problem in rule 1 as soon as one of them
 #     is filtered or renamed.
 #
+# Within this workflow there are two tiers. Two Linux jobs -- linux-gui-qt6 and
+# linux-unit-tests -- run on every non-docs PR and are what ci-gate normally
+# waits on. mac-gui and windows-gui-qt6 join them only when the PR touches
+# build-system, CI, or platform-specific paths (see `platform_risk` below).
+#
 # The full matrix (all the mac-headless-*, windows-gui-python-*, regression and
 # coverage variants) lives in ccpp.yml / mac.yml / windows.yml /
 # regression-tests.yml / windows-build-script.yml and runs on master pushes and
@@ -38,6 +43,7 @@ jobs:
       pull-requests: read
     outputs:
       docs_only: ${{ steps.filter.outputs.docs_only }}
+      platform_risk: ${{ steps.filter.outputs.platform_risk }}
     steps:
       - name: Classify changed paths
         id: filter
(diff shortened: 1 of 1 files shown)


dcwhite:

Issue #2664, opened 2026-08-06
Mac packaging: macdeployqt bundle chain runs 4x, 30 of the step's 35 minutes


dcwhite:
Now that #2619 has parallelized the build (~62 min serial → 42 min at `-j3`), **packaging is the second-largest cost in `mac-gui` and ~45% of the job**. It had not been looked at.

Measured on run [31078669153](https://github.com/SCIInstitute/SCIRun/actions/runs/31078669153) (branch `ci/parallel-build-jobs`, commit e9a6fc928), job `mac-gui`, step "Package (optional - Unix/macOS)": **34.6 min** of a 78 min job.

## Where the time goes

Shell-trace timestamps, step start 12:22:03 → 12:56:37:

| Phase | Wall |
|---|---|
| `cmake -DBUILD_BUNDLE:BOOL=ON` reconfigure | 3.2 s |
| `make -j3` | 10 m 28 s — 2 m 39 s recompile + **7 m 45 s bundle chain (pass 1)** |
| `make package` | 23 m 47 s — see below |
| `repair_package.sh` | 15.9 s |

`make package` internals:

| Sub-phase | Wall |
|---|---|
| relink cascade (3 factory files regenerated) | 44 s |
| **bundle chain (pass 2)** | 7 m 37 s |
| CPack productbuild "Run preinstall target" → **bundle chain (pass 3)** | 7 m 27 s |
| productbuild actually builds the `.pkg` | **15 s** |
| CPack TGZ "Run preinstall target" → **bundle chain (pass 4)** | 7 m 22 s |
| tar.gz actually compresses | **14 s** |

**The macOS bundle chain (`macdeployqt` + `install_name_tool` + framework copies) runs four times, ~7.5 min each = 30 m 11 s — 87% of the step.**

## Two hypotheses that turned out to be wrong

Recording these so nobody re-chases them:

- **Compression is not pathological.** productbuild + tar.gz together are **29 seconds** out of 34.6 minutes. Nothing to tune.
- **`macdeployqt` is not walking the Python framework.** Python appears in 3 lines of a `-verbose=2` run. The 198 MB / 7,827-file framework is untarred into `Contents/Frameworks` but not deeply scanned.

## Why one pass costs 7.5 min

[...]

PR #2680, opened 2026-08-09
CI: fix nightly Slack summary parsing, and make coverage actually instrument

dcwhite:
Two independent nightly-reporting bugs, found while looking at run [31250403273](https://github.com/SCIInstitute/SCIRun/actions/runs/31250403273). Both reported success while producing nothing.

## 1. The Slack summary only ever reported Windows

`nightly-slack.yml` globbed `artifacts/*/*-test-results.txt`. The artifact layout is not uniform:

```
test-results-Windows-headless/unit-test-results.txt          ← matched
test-results-Linux-headless/tmp/unit-test-results.txt        ← one level too deep
test-results-macOS-gui/tmp/unit-test-results.txt             ← one level too deep
```

The Unix upload step in `reusable-build.yml` mixes an absolute path (`/tmp/unit-test-results.txt`) with workspace-relative ones, so `upload-artifact` computes `/` as the least common ancestor and preserves the leading `tmp/`. Windows uses all-relative paths, so its LCA is the workspace. Linux and macOS counts have been silently missing from every nightly post.

### What changed

The inline shell is replaced by `scripts/ci/slack_test_summary.py`, which walks with `rglob` and reads the `junit-*.xml` files already sitting in those artifacts instead of grepping ctest's console summary. That yields failing test *names*, not just counts — which is the part worth seeing at 8am. On the run above:

| Job | Unit | Regression |
|---|---|---|
| Linux-headless | 1 failed / 488 | 8 failed / 247 |
| Windows-headless | 1 failed / 489 | 8 failed / 247 |
| macOS-gui | 4 failed / 490 | 12 failed / 309 |

Linux and Windows fail the *same* 8 regression cases; macOS fails a disjoint 12, mostly rendering. The single failing unit suite is different on all three (`Algorithms_Field_Tests` / `Core_Python_Tests` / four `.Test.ImportNetwork.*`). None of that could reach Slack before.

Output is Block Kit: a per-job field grid, a context block per failing report with up to 5 names and an "…and N more" tail, then the run link. The yellow-circle downgrade for hidden `continue-on-error` failures is preserved.

Note the junit totals (488/489/490) exceed ctest's console "out of 479" because ctest excludes disabled tests from its denominator. The XML count is the honest one.

### Why a script rather than inline YAML

`workflow_run` only fires for workflows on the default branch, so this notifier cannot be exercised from a PR — including this one. A standalone script can be run against downloaded artifacts, which is the only way any of it gets tested before going live:

```bash
gh run download 31250403273 --dir artifacts --pattern 'test-results-*'
WF_NAME=regression-tests WF_CONCLUSION=success python3 scripts/ci/slack_test_summary.py artifacts
```

Verified against three cases: the full artifact set above (9 blocks), no artifacts at all (build-only workflows collapse to header + footer), and a missing junit file (falls back to the old ctest-text regex).

This adds an `actions/checkout` step to the notifier, which it did not previously need. `workflow_run` always resolves the workflow against the default branch, so that is always master's copy of the script, never a PR's.

## 2. Coverage has never been instrumented

`src/CMakeLists.txt` set `CMAKE_CXX_FLAGS` for `ENABLE_COVERAGE` at line 1055, but every `ADD_SUBDIRECTORY` call is at lines 631–645. CMake captures directory-scoped variables when the subdirectory is added, so the flags applied to no target — while `-- Code coverage instrumentation enabled` still printed, which is what made it look healthy.

The full chain: no `-fprofile-instr-generate` anywhere in the 39k-line build log → zero `.profraw` → `coverage.sh` exits 1 with `error: no .profraw files produced` → `continue-on-error: true` swallows it and the step reports **success** → upload finds nothing, warns, uploads no artifact → green `mac-coverage` job, empty report, every night.

### What changed

- Moved the coverage block above the `ADD_SUBDIRECTORY` calls. `ENABLE_GCOV_DATA_FILES` had the identical bug and moves with it. Straight move, no logic change, inert unless one of the options is `ON` — so no normal build is affected.
- `if-no-files-found: error` on the coverage upload, so this cannot fail silently again.

## Testing / what to watch

The Slack script is tested locally as described. The workflow wiring itself cannot be tested before merge (see above).

**The coverage half is not verified by a build.** The argument is structural (CMake scoping) plus the empirical zero-`.profraw`; I have not run an instrumented build. The check is the first nightly `regression-tests` after this merges: `mac-coverage` should produce a `coverage-report-macOS` artifact containing `summary.txt`. If instrumenting every target pushes that job's runtime up a lot, that is worth a follow-up — an instrumented Debug build of the whole tree is not cheap.

Related: #2664 (where the nightly artifact digging started).

### DIFF
--- a/.github/workflows/nightly-slack.yml
+++ b/.github/workflows/nightly-slack.yml
@@ -12,16 +12,17 @@ name: nightly-slack
 #
 #  * `workflow_run` only fires for workflow files that live on the DEFAULT
 #    branch. This notifier therefore does nothing until it is merged to master,
-#    and cannot be exercised from a PR branch. That is expected for nightly
-#    infrastructure -- there is no way to unit-test a workflow_run trigger from a
-#    feature branch.
+#    and cannot be exercised from a PR branch. That is why the payload is built
+#    by scripts/ci/slack_test_summary.py rather than inline shell -- the script
+#    runs against downloaded artifacts locally, which is the only way any of this
+#    gets tested before it is live.
 #
 #  * It reports the REAL ctest pass/fail counts, not just the job conclusion.
 #    The ctest steps in reusable-build.yml carry `continue-on-error: true`, so
 #    regression-tests reports "success" even when hundreds of cases fail. The
-#    counts are recovered by downloading that run's `test-results-*` artifacts
-#    and parsing the "N% tests passed, M tests failed out of K" summary lines.
-#    A green job with failing tests is flagged yellow rather than green.
+#    counts (and the failing test names) come from the junit-*.xml files in that
+#    run's `test-results-*` artifacts. A green job with failing tests is flagged
+#    yellow rather than green.
 #
 #  * Filtered to `event == 'schedule'`: only the nightly cron pings Slack. Master
 #    pushes and workflow_dispatch runs do not (pushes still get GitHub's default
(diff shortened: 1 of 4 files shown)


dcwhite:

PR #2684, opened 2026-08-09
ViewScene movie recording with a proper output-file dialog (#279)

### LINKED ISSUE #279: ViewScene Record Movie UI upgrades
Replace file name text entry with file selection dialog. Make sure outstanding bugs are fixed (see GForge). The radio button controls shouldn't include the stop recording function (move to button?).

dcwhite:
Closes #279.

## Scope: the recorder had to exist first

#279 is a GForge ticket carried over during the 2013 migration — it was filed the same day as the `Test #NN:`-prefixed imports, and its body says "see GForge" outright. So it describes **v4's** Record Movie panel: the text entry to replace, and the radio group with "stop" in it. Neither was ever ported to v5, which has no movie recording at all (`git grep -i movie` over `src/` hits only `QMovie`, the executing-module GIF, and `TODO.md`).

The three asks still make sense as a spec — they just apply to a recorder that has to be built. This builds it and satisfies all three.

One note for anyone reading the thread: the `Triage 2026-06` comment says the movie path widget is live "in the ViewScene config dialog". It isn't.

## What's here

- **`MovieRecorder`** (`Interface/Modules/Render/MovieRecorder.{h,cc}`) — frames are grabbed on the GUI thread, since `grabFramebuffer` requires it, and encoded on a worker thread.
- **Three formats.** A numbered PNG sequence needs nothing. MP4 and animated GIF pipe raw RGBA to `ffmpeg` when it is found on `PATH`. **No new build dependency** — the video entries are disabled in the combo when ffmpeg is absent, and `validate()` says so plainly if a saved network selects one anyway.
- **`MovieRecord.ui` / `MovieRecordControls`** — output file chosen through `QFileDialog` (ask 1), and **separate Record and Stop buttons** rather than v4's radio group with a "stop" entry (ask 3). Format combo, frame rate, and a live status line.
- **Toolbar button** beside Screenshot: red dot, switching to a white square on a red button while recording.
- **State**: `MovieOutputPath` / `MovieFormat` / `MovieFrameRate` persist in `.srn5`.
- Docs section in `docs/modules/Render/ViewScene.md`.

## The outstanding bugs (ask 2)

The three GForge bugs the issue refers to survive in `TODO.md`. Each is addressed by construction:

| Bug | Fix |
|---|---|
| record movie fails silently if path doesn't exist | destination validated and created before recording starts; anything wrong goes to a message box. `TODO.md` entry moved to Resolved. |
| crash writing to a directory without permission | writability probed by actually creating a file — the permission bits lie on network shares and read-only volumes |
| record movie causes SCIRun to hang | the frame queue is bounded, so a slow disk or a stalled encoder drops frames and reports the count instead of blocking the GUI |

## Notes for review

- Movie size comes from a real `grabFramebuffer()` rather than `mGLWidget->size()`; those differ by the device pixel ratio, so HiDPI displays record at full resolution.
- `addComboBoxManager(combo, key)` stores the combo's **display text**, not its index — the `StringPairs` overload is used so state holds `png`/`mp4`/`gif` tokens and the GUI labels stay free to change.
- Deliberately not included: no keyboard shortcut (would need a new entry in the fixed `ShortcutDef::Id` table), no record-on-execute mode, no icon resource (the dot and square are drawn with `QPainter`).

## Testing

`MovieRecorderTests.cc` covers the format/suffix mapping and every validation branch: 8 pass, 1 skips where ffmpeg is installed.

Beyond the unit tests, I compiled all four touched translation units plus generated moc against this repo's real build flags (Qt 6.10.2, `-Wall`) — clean, the one warning being a pre-existing `QCheckBox::stateChanged` deprecation — and ran the recorder end to end headless. It produced a valid h264 MP4 at the requested frame rate with every frame present (`ffprobe`-confirmed), a valid GIF, and a numbered PNG sequence; an extracted frame confirmed no RGBA channel swap.

**Not verified:** the toolbar layout and popup have not been seen on screen — that needs a full Superbuild configure, which I did not have here. Worth a look before merging.


dcwhite:
@jessdtate HA this one I did while explaining to someone what Claude can do. It actually works pretty well, if ffmpeg is already installed. Let me know if you want to hold off on this feature though, since it's somewhat significant.

jessdtate:
this is pretty cool, thanks for working on it!  I want to test it out before merging.  A future step could be packaging ffmpeg or a install shortcut, which shouldn't be too hard.

PR #2685, opened 2026-08-10
ShowField: fix reverse-face colors on boundary faces of vector/tensor fields

### LINKED ISSUE #2439: Rendering bug with clipping planes and edge elements
this has been seen in both mac (itel chip Nvidia GPU) and windows

Image

the elements have the wrong coloring on the reverse faces.  This seems to only happen with elements that are on the edge of the mesh.

dcwhite:
Fixes #2439

## The bug

`renderFacesLinear` builds **two-sided** faces for cell-centered data: `aTexCoords.x` holds the colormap index of the face's first adjacent cell, `.y` the second. The `*_ColorMap` fragment shaders pick between them by facing direction:

```glsl
if(gl_FrontFacing) colorMapValue = texture2D(uTX0, vec2(vTexCoords.y, 0.0));
else               colorMapValue = texture2D(uTX0, vec2(vTexCoords.x, 0.0));
```

A face on the mesh boundary has only one adjacent cell, so both coordinates should get that cell's value. The scalar branch did this. The vector and tensor branches copied the scalar fallback verbatim:

```cpp
fld->get_value(vvals[0], cells[0]);
if (cells.size() > 1) fld->get_value(vvals[1], cells[1]);
else svals[1] = svals[0];        // never touches vvals[1]
```

So `vvals[1]` / `tvals[1]` kept whatever the previously iterated interior face left behind (or a default-constructed zero on the first face), and the reverse side of every boundary face got a colormap index from an unrelated element. It only becomes visible once a clipping plane exposes interior faces, which is how it was reported.

The typo predates the 2019 `renderFacesLinear` refactor (0d5c71004) — it is in the original two-sided face code.

## Test

Adds a regression test that pulls texture coordinates back out of the face VBO for a six-tet cube with scalar, vector, and tensor cell data, then checks that boundary faces carry one value on both sides and interior faces carry two.

- Fails on the unfixed code (`boundary face 0 … 0.851 vs 0.502`, vector and tensor paths).
- Passes with the fix.
- Full `Modules_Visualization_Tests` suite: 134/134 pass on macOS.


dcwhite:
@basisunus — here is a self-contained test net for this, plus what to look for. No SCIRunTestData needed, it builds its own mesh and data.

**Net:** `src/ExampleNets/regression/Renderer/showFieldTwoSidedFacesVectorCellData.srn5` (added in 8366acfc6)

```
CreateLatVol 24^3
  -> ConvertHexVolToTetVol
  -> CreateFieldData   RESULT = Vector(X, Y, Z);   Format Vector, Basis Constant
  -> ClipFieldByFunction  X*X + Y*Y + Z*Z > 0.16   (Element Center)
  -> ShowField (faces only, coloured by colormap)  -> ViewScene
     CreateStandardColorMap -> RescaleColorMap -> ShowField port 1
```

`CreateFieldData` with **Format = Vector** and **Basis = Constant** is the whole point — the field arriving at ShowField is `GenericField<TetVolMesh<TetLinearLgn<Point>>, ConstantBasis<Vector>, vector<Vector>>`. Scalar fields were never affected, so a scalar net looks identical before and after the fix and proves nothing. `ClipFieldByFunction` carves a sphere of radius 0.4 out of the middle so the mesh has an internal boundary, like the electrode/sphere in the issue screenshot.

`RESULT = Vector(X, Y, Z)` makes `|v|` the distance from the centre: smooth, and spanning the full colormap range. That matters — see below.

## Walk through

1. Open the net and hit execute. It takes a few seconds (~60k tets).
2. Open the ViewScene. You are looking at the outer surface of the cube.
3. **On master you should see scattered mis-coloured triangles over that surface** — individually wrong triangles in colours drawn from all over the range, on top of the smooth radial gradient. On this branch the surface is a clean gradient.
4. Now open **Clipping Planes** in the ViewScene controls, tick **Plane 1 → Visible**, and drag **D** until the plane cuts into the cube and the spherical cavity opens up.
5. Look at the **far wall of the cavity** (the inside surface curving away from you). On master that wall is speckled the same way. The near wall, and all the interior faces you have just exposed, are correct in both builds.
6. Turning **Show Edges** on in ShowField makes the individual bad triangles easier to pick out.

## Why those two surfaces and not the others

Cell-centred data makes each face two-sided: `aTexCoords.x` is the colormap index of the face's first adjacent cell, `.y` the second, and the fragment shader picks between them with `gl_FrontFacing`:

```glsl
if(gl_FrontFacing) colorMapValue = texture2D(uTX0, vec2(vTexCoords.y, 0.0));
else               colorMapValue = texture2D(uTX0, vec2(vTexCoords.x, 0.0));
```

`.y` is the one the vector/tensor paths left stale on boundary faces. A boundary face's winding normal points **out of the mesh** (I checked this directly — for a tet cube, 12 of 12 boundary-face normals point away from their owning cell), so a camera outside the mesh sees the outer surface **front-facing**, which reads `.y`. Same for the far wall of a cavity: its normal points into the cavity, i.e. toward you. Everything else — interior faces, the near cavity wall — reads `.x` or has two genuine values, and was always right. That is why it read as a clipping-plane bug: the cavity wall only becomes visible once you clip.

The wrong colours look random rather than merely "off by one element" because `TetVolMesh::compute_faces` fills `faces_` in **hash-table order**, so the stale value comes from an arbitrary cell somewhere else in the mesh, not a neighbour. That is also why the field here is deliberately smooth and full-range: against a smooth gradient, a value pulled from a random cell stands out.

## What I could not check

Everything above is verified on macOS: the unit test in this PR fails on master and passes here, and the net loads, executes and delivers the right field type. **I have no Windows verification** — headless has no GL context and I could not get a screenshot out of the offscreen path, so nobody has actually watched this render on the platform where you reported it. That confirmation is the useful thing you could add.

jessdtate (approved):

jessdtate (approved):

Issue #2398, opened 2022-08-23
Test on Windows 11


dcwhite:
Other CIBC software is having issues.

dcwhite:
Mainly VS 2013 runtime install issues.

dcwhite:
I can test this tomorrow for closure

dcwhite:
**Analysis (2026-06-13)**

Windows 11 testing. Comment says VS2013 runtime install issues with other CIBC software. SCIRun 5 uses VS2019/2022 runtimes and should be compatible with Windows 11. If the commenter can run the latest installer on Windows 11 and confirm no issues, this can be closed.

dcwhite:
I've built and run on Win11, but not tested installers yet. Let me do that with the Qt5 and 6 build output from Github Actions. Where are those binaries stored, btw? (question for claude)

dcwhite:
The CI installers aren't in a persistent store — they're GitHub Actions **run artifacts** from the `windows-build` workflow. Grab them from the [windows-build runs](https://github.com/SCIInstitute/SCIRun/actions/workflows/windows.yml): open a run (a master push, the nightly cron, or a manual `workflow_dispatch`) and scroll to **Artifacts** at the bottom of the run summary page. Four GUI installers:

- `SCIRunWindowsInstaller_qt5` — Qt 5.15.2
- `SCIRunWindowsInstaller_qt6` — Qt 6.3.*
- `SCIRunWindowsPythonInstaller_qt5` — Qt 5.15.2 + Python
- `SCIRunWindowsPythonInstaller_qt6` — Qt 6.3.* + Python

Each zip holds the NSIS `SCIRun-*-win64.exe` (uploaded by the "Upload installer" step in `reusable-build.yml`). Caveats: default 90-day retention (they expire), GitHub login required to download, and PRs don't produce them (the slim `pr.yml` tier only). macOS equivalents come from `mac.yml`, all Qt6.

Note these are ephemeral by design — if we want installers that stick around, the natural fix is publishing them as **GitHub Release assets** (permanent, public, no login) on tags, plus a rolling `nightly` prerelease for dev builds. Happy to wire that up as a follow-up.

dcwhite:
Waiting on #2590

dcwhite:
I did a smoke test of Version: v5.0-beta.2026-502-g299c6234e+Qt6.3.2 from the nightly build, and it worked OK. Closing this one now.

PR #2693, opened 2026-08-11
CI: fix empty nightly release; recurse for installers (#2398)

dcwhite:
## Problem

The `nightly` GitHub Release (added in #2588) exists but has **zero assets**. The publish job goes green while attaching nothing, so there's still no durable copy of the installers.

## Root cause

Each build artifact preserves its original build path inside it — the installer lands at `dist/<artifact>/bin/SCIRun/SCIRun-*.pkg` (and the `.msi` under `build/**`). The rename step in `publish-release.yml` searched with `find -maxdepth 1`, one level too shallow, so it found no installer, printed `No installer in .../ skipping`, then the cleanup line deleted the subdirs — **taking the real installers with them**. `files: dist/*` then matched nothing, but `softprops/action-gh-release` still exits 0.

From the mac nightly run log:
```
Starting download of artifact to: .../dist/SCIRunMacInstaller
No installer in SCIRunMacInstaller/, skipping
🤔 Pattern 'dist/*' does not match any files.
```

## Fix

Drop `-maxdepth 1` so `find` recurses into the artifact tree. Verified locally against the real nesting — the rename yields unique flat names:

```
SCIRunMacInstaller.pkg
SCIRunMacInstaller-26.pkg
SCIRunWindowsInstaller_qt6.exe
```

## Note (separate issue)

The **Windows matrix is currently failing** on unrelated `vs2026` build breakage, so it produces no installers regardless. This PR only fixes the mac side reaching the release; Windows won't populate until that build failure is resolved.

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -8,6 +8,8 @@ on:
   push:
     branches:
       - master
+    tags:
+      - 'v*'
     paths-ignore:
       - 'docs/**'
   schedule:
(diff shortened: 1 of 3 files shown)


jessdtate (approved):

dcwhite:
I merged this one manually, since it got the wrong branch base. We should see fixed builds published tomorrow.

PR #2697, opened 2026-08-13
CI: unblock the VS 2026 nightly (Qt stdext, CPython UWP coroutine)

dcwhite:
Both remaining `windows-2025-*-vs2026` failures in [run 31687060859](https://github.com/SCIInstitute/SCIRun/actions/runs/31687060859) are upstream headers meeting a toolchain that dropped deprecated extensions — neither is SCIRun code. Boost/b2, Cleaver2, Freetype, Glew, Tetgen, Teem all build clean under v145 now, so these two are what's left.

## The GUI job's "Qwt" error is a Qt error

```
D:\a\SCIRun\Qt\6.3.2\msvc2019_64\include\QtCore\qvarlengtharray.h(369,19):
  error C3861: 'stdext': identifier not found
  (compiling source file '_deps/qwt_upstream-src/src/qwt_symbol.cpp')
```

Qt 6.3.2's `qvarlengtharray.h` calls `QT_MAKE_CHECKED_ARRAY_ITERATOR`, defined on MSVC in `qcompilerdetection.h:107` as `stdext::make_checked_array_iterator`. MSVC 14.51 removed `stdext`, so this fires in every Qwt TU that instantiates `QVarLengthArray::operator=(initializer_list)` or `erase()`. Qwt's own source is fine — it's just the first thing in the build that includes QtCore.

The job had inherited `qt-version: '6.3.*'` from the windows-2022 matrix, where that pin is deliberate (it's our Qt6 floor). This job exists to exercise the compiler, not the Qt floor, so it moves to 6.10.2 like the other current-Qt jobs. Qt still has one such call in `QVarLengthArray::erase` as of 6.10 and dev; nothing in the Qwt build instantiates it.

## Headless is stuck on CPython's UWP launchers

```
...MSVC\14.51.36231\include\experimental\coroutine(37,1): error C2338:
  error STL1011: ... <experimental/coroutine> ... deprecated ... REMOVED SOON.
  (compiling source file '../PC/python_uwp.cpp')  [python_uwp.vcxproj]
```

`pythoncore`, `python`, `pythonw` and `pywlauncher` all build; only `python_uwp`/`pythonw_uwp` fail, on a header C++/WinRT drags in. Nothing selects projects out of `pcbuild.sln` and SCIRun ships none of the UWP launchers, so this takes the escape hatch the static_assert names, passed via `CL` on all three msbuild entry points (build.bat, Release, Debug).

## Testing

Both jobs are `allow-failure: true` and nightly-only, so this can't regress the PR tier. Verifying needs a `workflow_dispatch` of windows-build on this branch (the cron only runs on master).

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -97,6 +97,10 @@ jobs:
   # Nightly-only, not the PR tier (pr.yml), so a VS-2026 break can't block pull
   # requests (#2449). allow-failure until the externals catch up: Boost 1.90's
   # b2 can't bootstrap under VS 2026 (#2657), fixed by the 1.91 upgrade (#2548).
+  #
+  # Qt here is 6.10.2, not the '6.3.*' floor the windows-2022 jobs pin: Qt 6.3's
+  # QVarLengthArray calls stdext::make_checked_array_iterator, which the v145 STL
+  # no longer ships. This job exists to exercise the compiler, not the Qt floor.
   # ---------------------------------------------------------------------------
   windows-2025-headless-vs2026:
     uses: ./.github/workflows/reusable-build.yml
@@ -112,7 +116,7 @@ jobs:
     with:
       runner: windows-2025
       cmake-generator: 'Visual Studio 18 2026'
-      qt-version: '6.3.*'
+      qt-version: '6.10.2'
       variant: gui
       build-with-python: false
       artifact-name: SCIRunWindowsInstaller_qt6_vs2026
--- a/Superbuild/PythonExternal.cmake
+++ b/Superbuild/PythonExternal.cmake
@@ -113,6 +113,12 @@ ELSE()
     # PlatformToolset when unset, so pass it through the environment instead.
     SET(python_MSBUILD_TOOLSET_ENV "PlatformToolset=${CMAKE_VS_PLATFORM_TOOLSET}")
   ENDIF()
+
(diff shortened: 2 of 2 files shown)


dcwhite:
Verified on a `workflow_dispatch` of windows-build against this branch: [run 31735068915](https://github.com/SCIInstitute/SCIRun/actions/runs/31735068915).

Both original failures are gone — zero `stdext` hits in the GUI log, zero `STL1011` in the headless log:

**GUI (Qt 6.10.2, v145)** — Qwt now builds and installs clean, the exact step that failed before:

```
-- Qwt_external configure command succeeded.
   Performing build step for 'Qwt_external'
-- Qwt_external build command succeeded.
-- Qwt_external install command succeeded.
```

**Headless** — CPython builds through, `python_uwp`/`pythonw_uwp` included; nothing from `<experimental/coroutine>`.

Both jobs then hit the same next wall, which is the known #2657:

```
Bootstrapping b2
### "Unknown toolset: vcunk"
Failed to build Boost.Build engine.
```

Worth noting for #2657: it is written up as GUI-only, because until now the headless job died at CPython before Boost finished cloning. It is both jobs. Neither vs2026 job can go green until the Boost 1.91 upgrade (#2548) lands — these two fixes just clear the walls in front of it.

Also split out of this work: #2698 tracks the `QVarLengthArray::erase` call that still uses `stdext` in Qt 6.10 and `dev`, latent for now.

jessdtate (approved):

Issue #1663, opened 2017-06-29
signed packages


jessdtate:
productbuild should make it easy to created signatures for os x pacakges.  This will allow release on the apple store and such.  Ed should have a certificate to use.

dcwhite:
@jessdtate Any other CIBC products figure this out yet?

jessdtate:
no, but it is still on my wishlist

dcwhite:
@jessdtate has a new certificate. Need to see who wants to figure out how to use it first.

dcwhite:
**Investigation findings (2026-06-12)**

Signed macOS packages allow distribution outside direct download (App Store, Gatekeeper notarization) and remove the 'unidentified developer' warning on install.

**Current state:** The CI packaging step (reusable-build.yml) runs make package / cmake --build --target package to produce a .pkg file, but no signing or notarization steps exist.

**What is needed:**
1. A Developer ID certificate (comments say Jess has one) stored as a GitHub Actions secret
2. productbuild --sign 'Developer ID Installer: ...' to sign the .pkg
3. For macOS 10.15+, notarization is also required: xcrun notarytool submit + xcrun stapler staple
4. The signing identity and Apple ID credentials need to be stored as GitHub Actions secrets (MACOS_CERTIFICATE, MACOS_CERTIFICATE_PWD, APPLE_ID, APPLE_TEAM_ID, etc.)

**GitHub Actions implementation:** Add a 'Sign and notarize' step after the Package step in reusable-build.yml, gated on runner.os == 'macOS' and a secret being present (so unsigned builds still work for forks/PRs).

**Effort:** Small-medium for the CI workflow changes. The main prerequisite is having the certificate loaded into GitHub secrets, which is an admin task.

dcwhite:
## Why this matters more now

Branch `claude/mac-installer-publishing-2435b8` adds a rolling `nightly` prerelease to `mac.yml`: the four macOS `.pkg` installers the nightly already builds get attached to a permanent `releases/tag/nightly` link, and that link is posted to the CI Slack channel alongside the nightly test results.

Until now the installers were Actions artifacts — 90-day expiry, signed-in GitHub account required — so almost nobody hit the Gatekeeper wall. Once there's an anonymous download link, everyone who clicks it does. The release body currently ships the `xattr -dr com.apple.quarantine` workaround as a stopgap.

## Correcting the 2026-06-13 analysis above

[...]

PR #2702, opened 2026-08-13
CI: publish nightly mac installers to a rolling prerelease

dcwhite:
The mac nightly already builds four `.pkg` installers, but they only ever existed as Actions artifacts: 90-day expiry, and a signed-in GitHub account needed to download one. This attaches them to a rolling `nightly` prerelease so there is a permanent, anonymous download link, and posts that link to the same Slack channel as the nightly test results.

## `mac.yml` — new `nightly-release` job

Schedule-only (master pushes would rewrite the release several times a day), `needs` the four installer-producing jobs with `if: always()` so one red variant still publishes the rest.

Three details that aren't obvious from the diff:

- **The packages get renamed.** All four variants emit `SCIRun-<version>-Darwin.pkg` — CPack sets a single `PKG_FILE_NAME` in `src/CMakeLists.txt`. As release assets those identical names would overwrite each other, so each gets a runner/arch/python suffix: `SCIRun-5.0.0-macos26-arm64.pkg`, `-macos26-x86_64.pkg`, `-macos-latest-arm64.pkg`, `-macos-latest-arm64-nopython.pkg`.
- **Updating the release force-moves the git ref.** GitHub ignores `target_commitish` on an already-published release, so without the explicit ref update the release page would keep showing whatever commit the tag was first cut from, forever.
- **Yesterday's assets are deleted before upload,** so a variant that fails tonight doesn't leave a stale installer sitting next to fresh ones.

## `nightly-slack.yml` — installer link

Adds a `:package:` line to the existing `mac-build` message. It only appears when the `nightly` tag's sha matches the run's head sha — if the release job failed, the message omits the link rather than quietly pointing at the previous night's build.

## Testing

`workflow_run` and `schedule` triggers only fire from the default branch, so **none of this can be exercised from a PR branch** — the same caveat already documented at the top of `nightly-slack.yml`. The first real exercise is the first nightly after merge. What I did verify locally: both files parse as YAML, and the rename/notes-table logic was dry-run against a simulated artifact set including a missing variant.

Everything is additive — a new job at the end of `mac.yml` plus one hunk in `nightly-slack.yml` — so it should not conflict with the other CI work in flight (#2575, #2576, #2644).

## Not signed

The packages remain unsigned and un-notarized, so Gatekeeper will refuse them on download. The release body ships the `xattr -dr com.apple.quarantine` workaround as a stopgap and points at #1663, which now carries the full signing process, ordering constraints, and a TODO list. Publishing these behind a public link raises the priority of that issue but doesn't depend on it.

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -168,3 +168,112 @@ jobs:
       variant: headless
       build-testing: true
       run-smoke-test: true
+
+  # ---------------------------------------------------------------------------
+  # Rolling nightly prerelease
+  # ---------------------------------------------------------------------------
+  # Republishes the night's .pkg installers under a single `nightly` tag that is
+  # rewritten in place, so https://github.com/.../releases/tag/nightly is a
+  # permanent link. Actions artifacts expire after 90 days and need a signed-in
+  # GitHub account to download; release assets have neither limit.
+  #
+  # Schedule only -- master pushes would rewrite the release several times a day.
+  nightly-release:
+    # always(): a failed variant should not suppress the ones that worked. The
+    # release body lists what actually made it in.
+    if: ${{ always() && github.event_name == 'schedule' }}
+    needs: [mac-gui, mac-gui-26, mac-gui-26-intel, mac-gui-nonpython]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write   # create/update the release and move the tag
+      actions: read     # download this run's artifacts
+    env:
+      GH_TOKEN: ${{ github.token }}
+      GH_REPO: ${{ github.repository }}
+      TAG: nightly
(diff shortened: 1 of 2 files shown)


dcwhite:
Closing: this duplicates work already on master. `publish-release.yml` (#2398, #2588) has published mac and Windows installers to a rolling `nightly` release since 2026-08-08, and handles the `v*` draft-release case this branch does not.

I built this from a stale worktree whose branch predated that merge, and `git diff master...HEAD` hid it — a three-dot diff only shows what a branch adds over the merge base.

The two parts that are genuinely missing from master — a Slack link to the nightly release, and a note in the release body that the packages are unsigned — are split out into a much smaller PR against the existing publisher.

Issue #2578, opened 2026-07-25
Color map import


Sumientra:
- The ReadColorMapXml module does not seem to work. It reads my xml file but its output is empty and SCIRun will crash if you pipe it into any other module. 
- There is no WriteColorMapXml equivalent, which would be useful to have. 
- Would also be useful to have a module that converts a matrix to a colormap.
- Seconding Jake's request elsewhere to bring the "convert to RGB values" back to ShowField. There is something similar in ShowFieldGlyphs but it only shows points or spheres, not faces.

dcwhite:
I will look at this one this week!

Sumientra:
Tested:

- ReadColorMapXml works with my XML file after replacing "ColorMapNode" with "Point"
- WriteColorMapXml works
- ConvertMatrixToColorMap works after removing the first and last columns from the XML colormap, retaining only RGB values

dcwhite:
@Sumientra what tool exported the `ColorMapNode` file? If you can attach it, I'll use it as a test fixture — the reader should accept that dialect instead of you renaming tags.

Sumientra:
I had Gemini make it. I'll attach in case it's helpful, but I don't need SCIRun to be able to read this. I'm fine with the SCIRun format, now that I can export it with WriteColorMapXml. 

[ColorMap_TER.xml](https://github.com/user-attachments/files/31082899/ColorMap_TER.xml)

PR #2704, opened 2026-08-14
Colormap import fixes from #2578 testing feedback

dcwhite:
Follow-up to the testing feedback on #2578. Sumientra tested the three colormap modules and found each of the first two only worked after hand-editing the data.

## ReadColorMapXml: `<ColorMapNode>` had to be renamed to `<Point>`

The point-element whitelist was already two names deep (`Point`, `color`) after one earlier reporter, so this replaces name matching with attribute matching: any child element carrying `r`, `g` and `b` is a color stop. An unrecognized *wrapper* element is resolved the same way — the colormap is whichever element actually holds the points.

Since the rename alone made their file work, the attributes were already correct; only the tag differed. Existing dialects and the "no recognized points" rejection from #2579 behave exactly as before (covered by the pre-existing tests).

## ConvertMatrixToColorMap: 5 columns had to be trimmed to 3

Their matrix was position + RGBA. Columns now carry explicit roles — Ignore / Position / Red / Green / Blue / Alpha:

- **Auto-detect** (default, unchanged behavior for existing networks) handles 3 (RGB), 4 (RGBA) and now 5 (position + RGBA).
- **A new dialog** lists one row per input column with a role dropdown, for any other layout or when the guess is wrong. The column count is only known once a matrix arrives, so the table populates on execute, in the same way `BuildBEMatrix`'s does.
- A **Position** column is honored by resampling the piecewise-linear ramp its stops describe, since `ColorMap` itself stores colors evenly spaced and has no per-stop positions.

Two bugs fall out of knowing what each column is:

- The 0-255 autoscale used `maxCoeff()` over the *whole* matrix, so a 0-1 alpha alongside 0-255 colors was divided by 255 (everything transparent), and a position column would have been scaled off too. Color channels, alpha, and position are now each handled on their own terms.
- A 4-column matrix was always read as RGBA with no way to say `[x, r, g, b]`.

Duplicate role assignments and unassigned R/G/B are rejected with a specific message rather than silently producing a wrong map.

## Not in this PR

The fourth bullet of #2578 — "Conversion to RGB" coloring in ShowField — is untouched. The radio buttons exist but are hidden, and the faces path has no `COLOR_IN_SITU` branch at all (nodes and edges do), so faces need real renderer work rather than un-hiding a button. Worth its own issue.

## Testing

Built and ran against the existing externals:

- `Modules_Visualization_Tests` — 151/151 pass, including 10 for this module (6 new: 5-column auto-detect, explicit role mapping, unassigned and duplicate roles, and alpha surviving the byte rescale).
- `Core_Datatypes_Tests` — 151/151 pass, including 15 `ColorMapXmlTests` (2 new: the `ColorMapNode` dialect and an unknown wrapper element).

Both module- and dialog-factory generators were run to confirm the new UI entry in the `.module` config wires up.

Still needs a human at the GUI: the dialog's table has not been clicked through in a running SCIRun.

Closes nothing on its own — #2578 stays open for the ShowField item.


dcwhite:
@jessdtate wants to test

jessdtate (approved):

Issue #2711, opened 2026-08-14
CI: no test failure blocks anything; flip continue-on-error off per platform as each goes green


dcwhite:
Every ctest step in `reusable-build.yml` carries `continue-on-error: true` ([lines 330, 339, 358, 368](https://github.com/SCIInstitute/SCIRun/blob/master/.github/workflows/reusable-build.yml)). The consequence is worth stating plainly:

**No test failure, on any tier, blocks anything today.**

- The nightly `regression-tests` run reports ✅ with ~8-10 failures per platform. [Run 31688134956](https://github.com/SCIInstitute/SCIRun/actions/runs/31688134956) is green and has 26 failing tests across three platforms.
- The PR tier runs exactly one test job, `linux-unit-tests` (pr.yml), through the same reusable workflow — so its unit-test step is `continue-on-error` too, and `ci-gate` reports success no matter what fails. A PR that breaks the suite merges green.

## Plan: flip per platform, per suite, as each reaches zero

Rather than one global flip gated on the whole matrix being green:

**1. Linux unit → blocking.** One test away: #2710. This is the highest-value step in the list, because `linux-unit-tests` is the *only* test job gating pull requests. Fixing that one test converts the required check from decorative to real.

**2. Windows unit → blocking.** Blocked on #2708 (7 `FieldConversionTests` crashing with an access violation).

**3. macOS unit → blocking.** Blocked on #2709 (stale dialog-count constant, likely one line) and the 4 `ImportNetwork` v4-net timeouts.

**4. Linux + Windows regression → blocking.** Blocked on #2699 (4 Python networks) and #2701 (4 module-error networks). Both platforms fail the *identical* 8 networks and have for at least two nights, so these are deterministic and tractable — they do not depend on the macOS crash.

**5. macOS regression → blocking, last.** Blocked on #2700, an intermittent bus error that takes a different 7-13 ViewScene networks each night. This one cannot be verified by a single green run; it needs several consecutive clean nights, which is exactly why it should not gate the other five steps.

## If step 5 drags

If #2700 turns out to be a deep concurrency problem, consider a baseline ratchet for macOS instead of waiting: the junit XML is already written (`--output-junit junit-regression.xml`) and already published via `dorny/test-reporter`, so a step that diffs the failing set against a checked-in per-platform baseline and fails on anything *new* is small. That buys regression protection at the current pass rate, and the baseline file can only shrink. Not worth building if the interval is short.

## Related

[...]

Issue #2709, opened 2026-08-14
ModuleDialogFactoryTests.ListAllDialogs: dialog count constant is stale (152 vs a 151 ceiling)


dcwhite:
`Modules_Dialog_Factory_Tests` fails in the `mac-gui-regression` nightly:

```
dialog factory size: 152
src/Interface/Modules/Factory/Tests/ModuleDialogFactoryTest.cc:58: Failure
Expected: (descMap.size()) <= (NUM_DIALOGS + EXPECTED_RANGE), actual: 152 vs 151
```

The test bounds the dialog count with two hand-maintained constants, and the comment says as much:

```cpp
const int NUM_DIALOGS = 146;
const int EXPECTED_RANGE = 5;   // Require updating these numbers every few modules
```

152 dialogs now exist, the ceiling is 151, so dialogs have been added since the last bump. Fix is bumping `NUM_DIALOGS`.

Two things worth deciding while touching it:

- **It is not a mac bug.** The dialog factory only exists in GUI builds, and `mac-gui-regression` is the only nightly job that builds the GUI *and* runs tests — the linux and windows regression jobs are headless, so they never run this test. Any GUI build shows it.
- **The test as written will keep doing this.** It fails every time someone adds a sixth dialog, and the failure says nothing about what is wrong. If the intent is "the factory is populated," a lower bound alone would do it; the upper bound mostly generates this recurring chore. `ModuleTraitHasUIMatchesDialogFactory` (right below it, passing) is the test that actually checks dialogs against modules.

Blocks making the macOS test steps blocking (see the CI ratchet issue).

Issue #2708, opened 2026-08-14
Core_Python_Tests: 7 FieldConversionTests crash with an access violation on Windows


dcwhite:
`Core_Python_Tests` fails on `windows-headless-regression` every night ([2026-08-13](https://github.com/SCIInstitute/SCIRun/actions/runs/31688134956), and the night before). Seven `FieldConversionTests` cases die with an access violation, not an assertion:

```
[ RUN      ] FieldConversionTests.LoopThroughFieldFiles
unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.
Stack trace:

[  FAILED  ] FieldConversionTests.LoopThroughFieldFiles (3 ms)
```

`0xc0000005` is a null/bad-pointer dereference. Failing: `LoopThroughFieldFiles`, `RejectsEmptyDictionary`, `RejectsIncompatibleDictionary`, `RoundTripLatVol`, and three more in the same suite.

## The split is the clue

The sibling cases that avoid the Python path all pass, in the same binary, milliseconds apart:

```
[       OK ] FieldConversionTests.RoundTripLatVolUsingJustMatlabConversion (32 ms)
[       OK ] FieldConversionTests.RoundTripTetVolUsingJustMatlabConversion (1 ms)
[       OK ] FieldConversionTests.RoundTripTriSurfUsingJustMatlabConversion (0 ms)
[       OK ] FieldConversionTests.LoopThroughFieldFilesMatlabOnly (7 ms)
[ RUN      ] FieldConversionTests.LoopThroughFieldFiles
unknown file: error: SEH exception ...
```

Every `*UsingJustMatlabConversion` / `*MatlabOnly` variant passes; every case that goes through Python crashes. Also note `RejectsEmptyDictionary` crashes in 0 ms — it dies before doing any real work, which points at the Python object/interpreter the fixture hands it being null rather than at the conversion logic.

That may share a root cause with #2699, where `InterfaceWithPython` scripts cannot see the `scirun_*` builtins in headless runs — both look like the embedded interpreter not being brought up in a non-GUI context. Worth looking at them together, though the symptoms differ (crash here, `NameError` there).

## Reproducing

```bash
cd bin/SCIRun
ctest --output-on-failure -C Release -R Core_Python_Tests
```

Windows only so far — linux passes `Core_Python_Tests` while failing the #2699 networks, so the two platforms fail differently in the same area.

Blocks making the Windows test steps blocking (see the CI ratchet issue).

Issue #669, opened 2014-10-23
Module conversion: ShowColorMap, part 1


dcwhite:
Post-new renderer merge.

dcwhite:
@cbrightsci @benjaminlarson @brigb123 I'll let you guys figure out how to divy this one up. Ben should add the module and make the GUI and leave the guts macroed out.

benjaminlarson:
@dcwhite @cbrightsci @brigb123 Is anyone waiting on this? I have the GUI components done/implemented, but haven't figured out the text color option yet. I can PR now if you're ready, or wait till tomorrow.

dcwhite:
@benjaminlarson Just do PR now, I'll merge it

dcwhite:
@cbrightsci @brigb123 GUI and module skeleton are ready to fill in

brigb123:
The scale draws now (no text) in a3479fde84ab96aab51a5d31948d9282a4985711. Color map needs to be forwarded exactly using the Color map object.

brigb123:
3 color maps applied to faces and match scale bar. see 41d84fb28b528855ab81c9457583b2f6c942fd97

dcwhite:
We can plan on closing this one after the test lunch tomorrow, then open a "Part 2" for the text labels and other features. A good start.

PR #2712, opened 2026-08-14
Tests: scale the streamline length tolerance for the adaptive integrators (#2710)

### LINKED ISSUE #2710: GenerateStreamLinesTests.ManySeedsMultithreadedStreamlineLength: total length off by 0.22 against a 0.1 tolerance
`Algorithms_Field_Tests` fails on `linux-headless-regression` every night ([2026-08-13](https://github.com/SCIInstitute/SCIRun/actions/runs/31688134956) and the night before) — one case:

```
src/Core/Algorithms/Field/Tests/GenerateStreamLinesTests.cc:300: Failure
The difference between max and meshOutputByMethodTotalLength[method].second is 0.21573622395288794,
which exceeds 1e-1, where
max evaluates to 127.37573622395288,
meshOutputByMethodTotalLength[method].second evaluates to 127.16, and
1e-1 evaluates to 0.10000000000000001.
```

So the computed total streamline length is 127.3757 against an expected 127.16 — off by 0.216, about 0.17%, against a tolerance of 0.1.

## What to work out first

This is the **multithreaded** case, and the log shows the work split four ways on the runner:

```
GenerateStreamLinesAlgoP proc 0, start 0, end 250
GenerateStreamLinesAlgoP proc 1, start 250, end 500
GenerateStreamLinesAlgoP proc 2, start 500, end 750
GenerateStreamLinesAlgoP proc 3, start 750, end 1000
```

Two very different explanations, and it is worth deciding which before touching the tolerance:

1. **Benign** — accumulation order across threads shifts the sum, the expected value was recorded on a different thread count, and the tolerance is simply too tight for a 127-unit total. Fix is the tolerance (or making the expectation thread-count independent).
2. **Not benign** — the parallel path produces materially different streamlines than the serial one. 0.17% is a lot for pure float reassociation on a sum of this size, so this cannot be dismissed without checking.

The cheap discriminator: run it pinned to one thread and against several thread counts. If the total tracks the thread count, it is (2), and the tolerance is hiding a real difference in the parallel implementation.

Because the linux headless job is the *only* test job in the PR tier, this single test is what stands between that tier and a required check that can actually fail — see the CI ratchet issue.

```bash
cd bin/SCIRun
ctest --output-on-failure -R Algorithms_Field_Tests
```

dcwhite:
Fixes #2710.

`GenerateStreamLinesTests.ManySeedsMultithreadedStreamlineLength` fails on linux every night while passing on mac and Windows:

```
src/Core/Algorithms/Field/Tests/GenerateStreamLinesTests.cc:300: Failure
The difference between max and meshOutputByMethodTotalLength[method].second is 0.21573622395288794,
which exceeds 1e-1, where
max evaluates to 127.37573622395288,
meshOutputByMethodTotalLength[method].second evaluates to 127.16
```

## It is not a threading bug, despite the test name

#2710 asked for that question to be settled before touching the tolerance, since 0.17% is large for float reassociation. Two pieces of evidence, both from nightly logs:

- **Bit-identical between runs on linux.** `127.37573622395288` in the 2026-08-13 nightly and again in 2026-08-12. A thread-scheduling effect would not reproduce to 17 digits.
- **The same multithreaded code passes on mac and Windows.** All three jobs run with `UseMultithreading, true`; only linux is out of band.

Deterministic per platform, different between platforms — which is exactly the effect already documented a few lines above this assertion for the *point counts*:

> the exact number of integration substeps and colinear-point cleanup depends on floating-point rounding, so hardcoding exact counts is not portable and has repeatedly broken cross-platform CI

`RungeKuttaFehlberg` and `CellWalk` pick their substep count adaptively, so rounding changes how many steps a streamline takes and shifts the accumulated length. The fixed-step integrators do not have that freedom and reproduce exactly.

## The change

A 1% relative tolerance for the two adaptive methods; the flat `1e-1` stays for the fixed-step ones.

- ~6x headroom over the observed 0.17% drift, and still two orders of magnitude tighter than the 2% the point counts already get.
- Deliberately **not** a blanket relative tolerance: 1% of `AdamsBashforth`'s 6260.01 would have loosened that bound from 0.1 to 62.6, weakening a guard that has held on every platform.

Verified to compile with the exact compile command from a configured build (`-fsyntax-only`, clean). The behavioural check is linux CI — mac passes at the old, tighter bound, so a local mac run proves nothing here.

## Why this one first

`Algorithms_Field_Tests` is the only failing test in `linux-unit-tests`, and that is the only test job in the PR tier — so this is what stands between the required check and one that can actually fail. Step 1 of #2711.

Note the gate flip itself is deliberately **not** in this PR: `continue-on-error` should come off only after this lands and linux unit tests are observed green, and it needs the reusable workflow parameterised so it applies to linux alone (Windows unit is blocked on #2708, macOS on #2709).


basisunus:
I ran it on Ubuntu. There is a segfault. Not sure what happened.
basisunus@AHPC:~/Documents/SCIRun/build/SCIRun$ ctest -I 669,669
Test project /home/basisunus/Documents/SCIRun/build/SCIRun
    Start 669: .Test.ImportNetwork.Other.v4nets.example_nets.SCIRun.Modules.Visualization.GenerateStreamLines-cellwalk
1/1 Test #669: .Test.ImportNetwork.Other.v4nets.example_nets.SCIRun.Modules.Visualization.GenerateStreamLines-cellwalk ...***Exception: SegFault  0.43 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.45 sec

The following tests FAILED:
	669 - .Test.ImportNetwork.Other.v4nets.example_nets.SCIRun.Modules.Visualization.GenerateStreamLines-cellwalk (SEGFAULT)
Errors while running CTest
Output from these tests are in: /home/basisunus/Documents/SCIRun/build/SCIRun/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.

basisunus:
My Ubuntu is running on the old computer. There might be some problem because the terminal often becomes unresponsive at building or running tests, although it seems to be still running in the background.

dcwhite:
> I ran it on Ubuntu. There is a segfault. Not sure what happened. basisunus@AHPC:~/Documents/SCIRun/build/SCIRun$ ctest -I 669,669 Test project /home/basisunus/Documents/SCIRun/build/SCIRun Start 669: .Test.ImportNetwork.Other.v4nets.example_nets.SCIRun.Modules.Visualization.GenerateStreamLines-cellwalk 1/1 Test #669: .Test.ImportNetwork.Other.v4nets.example_nets.SCIRun.Modules.Visualization.GenerateStreamLines-cellwalk ...***Exception: SegFault 0.43 sec
> 
> 0% tests passed, 1 tests failed out of 1
> 
> Total Test time (real) = 0.45 sec
> 
> The following tests FAILED: 669 - .Test.ImportNetwork.Other.v4nets.example_nets.SCIRun.Modules.Visualization.GenerateStreamLines-cellwalk (SEGFAULT) Errors while running CTest Output from these tests are in: /home/basisunus/Documents/SCIRun/build/SCIRun/Testing/Temporary/LastTest.log Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.

That looks like the wrong test, can you just run ```./Algorithms_Field_Tests --gtest_filter='*GenerateStreamLinesTests*'```

basisunus (approved):
Tests passed on Windows and Ubuntu.
I was hoping to see where the divergence happens, which can be from a simple function like sqrt, and the errors can accumulate and lead to different control paths. But I guess it might take too much effort.
The real work is moving libxml2 out so that I can run it on Ubuntu 26.04.

PR #2713, opened 2026-08-15
Nightly: name installers for what they run on, verify OS floor, generate release notes

dcwhite:
Three improvements to the nightly release, all downstream of the same gap: nothing recorded what a published installer actually was.

## 1. Installer naming

Assets were named for the job that built them, so a downloader could not tell arch, OS floor, Qt or Python apart — and `NP` meant nothing to anyone.

| Before | After |
|---|---|
| `SCIRunMacInstaller.pkg` | `SCIRun-nightly-macos26-arm64-qt6-python.pkg` |
| `SCIRunMacInstaller-26.pkg` | *(duplicate of the above — no longer published)* |
| `SCIRunMacInstaller-26-intel.pkg` | `SCIRun-nightly-macos26-x86_64-qt6-python.pkg` |
| `SCIRunMacNPInstaller.pkg` | `SCIRun-nightly-macos26-arm64-qt6-nopython.pkg` |
| `SCIRunWindowsInstaller_qt5.exe` | `SCIRun-nightly-windows-x64-qt5-nopython.exe` |
| `SCIRunWindowsPythonInstaller_qt6.exe` | `SCIRun-nightly-windows-x64-qt6-python.exe` |
| — | `SCIRun-nightly-macos14-arm64-qt6-python.pkg` *(new, see below)* |

Stable releases use the same slug with the tag baked in: `SCIRun-v5.5.0-macos14-arm64-qt6-python.pkg`.

`macOS-latest` resolves to `macos-26` today, so `mac-gui` and `mac-gui-26` were building and publishing the same installer twice (107191395 vs 107191813 bytes). `mac-gui` keeps building — it goes red early when GitHub rolls the label forward — but no longer publishes.

## 2. OS verification

`scripts/ci/build_info.py` reads `LC_BUILD_VERSION` and `lipo -archs` back out of the packaged binary and writes `build-info.json` into the artifact. A new `max-os-floor` input warns when the measured floor is higher than intended — warns rather than fails, because the job cannot fix the runner image it was handed.

The check immediately flags something real. Pulled off the current live nightly release:

```console
$ vtool -show-build SCIRun.app/Contents/MacOS/SCIRun
     minos 26.0
       sdk 26.5
$ lipo -archs SCIRun.app/Contents/MacOS/SCIRun
arm64
```

**Every published mac installer was macOS 26+, arm64**, under the most generic and most-clickable name on the release. dyld refuses to launch it on macOS 14 or 15. Bundled Qt 6.10.2 is `minos 13.0`, so Qt is not the constraint — clang defaults the deployment target to the host OS version and nothing overrides it. That is exactly #2565.

Pinning the deployment target is left to #2565. What this PR does instead is **publish `mac-gui-14-arm`**, which already built a macOS 14-compatible GUI on the oldest image, ran the smoke test on it, and then threw the installer away. Zero build risk, and it gives users a mac installer that runs today.

## 3. Changelog

The release body was a fixed two-line blurb. `scripts/ci/nightly_release_notes.py` now generates it per run: a download table carrying each installer's OS floor, arch, Qt and Python, plus the commits since the previous nightly (capped at 50, with a compare link).

`mac.yml` and `windows.yml` publish to the *same* release on separate schedules, so it merges rather than overwrites — each run replaces only its own platform's block, and the changelog is recut only when the commit has actually moved. That is tracked with a `<!-- nightly-sha: -->` marker rather than the `nightly` tag, because the first publisher moves that tag out from under the second, which would otherwise produce an empty changelog.

<details>
<summary>Sample generated body</summary>

```markdown
Automated nightly build of `master`. Asset URLs are stable — each run overwrites them in place.

<!-- nightly-sha: a69b549734103a8b1d2d757cb12a0a17354e68e6 -->

## Downloads

<!-- assets:macos -->
| Installer | Arch | Runs on | Qt | Python |
|---|---|---|---|---|
| `SCIRun-nightly-macos14-arm64-qt6-python.pkg` | arm64 | macOS 14.5 or newer | 6.10.2 | yes |
| `SCIRun-nightly-macos26-arm64-qt6-python.pkg` | arm64 | macOS 26.0 or newer ⚠️ | 6.10.2 | yes |
Built from `a69b549`.

⚠️ These are built on the newest runner image and inherit its minimum OS, so they
will not launch on older macOS releases that SCIRun otherwise supports.
<!-- /assets:macos -->
```
</details>

## Also

The VS 2026 Windows build is `allow-failure` and was not in the publisher's `needs`, but *did* match the `*Installer*` artifact glob — so it reached the release or not depending on which job finished first. Renamed out of the `installer-*` namespace.

## Testing

Both scripts run offline and are testable without CI, following the `slack_test_summary.py` precedent:

- `build_info.py` verified against the real bundle extracted from the live nightly `.pkg` — correctly reports `minos 26.0` / arm64 and raises the warning.
- `nightly_release_notes.py` verified through a three-run sequence: first publish, second publisher on the same commit (must carry the other platform's table across and preserve the changelog), and a new commit (must recut against real `git log`).
- The publisher's rename shell executed against a simulated `dist/`, including the missing-`build-info.json` fallback path.
- All four workflow YAMLs parse.

Not verified: the workflow paths themselves. `publish-release.yml` only executes from the default branch, so this needs a `workflow_dispatch` run on master after merge.

Refs #2398, #2565.


dcwhite:
@jessdtate This should be next for approval

basisunus (approved):

PR #2714, opened 2026-08-15
Add VMesh::get_all_neighbors for delem inputs (#2434)

### LINKED ISSUE #2434: delem neighbors
vmesh and the mesh  classes have a get_neighbor and get_neighbors function for the delem inputs, but each requires on of the element neighbors.  There should be a function to get all the neighbors, which would be very similar to these, but without the checks.

dcwhite:
Fixes #2434.

## What

`get_neighbor()` and `get_neighbors()` both take a delem, but they also require one of the elements sitting on it — so you can ask *"who is on the other side of this face from me"*, but never *"who is on this face"*. That makes them awkward to use when you are walking a mesh's delems rather than its elems.

Adds the delem-only form:

```cpp
virtual void get_all_neighbors(Elem::array_type& elems, DElem::index_type delem) const;
```

## The implementation is thin, on purpose

While implementing this I found the check-free lookup already exists per mesh type — it is `get_elems(elems, delem)`, backed by `get_cells_from_face` / `get_faces_from_edge` / etc. It is just filed under the topology API and never named or documented as a neighbor call, which is presumably why it did not turn up when the issue was written.

So the base implementation forwards there rather than duplicating the traversal into thirteen mesh classes, all of which would be one-line forwards to the same primitive. Subclasses can still override if a mesh ever needs something different. Happy to expand it per-mesh if reviewers would rather see it spelled out.

## Point cloud gap

`VPointCloudMesh` had no `get_elems(..., DElem)` at all, so `get_all_neighbors` would have hit the base `ASSERTFAIL` there. Added it — node, elem and delem are all the same index in a point cloud.

## Tests

New `VMeshNeighborTests.cc`, 9 cases across TetVol, TriSurf, LatVol and PointCloud.

The main one is a property check run over every delem of a mesh: the returned set has no duplicates, every element in it genuinely lists that delem, and `get_neighbors(elems, elem, delem)` returns exactly the set minus the element you asked from. The rest pin counts on known geometry — the 6-tet cube splits into 12 boundary and 6 interior faces, the closed TriSurf cube has zero boundary edges, a 2x2x2 LatVol gives 24 boundary and 12 interior.

## Verification

Full headless build clean. All 54 unit-test binaries pass, including `Core_Datatypes_Legacy_Field_Tests`.

## Noted, not fixed

- `VMesh::size(DElem::size_type&)` returns 0 for point clouds, so a point cloud reports no delems even though delem indices are answerable. Pre-existing; the test iterates elem count rather than working around it.
- `src/Core/Datatypes/Mesh/VirtualMeshFactory.cc` has a parallel `VirtualMesh` neighbor interface, untouched here — it reads as an unfinished rewrite and is outside what #2434 asks for.

### DIFF
--- a/src/Core/Datatypes/Legacy/Field/Tests/VMeshNeighborTests.cc
+++ b/src/Core/Datatypes/Legacy/Field/Tests/VMeshNeighborTests.cc
@@ -1,0 +1,251 @@
+/*
+   For more information, please see: http://software.sci.utah.edu
+
+   The MIT License
+
+   Copyright (c) 2020 Scientific Computing and Imaging Institute,
+   University of Utah.
+
+   Permission is hereby granted, free of charge, to any person obtaining a
+   copy of this software and associated documentation files (the "Software"),
+   to deal in the Software without restriction, including without limitation
+   the rights to use, copy, modify, merge, publish, distribute, sublicense,
+   and/or sell copies of the Software, and to permit persons to whom the
+   Software is furnished to do so, subject to the following conditions:
+
+   The above copyright notice and this permission notice shall be included
+   in all copies or substantial portions of the Software.
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+   OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+   THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+   DEALINGS IN THE SOFTWARE.
+*/
+
+/// Tests for VMesh::get_all_neighbors (issue #2434): every element on a delem,
+/// without having to hand one of them in.
+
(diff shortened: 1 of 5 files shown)


dcwhite:
@jessdtate will take a closer look

src/Core/Datatypes/Legacy/Field/Tests/VMeshNeighborTests.cc
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  /// The properties get_all_neighbors has to hold for any mesh type:
+  /// every returned element really touches the delem, no duplicates, and the
+  /// result agrees with what get_neighbor/get_neighbors say once you pick one
+  /// of the elements to stand in as "from".
+  void checkDelemNeighborsAgree(VMesh* mesh)
+  {
+    VMesh::DElem::size_type numDelems;
+    mesh->size(numDelems);
```
jessdtate: I don't see how this will work on all mesh types.  I haven't checked them all, but pointcloudmesh doesn't have a func for delem::size_type, and therefore defaults to 0 (per VMesh.cc ln 77).

src/Core/Datatypes/Legacy/Field/VMesh.cc
```diff
@@ -734,6 +734,14 @@ VMesh::get_neighbors(Node::array_type&, Node::index_type) const
   ASSERTFAIL("VMesh interface: get_neighbors(Node::index_type,Node::index_type) has not been implemented");
 }
 
+void
+VMesh::get_all_neighbors(Elem::array_type& elems, DElem::index_type delem) const
```
jessdtate: I'm not sure introducing the delem in the function call accurately addresses the original request.  I'll look into this some more

jessdtate: more to the point, this function is just a wrapper for another function and it points to two options: it does not address the original request, or the original request was already implemented.

PR #2720, opened 2026-08-18
Harden ES/Spire render pipeline against uninitialized members

dcwhite:
Audits constructors across the SCIRun ES/Render glue code and the vendored spire
libraries (`es-render`/`es-general`) for members left uninitialized on default
construction — the same pattern behind the "red ViewScene" bug (#881) and its
earlier point fixes (edf4772f8, 3818e21d1).

These are real uninitialized-read paths, not a style sweep:

- `StaticWorldLight` set only `lightDir[0]`, leaving 3 of 4 light directions as garbage.
- `AABB2D` had no constructor at all.
- `StaticCameraData::nearTopLeft`/`farTopLeft` are read before `setProjection()` ever runs.
- `WidgetUpdateService::buttonPushed_` could be read before any button press.
- `VecUniform`/`MatUniform` never initialized their actual uniform value.
- `DebugRender`'s first/last point could be read before any line was added.
- Two local `glm::mat4` variables (in `RenderClickBox2DSys` and `getTexQuadTransform`)
  fed partially-uninitialized transforms into real draw calls.

The rest are members hardened to a defined default so the same class of bug can't
resurface through them.

Also adds `.clang-tidy` configs (`cppcoreguidelines-pro-type-member-init`,
`cppcoreguidelines-init-variables`) scoped to `src/Externals/spire/` and
`src/Interface/Modules/Render/ES/`, so members added later get flagged instead of
relying on manual constructor-list bookkeeping.

Garbage in the render path shows up as intermittent, platform-dependent visual
failures rather than clean crashes, which is why this is worth doing as an audit
rather than one bug at a time.

## Notes for review

- No file in this branch is touched by any other open PR.
- Most of the diff is one-line default initializers in vendored spire headers; the
  substantive changes are the bullets above.


src/Interface/Modules/Render/ES/Core.cc
```diff
@@ -43,7 +43,7 @@
 namespace SCIRun {
 namespace Render {
 
-ESCore::ESCore() : mCurrentTime(0.0f), r_(0.0f), g_(0.0f), b_(0.0f), a_(0.0f)
+ESCore::ESCore()
```
dcwhite, line 46: This looks scary (introducing the red view scene back), but the initializers are all inline in the header now, so all constructors are safe (if a new one is added)

dcwhite:
Needs testing on Windows and Mac.

basisunus (approved):
I didn't see any issues using a simple test.

Issue #1746, opened 2018-02-12
Modules to be converted--Entire list


dcwhite:
TODO

jansonap:
I'd like to add "WriteBundle" to this list under "List: Bundles" and request that full bundle support/conversion be moved up the list in priority of implementation (if that ranking exists)

dcwhite:
Will do. At some point, I noted that WriteBundle was mistakenly missing from this list. Thanks for the reminder.

dcwhite:
@jessdtate Intern work?

jessdtate:
Yeah, this is something they can tackle for sure.  Perhaps when QT5 is running smoother.

dcwhite:
**Analysis (2026-06-12)**

Master checklist tracking module conversion to .module JSON config files (related to issue #101). Active work is on branch `module-config-cleanup`. 

Per current memory: 84 modules remained as of batch 4; next up is BuildSurfaceLaplacianMatrix. This issue is the tracking vehicle for that work — no separate action needed here.

dcwhite:
**Data point for this list:** a full v4.7→v5 parity scan is now in #2615, weighted by usage across the 434 shipped v4 example networks.

Short version: raw coverage is 33%, but **155 missing and 87 dormant modules appear in zero shipped v4 networks**. The usage-weighted gap is **69 modules** — 38 missing, 31 dormant — which is a very different planning target than the raw counts.

Two pieces broken out: #2614 (v4 volume-rendering pipeline, dormant, the largest coherent gap) and #2613 (parameter-level parity audit — the current comparison is name-level only, so a module ported with fewer options still counts as "present").

Issue #1443, opened 2016-05-23
WriteField: Add g3d as saveable object type


cbrightsci:
Look at uvfconverter in iv3d to get algorithm for this.

dcwhite:
Since this will depend on and adapt types from Duality/G3D.h, the most direct way to implement this might be to add Duality_Client as an external project.

johaenns:
An example script for my current way to export obj's with color and transparency from SCIRun so that they can be converted to g3d's using uvfconvert

[python-obj.tar.gz](https://github.com/SCIInstitute/SCIRun/files/542772/python-obj.tar.gz)

g3d's can be viewed in ImageVis3d by adding a geometry object to a (dummy) volume image, though there might be some rendering issues.

Dummy volume image:

[dummy-uvf.tar.gz](https://github.com/SCIInstitute/SCIRun/files/542783/dummy-uvf.tar.gz)

dcwhite:
@johaenns Do you still need this?

dcwhite:
@jessdtate Beta or post-beta?

jessdtate:
I think they still need this, but I'm not sure how badly. I think post-beta will be fine

dcwhite:
**Triage 2026-06:** New module: `WriteG3D` for ImageVis3D mobile — the `uvfconverter` algorithm from ImageVis3D would need to be integrated. ImageVis3D mobile may no longer be actively supported. Suggest checking if ImageVis3D mobile is still a target platform before investing in this module.

dcwhite:
**Status check, 2026-08:** this is still open and still wanted (per @jessdtate above, post-beta was fine), but it's worth recording how partial the current state is, because the code in the tree doesn't match what the module name implies.

What exists today, from @cbrightsci's `ad_g3d` branch merged in #1780:

- `WriteG3D` is built and live, but it does **not** write G3D. `WriteG3D::call_exporter` delegates to `ObjToFieldReader::write()`, so it emits OBJ — essentially @johaenns's documented workaround (export OBJ, convert with `uvfconvert`) wired in as a placeholder. The file-type list `"IV3D (*.g3d);;ObjToField (*.obj)"` is hardcoded.
- `WriteG3D.module` sits in `Modules/Factory/Config/TODO/`, matching #1780's note that it was merged without a factory config.
[...]

PR #2721, opened 2026-08-19
Duplicate-code removal: shared-helper extractions and verified dead files

dcwhite:
Duplicate-code removal driven by a CPD sweep. Net **-3,092 / +611** across 12 files.

This is the mergeable subset of #2715, which stays a read-only reference artifact and
should not be merged. Everything here was re-derived against current master rather than
cherry-picked, and the two deletions that turned out to be wrong have been reverted (see
below).

## What's in it

**Structural de-duplication**

- `ArrayMathInterpreter.cc` — extract `setProgramVarFromSink`/`setProgramVarFromSource`,
  collapsing a long chain of near-identical per-type blocks (-604 lines).
- `NrrdData.cc` — extract `ioNrrdData()`, removing a nrrd type switch duplicated between
  read and write paths.
- `NrrdToField.cc` — extract `computeNrrdSpatialParams()`.
- `CARPFiber_Plugin.cc` — `CARPFiber_reader` was a near-verbatim copy of
  `TextToTetVolField_reader`; it now delegates.
- `GetFieldBoundaryAlgo.cc` — the 2-arg `run()` was a copy of the full overload; it now
  delegates.
- `CalculateBundleDifference.cc` — extract `computeQuadrilateralArea()`.
- `HexTricubicHmt.h` — extract `evaluateHexCubicBasis()`, shared by `interpolate` and
  `derivate`, and return the template type `T` rather than `double` so precision follows
  the instantiation.

**Dead file removal** (verified unreferenced, not merely commented out of a build)

- `HexElementsWeights.h` — unused duplicate of `HexElementWeights.h`; header-list entry only.
- Both copies of `TriSurfMeshTests.cc` — one is entirely `#if 0` behind a `/// @todo`, the
  other unbuilt.

## Two deletions reverted

The sweep also proposed deleting `SelectAndSetFieldData.cc`/`3.cc` and the G3D
reader/plugin, on the grounds that both are commented out of their CMakeLists. That
marker means at least three different things in this tree, and in both cases it did not
mean dead:

- `SelectAndSetFieldData*` is a **dormant SCIRun 4 module awaiting its v5 port** under
  #1746. No v5 replacement exists; deleting it discards the source the port works from.
- The G3D reader/plugin is **parked work-in-progress** from #1443, which is still open.
  It holds real G3D I/O, not stubs. Full state of the G3D code is written up on #1443 —
  removing it is a close-the-feature decision, not a cleanup, and it needs an answer on
  whether ImageVis3D mobile is still a target.

Both are restored byte-for-byte here, and both stay commented out of the build, so this
PR does not change what gets compiled.

## Incidental build fix

The G3D commit deleted `G3DToField_Plugin.h` while `IEPluginInit.cc` — which is compiled —
still `#include`d it. The include was not dropped until six commits later, so that window
of the branch never built. Restoring the header and its include closes it, and the branch
now compiles at every commit.

## Notes for review

- No file here is touched by any other open PR.
- No behavior change intended anywhere; the only semantic change is
  `evaluateHexCubicBasis` returning `T` instead of `double`, which is a fix.


jessdtate (approved):
I'm not sure if all of these will be used multiple times (array mathinterpreter functions are probably only for the carp plugin), but I don't think it's a problem either.

Issue #2700, opened 2026-08-13
mac GUI regression: intermittent bus error in the rendering path takes a random handful of ViewScene networks each night


dcwhite:
The `mac-gui-regression` job loses 7-13 ExampleNetwork tests a night to `Bus error`, and **the set changes every night** — this is one flaky crash, not a pile of broken networks.

| night | run | bus-error networks |
|---|---|---|
| 2026-08-13 | [31688134956](https://github.com/SCIInstitute/SCIRun/actions/runs/31688134956) | `createGeometricTransform`, `createLatVol_new`, `mapFromSourceToDestination`, `editMeshBBox2`, `embb2`, `embbSimple`, `scaleBar1`, `vectorMagnitude` (10 failures total / 312) |
| 2026-08-12 | [31584453181](https://github.com/SCIInstitute/SCIRun/actions/runs/31584453181) | `buildFE_1e5_viz`, `createGeometricTransform`, `generateSampleRandomNoWidgets`, `mapFromSourceToDestination`, `setuptdcs_colin27_patchelc`, `vsBars`, `2ObjectsView_new_dynamic`, `ShowAndEditDipolesFieldOutput`, `showMeshBBox`, `vectorMagnitude`, `viewSceneLockViewWidgets`, `widgetFeedbackTest`, plus `readTetVol` (SEGFAULT) (15 failures / 309) |

Only 4 networks are common to both nights. Every network in either list drives ViewScene/widgets; nothing in the non-rendering half of the suite crashes.

## What the crash looks like

Output stops immediately after the renderer comes up — no assert, no message, no partial execution:

```
[warning] Waiting for rendering system initialization....
[warning] Done waiting.
[RENDERER LOG] Unable to find uniform with name: uFogSettings in shader with ID: 17
[RENDERER LOG] Unable to find uniform with name: uFogColor in shader with ID: 17
[RENDERER LOG] Unable to find uniform with name: uFogSettings in shader with ID: 11
[RENDERER LOG] Unable to find uniform with name: uFogColor in shader with ID: 11
<bus error; nothing further>
```

(The missing-uniform lines appear in passing runs too — noise, not the cause.)

## Notes for whoever picks this up

- Regression tests run `ctest -j3`, so three SCIRun GUIs render concurrently on one macOS runner. A random subset dying, drawn from exactly the set of tests that touch the renderer, has the shape of a concurrency or shared-context problem rather than a per-network bug.
[...]

Issue #2733, opened 2026-08-24
Continuous transparency sort leaks an OpenGL index buffer every frame


dcwhite:
`RenderBasicSysTrans::groupExecute` leaks one OpenGL index buffer per frame, per transparent pass, whenever the ViewScene transparency mode is set to **Continuous sort**.

## What happens

[RenderTransBasicSys.cc:288-291](https://github.com/SCIInstitute/SCIRun/blob/master/src/Interface/Modules/Render/ES/systems/RenderTransBasicSys.cc#L288-L291):

```cpp
case RenderState::TransparencySortType::CONTINUOUS_SORT:
{
  iboID = sortObjects(dir, ibo, pass, iboMan);
  break;
}
```

`sortObjects` ends by calling `addIBO` ([line 225](https://github.com/SCIInstitute/SCIRun/blob/master/src/Interface/Modules/Render/ES/systems/RenderTransBasicSys.cc#L225)), which is `glGenBuffers` + `glBufferData` ([line 162](https://github.com/SCIInstitute/SCIRun/blob/master/src/Interface/Modules/Render/ES/systems/RenderTransBasicSys.cc#L162)). The returned id is used for the draw and then dropped on the floor. Nothing frees it, and this runs on every frame the pass is visible.

The `UPDATE_SORT` branch immediately below gets this right — it keeps the id in `sortedObjects` and calls `removeIBO` before replacing it ([line 317](https://github.com/SCIInstitute/SCIRun/blob/master/src/Interface/Modules/Render/ES/systems/RenderTransBasicSys.cc#L317)). That is the only call to `removeIBO` in the tree, so there is no other cleanup path the continuous case could be relying on.

`sortObjects` also takes a `ren::StaticIBOMan` argument that it never names or uses, so the buffers it creates are invisible to the IBO manager as well.

## Impact

Per frame, per transparent pass, at 60 fps: one buffer id plus a full copy of the index buffer uploaded to the driver and never released. A moderately sized transparent surface is a few MB of index data, so this grows fast once someone leaves the mode on.

## Not related to the nightly crashes

[...]

Issue #968, opened 2015-05-11
Transparency problem


dcwhite:
Tried to display two colored latvols, and the purple one "showed through" the green one even when the former was behind the latter.

![viewscene_2015 05 11 131150 868_2](https://cloud.githubusercontent.com/assets/1399383/7572936/53dc10a2-f7df-11e4-9e9c-a3fdd5d048bc.png)
![viewscene_2015 05 11 131144 630_1](https://cloud.githubusercontent.com/assets/1399383/7572938/53de584e-f7df-11e4-9716-7508e5de1576.png)
![viewscene_2015 05 11 131155 680_3](https://cloud.githubusercontent.com/assets/1399383/7572937/53de0790-f7df-11e4-8b40-a3666396ea7c.png)

dcwhite:
First screenshot shows purple latvol behind green one. Second two are correct, purple is on top of green one.

cbrightsci:
@dcwhite could you send me your network? I've created my own without seeing the issue. I'd like to see the exact configuration you are using.

dcwhite:
@cbrightsci It's in github now, just checkout my branch.

dcwhite:
https://github.com/SCIInstitute/SCIRun/commit/768279cffa3fa6c52a8ee512543c0d9e190cc606

cbrightsci:
It appears that this is due to the way the geometry objects are passed to the renderer. The transparency renderer takes the vertices of the objects and reorders them to get the depth. When two showfields are input into a viewscene it makes them into two objects. This means that all the vertices aren't being sorted at the same time. First one showfield data is sorted and rendered then the other is sorted and rendered, thus making the depth funky. This will take a bit of an overhaul on how the data is being handled right now. :(

dcwhite:
Sounds fun. :persevere:

dcwhite:
@jessdtate IBBM

pip010:
Just curious, any chance of supporting depth-pealing and avoid sorting all together?

dcwhite:
We don't have anyone working on deep graphics problems in SCIRun right now, so the answer is "unknown at this time."

dcwhite:
We at last good people working on deep graphics problems, so look for a fix next year maybe...

Haydelj:
[...]

PR #2734, opened 2026-08-24
Bounds-check the transparency sort against the VBO (#2700)

dcwhite:
`RenderBasicSysTrans::sortObjects` walks the index buffer as raw `uint32` triples and indexes straight into the vertex buffer:

```cpp
uint32_t* ibo_buffer = reinterpret_cast<uint32_t*>(pass.front().ibo.data->getBuffer());
size_t num_triangles = pass.front().ibo.data->getBufferSize() / (sizeof(uint32_t) * 3);
...
float* vertex1 = reinterpret_cast<float*>(vbo_buffer + stride_vbo * (ibo_buffer[j * 3]));
```

Nothing checks that the indices are really 32-bit, that the stride is large enough to hold a position, or that the index is within the vertex buffer. When any of that is untrue the read lands at an arbitrary offset past the end of the VBO and the process dies.

That signature is **10 of the 30 macOS crash reports** the nightly has collected since #2707 merged, always at `sortObjects+328` — the vertex read in the depth loop — faulting on a page-aligned address tens of GB from the VBO base. Full triage in [#2700](https://github.com/SCIInstitute/SCIRun/issues/2700#issuecomment-5391313816).

## What this does

Validates index size, vertex stride and every index before dereferencing. On failure it logs the pass name once and returns the unsorted IBO, so the network finishes and the log names the offending geometry instead of the process vanishing.

## What this does not do

This is a guard, not a root-cause fix. Two things to be clear about:

- It covers only the 10 sortObjects crashes. The other 14 are the same out-of-range read happening inside the GL driver (`glDrawElements` → `gleRunVertexSubmitImmediate`) on the non-transparent path. Catching those needs validation once at upload time in `SRInterface::add*BufferObjects` — per-frame index scanning is too expensive. That is a follow-up.
- If the mismatch turns out to be transient — a geometry buffer swapped under an in-flight frame, which `SRInterface` has no locking to prevent — the indices could be valid at check time and stale by the read, and this will not stop the crash. Either way the next nightly is informative: sortObjects crashes stopping means the check caught it and the log names the geometry; them continuing points at a lifetime race and a lock in `SRInterface`.

## Performance

No measurable cost. Three loop-invariant, always-false compares in a loop dominated by three dependent random-access loads; measured delta across 10k–2M triangles ranges from -6% to +6% with no consistent sign, i.e. noise. The loop is 2–25x cheaper than the `std::sort` on the next line regardless. The one-time checks are outside the loop. Note also that with the default `UPDATE_SORT` mode this function runs on first sight of a pass and on significant camera rotation, not per frame.

## Testing

Compiles clean under `-Wall`. Not reproduced locally — the crash has only ever been seen on the CI runners' paravirtualized GL, so the nightly is the test.

Refs #2700.

### DIFF
--- a/src/Interface/Modules/Render/ES/systems/RenderTransBasicSys.cc
+++ b/src/Interface/Modules/Render/ES/systems/RenderTransBasicSys.cc
@@ -63,6 +63,10 @@
 #include "../comp/LightingUniforms.h"
 #include "../comp/ClippingPlaneUniforms.h"
 
+#include <Core/Logging/Log.h>
+
+#include <set>
+
 namespace es = spire;
 namespace shaders = spire;
 using namespace SCIRun::Graphics::Datatypes;
@@ -138,6 +142,15 @@ private:
 
   std::vector<SortedObject> sortedObjects;
 
+  /// Passes already reported as unsortable, so the warning doesn't repeat every frame.
+  std::set<std::string> warnedPasses;
+
+  void warnUnsortable(const std::string& passName, const std::string& reason)
+  {
+    if (warnedPasses.insert(passName).second)
+      logWarning("ViewScene: transparency sort skipped for \"{}\" ({}); drawing unsorted.", passName, reason);
+  }
+
   class DepthIndex {
   public:
     size_t mIndex;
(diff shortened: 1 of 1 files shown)


basisunus (approved):
Checking is good. I guess it should always be 32bit for any GL values. Isn't transparency sort itself a performance bottleneck? I think there is also the issue of rendering multiple transparent objects.

dcwhite:
Thanks. On the three questions:

**32-bit indices** — not guaranteed, that's why the check is there. `SRInterface` draws with `GL_UNSIGNED_BYTE`/`SHORT`/`INT` depending on `ibo.indexSize` (`SRInterface.cc:619-630`, `:890-895`), but `sortObjects` always reinterprets the buffer as `uint32` triples. In practice everything we upload today is 4-byte, so the check should never fire; if it ever does, the log names the pass instead of the process dying.

**Sort cost** — yes, and this PR doesn't change it. Same loop, plus three compares per triangle that measure as noise against the three random-access loads already there; the `std::sort` afterward dominates either way. The real cost problem is #2733 (continuous mode re-sorts and leaks an IBO every frame), which is separate.

**Multiple transparent objects** — also pre-existing: the sort is per-pass, so objects are never depth-ordered against each other (#968). Fixing that is a different design (per-fragment or a global sort across passes), not a bounds check.

Scope here is only the out-of-range read behind 10 of the 30 nightly mac crash reports in #2700.

PR #2738, opened 2026-08-26
CI: move the nightly tag on publish, prune stale-scheme assets

dcwhite:
## Summary
- The `nightly` git tag has been pinned at its very first publish commit (`f691af07`, 2026-08-08) ever since, because `softprops/action-gh-release` only creates a tag the first time it publishes to it and never moves an existing one — every subsequent nightly run refreshed the release body/assets but left `git checkout nightly` pointing at 3-week-old source. Now force-move the tag to `HEAD` on every nightly publish.
- The asset-rename in f818b35e4 switched installer names to a measured slug (`SCIRun-nightly-<slug>.<ext>`), but the upload step only replaces assets with matching filenames, so the old job-named installers (`SCIRunMacInstaller.pkg`, `SCIRunWindowsInstaller_qt5.exe`, ...) never got removed and just sat alongside the new ones. Now delete any existing release asset that shares this run's platform (same file extension) but isn't part of this run's fresh output before publishing.

## Test plan
- [ ] Trigger a nightly run (mac and windows) via `workflow_dispatch` or wait for the schedule, and confirm:
  - [ ] `refs/tags/nightly` moves to the run's commit
  - [ ] the old-scheme `.pkg`/`.exe` assets are gone from the release after the corresponding platform's run
  - [ ] the release still contains all expected current-scheme assets for both platforms (mac run must not touch windows assets and vice versa)

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -40,6 +40,18 @@ jobs:
         with:
           fetch-depth: 0
 
+      # softprops/action-gh-release only creates the `nightly` tag the first
+      # time it publishes to it; once the tag exists, later runs update the
+      # release body/assets but never move the tag itself, so `git checkout
+      # nightly` silently pins you to whatever commit first created it. Force
+      # the ref to HEAD every run. Safe from a trigger loop: mac.yml/windows.yml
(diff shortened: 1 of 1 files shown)


dcwhite:
I rammed this one through, nightly build was messed up

Issue #1, opened 2012-07-27
Save network file from GUI


dcwhite:
Be able to save out a network file using boost.Serialization.  Must contain:
--module description list
--connection list
--module positions in the window
--module state

Issue #2732, opened 2026-08-24
setuptdcs_colin27_patchelc SEGFAULTs in SetupTDCSAlgorithm::create_lhs on the mac nightly


dcwhite:
Split out of #2700, which is scoped to the intermittent crash in the rendering path. This one is not a rendering crash — it dies on a module execution thread, in `SetupTDCS`.

`.Test.ExampleNetwork.setuptdcs_colin27_patchelc_srn` SEGFAULTs on the macOS GUI regression job. Since the crash reports added in #2707 started landing, it has produced a crash report on 2 of the last 4 nightlies:

| night | run | report |
|---|---|---|
| 2026-08-21 | [32467314922](https://github.com/SCIInstitute/SCIRun/actions/runs/32467314922) | `SCIRun_test-2026-08-21-111736.ips` |
| 2026-08-22 | [32563912681](https://github.com/SCIInstitute/SCIRun/actions/runs/32563912681) | `SCIRun_test-2026-08-22-105950.ips` |

Both are in `SetupTDCSAlgorithm::create_lhs`, at two different call sites (`+8164` and `+12376`), reached through the mesh-search code:

```
libCore_Datatypes_Legacy_Field   TriSurfMesh<TriLinearLgn<Point>>::find_closest_node<VNodeIndex<long long>> +872
libCore_Datatypes_Legacy_Field   TriSurfMesh<TriLinearLgn<Point>>::find_closest_node<VNodeIndex<long long>> +748
libAlgorithms_BrainStimulator    SetupTDCSAlgorithm::create_lhs(...) +8164
libAlgorithms_BrainStimulator    SetupTDCSAlgorithm::run(...) +452
libAlgorithms_BrainStimulator    SetupTDCSAlgorithm::run(...) +1028
libModules_BrainStimulator       SetupTDCS::execute() +1820
libDataflow_Network              Module::executeAndHandleExceptions() +60
libDataflow_Network              Module::executeWithSignals() +516
libInterface_Application         ModuleWidget::executeWithSignals() +116
libEngine_Scheduler              DynamicExecutor::ModuleExecutor::run() const +460
                                 _pthread_start
```

Both faults are `SIGSEGV / KERN_INVALID_ADDRESS` on an obviously corrupt pointer — `0xff000000ff424245` and `0xee36369530b63678`, both flagged by the OS as a possible pointer-authentication failure. That is a garbage or freed object being dereferenced, not a plain null check that got missed.

[...]

Issue #2683, opened 2026-08-09
InterfaceWithTetGen: ODR violation from directory-scoped -DWITH_TETGEN


dcwhite:
Latent ODR violation on `master`, present in the **default** build configuration. Found while diagnosing the `WITH_VTK=OFF` breakage on #2667; unrelated to VTK, so filing separately.

## The problem

`InterfaceWithTetGen.h:70` makes the class definition conditional on a preprocessor flag:

```cpp
#ifndef WITH_TETGEN
  DISABLED_WITHOUT_ABOVE_COMPILE_FLAG      // adds an isImplementationDisabled() override
#endif
```

`WITH_TETGEN` is defined with **directory-scoped** `ADD_DEFINITIONS`, in exactly two places:

- `src/Modules/Legacy/Fields/CMakeLists.txt:232` (covers that directory and `Tests/` below it)
- `src/Modules/Factory/CMakeLists.txt:101`

But the header is also included from a third directory that neither call covers:

- `src/Interface/Modules/Fields/InterfaceWithTetGenDialog.cc` / `.h`

So in a `WITH_TETGEN=ON` build — and note `Superbuild.cmake:131` has **`OPTION(WITH_TETGEN "Build Tetgen." ON)`, so this is the default** — two translation units in the same program compile two different definitions of `class InterfaceWithTetGen`:

| TU | sees `WITH_TETGEN` | class has the disable override |
|---|---|---|
| `Modules/Legacy/Fields/InterfaceWithTetGen.cc` | yes | no |
| `Modules/Factory/ModuleFactoryImpl1.cc` | yes | no |
| `Interface/Modules/Fields/InterfaceWithTetGenDialog.cc` | **no** | **yes** |

That's an ODR violation — ill-formed, no diagnostic required.

## Severity: latent, not currently biting

Being honest about the risk, because it argues for a cheap fix rather than an urgent one:

- The override occupies an existing vtable slot (`isImplementationDisabled()` is already virtual on `ModuleInterface`), so object layout is unchanged.
- Which definition wins is decided by whichever TU emits the vtable, via the key function — a non-inline virtual defined in `InterfaceWithTetGen.cc`, which is on the correct side. So the right definition wins today.
- The dialog TU only reaches into the header for `Parameters::*` names; it never instantiates the module or calls `isImplementationDisabled()`.

[...]

PR #2743, opened 2026-09-05
Genelec splitted

jessdtate:
new module GenerateElectrodeFromPoints, Partial port of GenerateElectrode from SR4.  The current version has no widgets features, but will make a wire electrode from a set of control points.  the control points can be created with widget modules.


jessdtate:
turns out the generateElectrode module is in the master branch without a UI, so I duplicated the on I made for GenerateElectrodeWithPoints (the split out version) and reverted the orginal module otherwise.

dcwhite:
Tested on Mac, it looks good! Which old toolkit networks used the module, can we test those?

dcwhite (approved):

dcwhite:
## Review

Reviewed at `5c72922`. The new module + algorithm are structured correctly (proper Module/Algo split, `.module` config, dialog wired through the managers) and the two `FromPoints` example nets do cover both the wire and planar paths. But there's a null-pointer dereference on the only code path the module has, ~2100 lines of files that don't compile or aren't reachable, an unexplained numerical change to `BuildBEMatrixAlgo`, and leftover debug output. Detail below, roughly in severity order.

### Blocking

**1. `impl_` is never initialized — every execute() dereferences a null `shared_ptr`.**
`GenerateElectrodeFromPointsAlgo.cc:50` declares a *local* `GenerateElectrodeFromPointsImpl impl_;` inside the constructor body. It shadows nothing and is destroyed immediately; the member `impl_` (`SharedPointer<...>`, i.e. `std::shared_ptr`, `GenerateElectrodeFromPointsAlgo.h:90`) stays null. Then `GenerateElectrodeFromPointsAlgo.cc:175` does `impl_ -> CalculateSpline(t,p,tt,pp);`.

This "works" only by accident: `GenerateElectrodeFromPointsImpl` has no data members and `CalculateSpline` never touches `this`, so the null `this` is never read. It's still UB, and it will start crashing the moment anyone adds state to the Impl or builds with a sanitizer/UBSan-enabled config.

Worth noting that master's `GenerateElectrode.cc` calls this as a free function (`impl::CalculateSpline`) — the pimpl was added by the port and buys nothing, since the class is stateless. Simplest fix is to drop `GenerateElectrodeFromPointsImpl` and `impl_` entirely and make the two `CalculateSpline` overloads free functions in an anonymous namespace in the .cc (they don't need to be in the header at all).

**2. `runImpl` errors on null input and then dereferences it anyway.**
`GenerateElectrodeFromPointsAlgo.cc:663-671`:
```cpp
if (!input)
{
  error("input field required for GenerateElectrodeFromPointsAlgo");
}
VMesh* smesh = input->vmesh();
```
Missing `return false;`. Also `FieldInformation fis(input);` at line 656 runs *before* the check and would already have dereferenced it.

**3. `GenerateElectrodeFromWidget.cc/.h` (1691 lines) are added but not in any CMakeLists — they never compile.**
They're not referenced from any `CMakeLists.txt` or `.module` config. They're also a renamed WIP copy of the existing `GenerateElectrode` module: same class name `GenerateElectrode`, same include guard `MODULES_LEGACY_FIELDS_GENERATEELECTRODE_H` as `GenerateElectrode.h`, and duplicate `ALGORITHM_PARAMETER_DEF`s. So if anyone ever does add them to the build they'll collide with `GenerateElectrode.cc` on all three counts. Either wire them up (with the collisions resolved) or leave them out of this PR and keep them on a WIP branch — 1691 lines of file that no compiler ever looks at will rot.

**4. `GenerateElectrode.module` now registers `GenerateElectrodeDialog`, but the module still declares `NoAlgoOrUI`.**
`GenerateElectrode.h:74` (unchanged by this PR) is still `MODULE_TRAITS_AND_INFO(ModuleFlags::NoAlgoOrUI)`, and `ModuleTraits.h:62` derives `HasUI` from the `ModuleHasUI` bit — which `NoAlgoOrUI = 0` doesn't have. So no UI button appears and the dialog is unreachable. That makes `GenerateElectrode.ui` (456 lines) + `GenerateElectrodeDialog.cc/.h` (193 lines) dead code as committed. Flipping the trait to `ModuleHasUI` is the one-line fix (the commented-out line 76 is right there), but see #7 below before doing it.

**5. Unrelated numerical change to `BuildBEMatrixAlgo::getOmega`.**
`BuildBEMatrixAlgo.cc:95-113` adds `&& NomGamma/DenomGamma > 0` to the three gamma guards. On the merits this looks like a real fix — `log(NomGamma/DenomGamma)` with a negative ratio produces NaN, and the commit message ("check in BEM algo to avoid nans") says that's the intent. But:
- It silently changes BEM forward-matrix results: where the old code produced NaN, the new code leaves `gamma[i] = 0`, which is a *value*, not a signal. Any net that previously produced NaN now produces a plausible-looking wrong number instead. That deserves its own PR with a reference to the issue it fixes and at least a comment saying why zero is the right fallback.
- It has nothing to do with generating electrodes, and it isn't mentioned in the PR description.
- The three `gammacheck0/1/2` locals are declared far from use and never read after the `if`; folding them back into the `if` conditions keeps the diff to three appended clauses. Indentation on lines 95, 100, 106, 112 is also off (4-space and trailing whitespace), which is most of why CodeFactor is red.

Please split this out.

### Should fix before merge

**6. Debug output left in.** `GenerateElectrodeFromPointsAlgo.cc:731` — `std::cout <<" direction for mesh = "<<direction<<std::endl;` fires on every planar execute. There are also ~15 commented-out `std::cout` lines through `runImpl` and roughly half of `GenerateElectrodeFromPointsDialog.cc` is commented-out code. The commit message says "Leaving debug notes until widgets work", which is fine mid-branch, but this is the cleanup pass. If any of it is a reminder for the widget follow-up, a one-line `// TODO(#xxxx):` beats 60 commented lines.

**7. `GenerateElectrodeDialog` and `GenerateElectrodeFromPointsDialog` are near-verbatim duplicates.** The `.cc` files differ only in the class name and in which lines are commented out; the two `.ui` files differ in three `enabled` booleans and a group-box name. That's ~650 duplicated lines for two modules whose dialogs are meant to converge. Options, roughly in order of preference: (a) hold `GenerateElectrode`'s dialog until the widget work lands and drop it from this PR, (b) give `GenerateElectrode` the *same* dialog class (the dialog factory maps module name → dialog, so two modules can share one, and the extra controls just have to tolerate absent state keys), or (c) share a base class. Duplicating the pair now means every future dialog fix has to be applied twice.

Related: because `GenerateElectrode`'s `setStateDefaults` (master, `GenerateElectrode.cc`) only sets `ElectrodeLength/Thickness/NumberOfControlPoints/Type/Resolution`, its copy of the dialog has `Width`, `Projection` and `UseFieldNodes` commented out but still present in the `.ui`. If the trait from #4 gets flipped, users see three permanently dead controls.

**8. `NumberOfControlPoints` and `UseFieldNodes` are exposed in the UI but the algorithm never reads them.** Both are `addParameter`'d in the algo ctor, pushed through `setStateDefaults`/`execute`, and given a spin box and a check box in the dialog — but `grep` finds zero `get(Parameters::NumberOfControlPoints)` / `UseFieldNodes` in `GenerateElectrodeFromPointsAlgo.cc`. Two controls that visibly do nothing is worse than two controls that aren't there. Either implement them or drop them from the dialog until the widget follow-up needs them (`NumberOfControlPoints` in particular only means something once widgets can add/remove points, which is the commented-out `AddPoint`/`RemovePoint` code).

**9. Duplicate parameter declarations make two headers mutually exclusive.** `GenerateElectrodeFromPointsAlgo.h:56-63` re-declares `ElectrodeLength`, `ElectrodeThickness`, `NumberOfControlPoints`, `ElectrodeType` and `ElectrodeResolution` in `SCIRun::Core::Algorithms::Fields` — the same namespace where `GenerateElectrode.h:43-47` already declares them. `ALGORITHM_PARAMETER_DECL` expands to a `static const AlgorithmParameterName&` at namespace scope, so any TU that includes both headers fails to compile with a redefinition error. The commented-out include at `GenerateSinglePointProbeFromField.h:36` is presumably exactly that landmine going off, and the commit log ("Namespace error still", "modified param names to be unique as a temporary measure") tells the same story.

It builds today only because the factory generator emits module headers and algorithm headers into *separate* translation units (`ModuleFactoryGenerator.cc:143` vs `:230`). That's luck, not design. The matching `ALGORITHM_PARAMETER_DEF`s are also now defined in two libraries — `Modules_Legacy_Fields` and `Core_Algorithms_Legacy_Fields` both export `Parameters::_init_ElectrodeLength()` etc. — which is an ODR violation of the same flavour as #2683.

Cleanest fix: declare the electrode parameters once, in `GenerateElectrodeFromPointsAlgo.h`, have `GenerateElectrode.h` include it, and delete the module-side `DECL`/`DEF` block. Then drop the commented include at `GenerateSinglePointProbeFromField.h:36` rather than leaving it as a tombstone.

### Regression nets

**10. Four new nets in `ExampleNets/regression/Modules/` means four new CTest cases, and only two of them test this PR.**
`Testing/CMakeLists.txt:148` globs that directory, so every `.srn5` dropped in becomes a `.Test.ExampleNetwork.*` test automatically. All four contain `ViewScene`, so all four pick up the `gpu` `RESOURCE_LOCK` (`Testing/CMakeLists.txt:185-190`) and run serialized — the added wall-clock cost is real.

- `generateElectrodeFromPointsExample.srn5` (wire) and `generateElectrodeFromPointsPlanarExample.srn5` (planar) — good, these are the ones that matter.
- `generateElectrodePlanar.srn5` and `generateElectrodeSmall.srn5` both drive the *old* `GenerateElectrode` module, and both have `ElectrodeType = wire`. Despite its name, `generateElectrodePlanar` exercises nothing planar (master's `GenerateElectrode` has no planar path — it's still behind `SCIRUN4_CODE_TO_BE_ENABLED_LATER`) and nothing new in this PR. It's also largely redundant with `generateElectrodeSmall`. I'd drop both, or keep one and rename it so it doesn't claim to cover planar.

Also worth knowing: regression tests don't run in PR CI, only nightly — so these four land untested by the checks on this PR, and any failure shows up in someone else's nightly triage. Please run `scripts/run-regression-tests.sh bin/SCIRun -- -R generateElectrode` locally before merge.

On the plus side: I checked, and none of the four nets embed absolute paths or depend on `SCIRUN_TEST_RESOURCE_DIR` — they all generate their input from `CreateLatVol`. That's the right way to write these.

**11. No unit tests.** `GenerateElectrodeFromPointsAlgo` is a pure function of a field plus parameters — it's about the most testable thing in the codebase. A test that feeds five known points and asserts node/element counts for wire and planar would have caught #1 and #2 immediately, and would pin the spline behaviour before the widget rework touches it.

### Minor / notes

- **`.module` metadata is copy-pasted and now wrong both ways.** `GenerateElectrodeFromPoints.module` says `"status": "Ported module: no UI yet"` — it has a UI. `GenerateElectrode.module` says `"description": "Makes a wire geom with interactive widgets"` alongside `"status": "Ported module: no widgets yet"` — the description describes the goal, not the module. And the description is identical for both modules, which is the one string that has to distinguish them in the module list.
- **Arbitrary 50-node cap** (`GenerateElectrodeFromPointsAlgo.cc:674`) with the message "There are more input nodes than we have arbitrarily decided to allow." That'll read as a bug to a user with 60 control points. Either say what the limit is and why, or make it a parameter.
- **Unused locals** in `runImpl`: `FieldInformation fis(input);` (line 656) and `VMesh::Node::array_type a;` (line 680).
- **`smesh->synchronize(Mesh::ELEM_LOCATE_E)`** (line 671) — only `get_center` on nodes is used afterwards, so this looks unnecessary. Given #2732, spinning up synchronize workers we don't need isn't free.
- **No doc page.** `docs/modules/NewField/` has a `.md` per module plus an `index.rst`; a new user-visible module with a dialog should get one. (`GenerateElectrode` doesn't have one either, so this is a pre-existing gap you'd be extending.)
- **Empty-input UB.** `get_centers` does `t[0]=0` on a vector sized from `p.size()`, and `Make_Mesh_Wire`'s element loop is `idx < final_points.size()-1` on an unsigned type. With a zero-node input field both underflow. The 50-node cap guards the top end but nothing guards the bottom; `CalculateSpline` needs ≥3 points, so an explicit "at least 3 control points required" error in `runImpl` would cover it.
- **`ElectrodeLength` semantics are surprising** — `get_centers` builds `t` as the cumulative arc length of the control points but samples `tt` over `[0, ElectrodeLength]`, so the spline is only followed as far as `ElectrodeLength` regardless of where the points actually are. This is a verbatim port of master's `get_centers` (`GenerateElectrode.cc:1118-1126`), so not a regression from this PR — but it's presumably why the example nets set `ElectrodeLength` to 5.0 and lean on `ScaleFieldMeshAndData`, and it's the kind of thing that belongs in the doc page from the previous bullet.
- Trailing whitespace: 16 lines in `GenerateElectrodeFromPointsAlgo.cc`, 9 in the header, plus the `BuildBEMatrixAlgo.cc` lines. Mixed 2/4-space indentation throughout the new algo (`Make_Mesh_Planar` is 8-space in places).
- `GenerateElectrodeFromPointsImpl(){};` (header line 66) — stray semicolon; moot if #1 is fixed by deleting the class.
- `aa`/`bb` in `Make_Mesh_Planar` are uninitialized if `projection` matches none of the three options. Unreachable through `addOption`, but a `default`/`else` that errors would be cheaper than the compiler warning.
- CI: mac and Windows are both `skipping` on this run, so this has only ever been built on Linux (plus your local mac). Given the new Qt dialog and the `.ui` files, worth getting a mac/Windows build in before merge.

### Summary

The core port is sound and the `FromPoints` split is the right call. What I'd want before merge: fix #1 and #2 (both are one-liners), split #5 into its own PR, and decide what to do with the ~2100 lines that don't build or aren't reachable (#3, #4, #7) — my preference would be to land `GenerateElectrodeFromPoints` alone, cleanly, and keep the `GenerateElectrode`-with-widgets half on the branch until it works. #9 is the one I'd most like to see addressed while the context is fresh, because the commented-out include is going to confuse whoever hits it next.

dcwhite:
@jessdtate I approved it already but you should fix the things above--if you want I can handle it after you merge.

PR #2746, opened 2026-09-12
Fix Windows link error from duplicate GenerateElectrode parameters

dcwhite:
## Summary
- #2743 added `GenerateElectrodeFromPointsAlgo`, which defines the shared `Electrode*` `AlgorithmParameterName`s in `Core_Algorithms_Legacy_Fields`. `GenerateElectrode.cc` still carried its own `ALGORITHM_PARAMETER_DEF` for five of them, so `Modules_Legacy_Fields.dll` exported a second copy of each `_init_*` and `link.exe` failed with LNK2005/LNK1169.
- GCC/Clang silently accept the duplicate, which is why only `windows-gui-qt6` went red — on master at 58a0b7550 ([windows-build run](https://github.com/SCIInstitute/SCIRun/actions/runs/34640038492)) and on every PR that has merged master since, e.g. #2667 ([failed job](https://github.com/SCIInstitute/SCIRun/actions/runs/34640087246/job/103407795977)).
- `GenerateElectrode.h` now includes the algo header for the declarations instead of re-declaring them; the module `.cc` drops its `_DEF`s.

Heads-up: `genelec_widgets` still has the same duplication and adds `GenerateElectrodeFromWidget.cc` with a third copy, so it will need the same treatment.

## Test plan
- [x] `GenerateElectrode.cc` and `GenerateElectrodeDialog.cc` compile with the real build flags; `nm` on the new object shows the `_init_Electrode*` symbols as undefined references (`U`) rather than definitions (`T`)
- [ ] `windows-gui-qt6` goes green on this PR


dcwhite:
The PR tier skips `windows-gui-qt6` for a pure-C++ change (see #2747), so I ran the full Windows matrix by hand against this branch: https://github.com/SCIInstitute/SCIRun/actions/runs/34661244343

- `windows-gui-qt6` (the job that was red on master and on #2667): **green**
- `windows-gui-qt5`, `windows-gui-python-qt5`, `windows-headless`: green
- `windows-gui-python-qt6`: still running at time of writing
- `windows-2025-*-vs2026`: red, but they are `allow-failure` and were equally red on master's last green run (299c6234e) -- unrelated

PR #2748, opened 2026-09-14
Fix macOS minimize/zoom buttons greyed out on floating module dialogs

dariadobrolinski:
## Description

On macOS, when a module's options dialog or the ViewScene render window is undocked into a floating window, the minimize and zoom (maximize) title bar buttons appear but are greyed out and non-functional.

This adds `Qt::WindowMinMaxButtonsHint` alongside `Qt::Window` when setting window flags on these floating dock widgets, so macOS enables the buttons instead of just showing disabled placeholders.

## Changes

- `ModuleOptionsDialogConfiguration.cc`: set `WindowMinMaxButtonsHint` when configuring a module's dockable dialog
- `ModuleWidget.cc`: set `WindowMinMaxButtonsHint` when a module dock widget becomes floating
- `SCIRunMainWindowSlotsPrivate.cc`: set `WindowMinMaxButtonsHint` when the main window's dock widgets become floating

## Testing

Manually verified on macOS that floating module dialog and ViewScene windows now have working minimize/zoom buttons.


dcwhite:
There are problems with this one on Windows/Linux; I'll get a build going and post screenshots. Do not approve.

jessdtate (dismissed):

Issue #2749, opened 2026-09-14
Flag convention: BUILD_OSPRAY is the real option; the documented WITH_OSPRAY is silently ignored


dcwhite:
Part of #2668.

#2668 lists `WITH_OSPRAY` as already complying with the convention. It does not:
the Superbuild option is `BUILD_OSPRAY`, and the documented `WITH_OSPRAY` spelling
is silently ignored.

## What happened

Commit a470ba6e9 (2023-10-25, "Changed flags for Ospray") renamed the Superbuild
options `WITH_OSPRAY` → `BUILD_OSPRAY` and `WITH_OSPRAY_LOCAL` → `PREBUILT_OSPRAY`,
then derived a *normal* variable named `WITH_OSPRAY` from them to forward to the
inner build:

```cmake
# Superbuild/Superbuild.cmake:298
IF(BUILD_OSPRAY OR PREBUILT_OSPRAY)
  SET(WITH_OSPRAY ON)
ELSE()
  SET(WITH_OSPRAY OFF)
ENDIF()
```

That `SET()` shadows any `-DWITH_OSPRAY=ON` the user passed on the command line.
So `WITH_OSPRAY` is a valid inner-build variable that the Superbuild clobbers
before it can matter.

Everything user-facing still says `WITH_OSPRAY`:

| where | says |
|---|---|
| `docs/start/build.md:132` | `cmake -DWITH_OSPRAY=True ../Superbuild/` |
| `CLAUDE.md:42` | `WITH_OSPRAY=ON — OSPRay ray-tracing renderer` |
| `scripts/makeSCIRunRelease.sh:23` | `-DWITH_OSPRAY:BOOL=OFF` |
| `.github/workflows/reusable-build.yml:210,270` | `with-ospray` input → `-DWITH_OSPRAY:BOOL=ON` |

Every one of those is a no-op. Nobody noticed because the only CI jobs that set
`with-ospray: true` — `ccpp.yml` `linux-gui-ospray` and `mac.yml` `mac-gui-ospray`
— are both `if: false`.

## Fix

Under the #2668 rule the dependency knob is `WITH_OSPRAY`. `PREBUILT_OSPRAY` is
not a dependency flag but a *how-to-obtain-it* modifier, so it stays.

- [ ] `OPTION(WITH_OSPRAY "Build with OSPRay support." OFF)` replaces `BUILD_OSPRAY`.
[...]

Issue #2671, opened 2026-08-07
Flag convention: replace BUILD_HEADLESS with WITH_GUI (inverted sense)


dcwhite:
Part of #2668.

`BUILD_HEADLESS` is a negative-sense flag wearing an artifact prefix. Under the
convention it becomes `WITH_GUI`, default `ON`.

This is the only rename in #2668 that changes **meaning** as well as spelling,
so it carries all of the risk and should land on its own.

## Why bother beyond consistency

13 of its 27 CMake sites are `IF(NOT BUILD_HEADLESS)` — 6 in
`src/CMakeLists.txt`, 3 in `Superbuild/Superbuild.cmake`, the rest scattered.
Every one is a double negative that disappears. Reading "if we are not building
without a GUI, add Qwt" is the current state of the Superbuild.

## Blast radius

| where | sites |
|---|---|
| `CMakeLists.txt` + `*.cmake` | 27 (13 of them `NOT`-guarded) |
| `.cc` / `.h` / `.cpp` under `src/` | 8 |
| `.github/`, `scripts/`, `build.sh`, `build.ps1` | 5 files |

The 5 non-source files are `mac.yml`, `reusable-build.yml`,
`scripts/headless_build.sh`, `build.ps1`, `build.sh`.

## The polarity trap

A stale `-DBUILD_HEADLESS=ON` — in a CMake cache, a CI job, or someone's
`CMakeUserPresets.json` — would become an unrecognized variable while `WITH_GUI`
defaults `ON`. The result is a silent GUI build where a headless one was
requested, which on a build machine without Qt fails late and confusingly, and
on one with Qt succeeds and ships the wrong thing.

So this one gets a hard error, not a deprecation warning:

```cmake
if(DEFINED BUILD_HEADLESS)
  message(FATAL_ERROR
    "BUILD_HEADLESS was replaced by WITH_GUI (inverted sense). Use -DWITH_GUI=OFF.")
endif()
```

## Acceptance criteria

- [ ] `OPTION(WITH_GUI "Build the SCIRun GUI." ON)` replaces
      `OPTION(BUILD_HEADLESS ... OFF)`, and the `SCIRUN_CACHE_ARGS` entry that
      forwards it to the inner build is updated.
- [ ] All 13 `IF(NOT BUILD_HEADLESS)` become `IF(WITH_GUI)`; the remaining
      positive uses invert correspondingly.
- [ ] Compile definition updated. Note it is currently pushed two different
[...]

Issue #2670, opened 2026-08-07
Flag convention: rename BUILD_WITH_PYTHON to WITH_PYTHON


dcwhite:
Part of #2668.

`BUILD_WITH_PYTHON` is a dependency flag wearing the retired `BUILD_WITH_`
prefix. Under the convention it becomes `WITH_PYTHON`, matching `WITH_TETGEN`
and `WITH_OSPRAY`, which already gate their externals the same way.

Pure rename — no change in sense, no change in default (`ON`).

## Blast radius

Largest of the three renames, but mechanical:

| where | sites |
|---|---|
| `.cc` / `.h` / `.cpp` under `src/` | 63 |
| `CMakeLists.txt` + `*.cmake` | 37 |
| `.github/`, `scripts/`, `build.sh`, `build.ps1`, `docs/` | 1 |

The C++ sites are `#ifdef BUILD_WITH_PYTHON` guards fed by `ADD_DEFINITIONS`,
so the compile definition has to move in lockstep with the CMake variable or
the guards silently evaluate false and Python support vanishes from a build
that still claims to have it.

## Acceptance criteria

- [ ] `OPTION(WITH_PYTHON ...)` replaces `BUILD_WITH_PYTHON` in
      `Superbuild/Superbuild.cmake`, including the entry in `SCIRUN_CACHE_ARGS`
      that forwards it to the inner build.
- [ ] All 63 C++ guards and the compile definition that feeds them renamed
      together.
- [ ] Deprecation shim: if `BUILD_WITH_PYTHON` is set in the cache, copy it to
      `WITH_PYTHON` and emit a `DEPRECATION` warning. Sense is unchanged, so a
      warning is sufficient here — unlike the headless flip, a stale value still
      means what the user intended.
- [ ] Shim scheduled for removal after one release.
- [ ] A `WITH_PYTHON=OFF` build still configures and links — the non-Python path
      is the one that rots unnoticed, so exercise it deliberately.

Issue #2668, opened 2026-08-07
Pick one convention for CMake feature flags: WITH_ for dependencies, BUILD_ for artifacts


dcwhite:
## Problem

Feature flags in the Superbuild have accumulated seven prefixes with no rule
distinguishing them. Current inventory:

```
BUILD_TESTING, BUILD_SHARED_LIBS                     <- CMake's, not ours
BUILD_HEADLESS, BUILD_BUNDLE,
BUILD_DOCUMENTATION, BUILD_DOXYGEN_DOCUMENTATION
WITH_TETGEN, WITH_OSPRAY
BUILD_WITH_PYTHON, BUILD_WITH_SCIRUN_DATA
ENABLE_COVERAGE, ENABLE_GCOV_DATA_FILES
RUN_UNIT_TESTS, RUN_BASIC_REGRESSION_TESTS, RUN_IMPORT_TESTS
GENERATE_MODULE_FACTORY_CODE, REGENERATE_MODULE_FACTORY_CODE
DOWNLOAD_TOOLKITS
DISABLE_ExternalProject_UPDATE
```

Three of these say the same thing three ways. `WITH_TETGEN` and
`BUILD_WITH_PYTHON` are the same kind of knob — an optional third-party
dependency the Superbuild fetches, builds, and then compiles code against — and
they are spelled differently for no reason anyone can reconstruct.

This is worth settling now rather than later for two reasons, both time-boxed:

- **#2646 (CMakePresets.json)** bakes flag names into a checked-in file that
  becomes the copy-paste source for everyone's local configure. Renaming after
  presets land means editing presets, everyone's `CMakeUserPresets.json`, and
  the docs that quote them.
- **#2476 (`WITH_VTK`)** adds a new dependency flag on the `refactor_add_vtk`
  branch. That branch currently gates nothing behind a flag, so the name is
  still unclaimed, and #2476 itself hedges between `WITH_VTK` and
  `WITH_VTK_RENDERER`. Settling the convention first makes that a decision
  already made rather than a fresh argument.

## Proposed convention

Let the prefix answer *what kind of knob is this*. Note that `BUILD_` is
already half-owned by CMake — `BUILD_TESTING` and `BUILD_SHARED_LIBS` come with
defined meanings we cannot redefine — which constrains the answer more than
anything else does.

**`WITH_<DEP>`** — an optional third-party dependency. The Superbuild fetches
and builds it; the inner build compiles the code that uses it.
[...]

PR #2750, opened 2026-09-14
Document the WITH_/BUILD_ option rule; remove dead BUILD_WITH_SCIRUN_DATA

### LINKED ISSUE #2669: Flag convention: document the WITH_/BUILD_ rule
Part of #2668.

Write the convention down before any renames happen, so the rule exists
independently of the mechanical work and the next new flag does not restart the
discussion.

## The rule

**`WITH_<DEP>`** — an optional third-party dependency. The Superbuild fetches
and builds it; the inner build compiles the code that uses it.
→ `WITH_TETGEN`, `WITH_OSPRAY`, `WITH_PYTHON`, `WITH_GUI`, `WITH_VTK`.

**`BUILD_<THING>`** — an extra artifact this project emits.
→ `BUILD_TESTING`, `BUILD_SHARED_LIBS`, `BUILD_DOCUMENTATION`, `BUILD_BUNDLE`.

**`BUILD_WITH_`** — retired. It is the union of the two rules above.

`ENABLE_` / `RUN_` / `GENERATE_` / `DOWNLOAD_` stay as-is — behavior knobs
rather than dependency or artifact knobs.

The reason the split falls where it does: `BUILD_` is already half-owned by
CMake. `BUILD_TESTING` and `BUILD_SHARED_LIBS` have defined meanings we cannot
redefine, so `BUILD_` has to keep meaning "artifact" and dependencies need the
other prefix, not the reverse.

Not adopted: a `SCIRUN_` prefix on everything for cache-namespace hygiene. It is
the textbook advice and it is wrong here — SCIRun is a terminal application,
never consumed as a subproject, so there is nothing to collide with.

## Acceptance criteria

- [ ] Convention recorded in `CLAUDE.md` alongside the existing build-system
      section.
- [ ] Recorded wherever build options are listed for users, so people
      configuring a build see it.
- [ ] States the rule for *new* flags explicitly — this is the part that has to
      survive after the renames in #2668's other children are done.

Small and independent: this can land first and does not touch any build logic.

### LINKED ISSUE #2672: Remove BUILD_WITH_SCIRUN_DATA: the gforge SVN repo it checks out is gone
Part of #2668.

`BUILD_WITH_SCIRUN_DATA` was slated for a rename to `DOWNLOAD_SCIRUN_DATA`
(it checks out data, so it belongs with `DOWNLOAD_TOOLKITS`, not under `WITH_`).
Checking it before writing that rename showed the option cannot work at all,
so this is a removal instead.

## Evidence

`Superbuild/SCIRunDataExternal.cmake:29` points at:

```
https://gforge.sci.utah.edu/svn/CIBCData
```

- The host still resolves (155.98.58.58, stale DNS) but does not respond —
  connection times out on both the repo path and the site root. GForge was
  decommissioned years ago.
- Last touched April 2017 (`13ae3a6a1`, "Fix typo"). Four commits total, all
  from the same week in 2017.
- `SCI_DATA_DIR` is set by the external and forwarded into the inner build at
  `Superbuild/Superbuild.cmake:332` — and read by **nothing**. Zero consumers
  anywhere in `src/`.
- `FIND_PACKAGE(Subversion)` runs unconditionally at
  `Superbuild/Superbuild.cmake:264` on every configure for every user, solely
  to gate this dead option (it silently forces it `OFF` when svn is missing).
  Subversion appears nowhere else in the build.

Two test files mention `SCIRunData` — `AlgorithmBaseTests.cc:38` and
`LegacyNetworkFileImporterTests.cc:323` — but both are pure string assertions.
The first feeds a fake `E:\scirun\trunk_ref\SCIRunData` into
`Preferences::setDataDirectory()` to test placeholder substitution; the second
asserts a filename string parsed out of a v4 `.srn` file. Neither opens a file.
Nothing in the test suite consumes the actual dataset.

## Scope

- [ ] Delete `Superbuild/SCIRunDataExternal.cmake`.
- [ ] Remove `OPTION(BUILD_WITH_SCIRUN_DATA ...)`, the `ADD_EXTERNAL` call, and
      the `SCI_DATA_DIR` entry in `SCIRUN_CACHE_ARGS`.
- [ ] Remove the now-orphaned `FIND_PACKAGE(Subversion)` block.
- [ ] Update the note in `Superbuild/VERSIONS.cmake:153` that references
      `SCIRunDataExternal.cmake`, and the one in
      `docs/dev_doc/DependencyUpdates.md:104`.
- [ ] No deprecation shim. The option has been non-functional for years; anyone
      setting it was already getting nothing. A `FATAL_ERROR` would be worse
      than silence here, but a one-line `DEPRECATION` warning is acceptable if
      preferred.

## Note

The *data* may still exist somewhere even though the SVN path is gone — the
README offers `https://github.com/CIBC-Internal/SCIRunTestData` as the current
example-data source, with a 4.7-era zip as the legacy fallback. Those legacy
links are themselves dead, which is tracked separately as a docs bug; it does
not block this removal, since the build-side SVN checkout is not the mechanism
anyone would use to ship data today regardless.

dcwhite:
Closes #2669, closes #2672. First PR of #2668.

## What

**#2669 — the rule is written down.** `WITH_<dep>` = optional third-party dependency the Superbuild fetches and builds; `BUILD_<thing>` = extra artifact, consistent with CMake's own `BUILD_TESTING` / `BUILD_SHARED_LIBS`; `BUILD_WITH_` retired. Added to `CLAUDE.md` and to `docs/start/build.md`, which also gains an option table it never had. The three pre-rule names (`BUILD_WITH_PYTHON`, `BUILD_HEADLESS`, `BUILD_OSPRAY`) are documented as they actually are today and marked for #2670 / #2671 / #2749, so no doc claims a name that doesn't work yet.

**#2672 — `BUILD_WITH_SCIRUN_DATA` removed.** It checked out `https://gforge.sci.utah.edu/svn/CIBCData` (decommissioned) into `SCI_DATA_DIR`, which nothing reads, and made `FIND_PACKAGE(Subversion)` run on every configure for every user. `SCIRunDataExternal.cmake`, the `OPTION`, the `ADD_EXTERNAL`, the cache arg, and the Subversion probe are all gone.

**New `Superbuild/DeprecatedFlags.cmake`** with two helpers the follow-up renames share:

- `scirun_renamed_option(OLD NEW)` — copy the old value forward, `DEPRECATION` warning, unset the old cache entry.

- `scirun_removed_option(OLD WHY [FATAL])` — warn and ignore, or hard-error when `FATAL` is passed (for #2671, where the sense inverts).

Included at the top of `Superbuild.cmake` so shims run before anything reads an option — in particular before the early `if(BUILD_WITH_PYTHON)` block at line 38 that #2670 will need to survive.

Also fixed: the `build.md` OSPRay example told people to pass `-DWITH_OSPRAY`, which the Superbuild has silently ignored since a470ba6e9. Corrected to `BUILD_OSPRAY` until #2749 renames it.

## Verification

- Helper exercised in a scratch CMake project: nothing set → silent; renamed → value carried, warns, old entry gone; removed → warns; removed `FATAL` → configure fails.

- Full Superbuild configure (headless, no Qt) with `-DBUILD_WITH_SCIRUN_DATA=ON`: one deprecation warning, then clean. Zero `SCI_DATA_DIR` / `Subversion` entries in the resulting cache or generated files.

- No behavior change for anyone not setting the removed flag.


dcwhite:

Small PR; the whole thing is reviewable in a few minutes. Nothing here changes any build's output.

1. **`Superbuild/DeprecatedFlags.cmake`** (new, 29 lines) — the only logic. Two functions:
   - `scirun_renamed_option(OLD NEW)`: if `OLD` is in the cache, copy it to `NEW` (unless `NEW` was also given), warn, drop `OLD`.
   - `scirun_removed_option(OLD WHY [FATAL])`: warn-and-ignore, or hard-error with `FATAL`.
   Both are `FUNCTION`s so the `SET(... CACHE ... FORCE)` is the only thing that escapes scope. Worth a look: the `IF(NOT DEFINED ${NEW})` guard means an explicit new-name value on the command line beats a stale old-name one in the cache.
2. **`Superbuild/Superbuild.cmake`** — three deletions (option, `FIND_PACKAGE(Subversion)` block, `SCI_DATA_DIR` cache arg) and one `INCLUDE` + `scirun_removed_option` call at the very top. The top placement is deliberate: shims must run before anything reads an option, and #2752 depends on that (the Python block at line ~46 reads `WITH_PYTHON` well before its `OPTION`).
3. **`docs/start/build.md`** — new "Build options" table. The names in it are the ones that exist *at this commit*, not the post-rename ones; #2752 and #2753 each update their row. Check that the defaults column matches `Superbuild.cmake`.
4. **`CLAUDE.md`** — the rule. Skim for whether it's the rule you'd want the next person to follow.

Skip: `VERSIONS.cmake` and `DependencyUpdates.md` are one-line comment edits; `SCIRunDataExternal.cmake` is a straight delete.

dcwhite:
Start here for cmake stuff

jessdtate (approved):

Issue #2751, opened 2026-09-14
OSPRay cannot be built from source: OsprayExternal.cmake depends on TBB/RKCommon/Embree externals deleted in a844a8e63


dcwhite:
Found while doing #2749.

`cmake -DWITH_OSPRAY=ON ../Superbuild` (or `-DBUILD_OSPRAY=ON` before the
rename) fails at configure, on master, before any download happens:

```
CMake Error at .../ExternalProject.cmake:1647 (get_property):
  get_property could not find TARGET rkcommon_external.  Perhaps it has not
  yet been created.
Call Stack (most recent call first):
  OsprayExternal.cmake:37 (ExternalProject_Get_Property)
```

## Cause

a844a8e63 (2026-05-05, "updated based on dan's comments") deleted
`Superbuild/TBBExternal.cmake`, `RKCommonExternal.cmake` and
`EmbreeExternal.cmake` and commented out their `INCLUDE`s in
`Superbuild.cmake`:

```cmake
    #INCLUDE(${SUPERBUILD_DIR}/TBBExternal.cmake)
    #INCLUDE(${SUPERBUILD_DIR}/RKCommonExternal.cmake)
    #INCLUDE(${SUPERBUILD_DIR}/EmbreeExternal.cmake)
    ADD_EXTERNAL(${SUPERBUILD_DIR}/OsprayExternal.cmake Ospray_external)
```

but `OsprayExternal.cmake:29-43` still lists all three as dependencies and
calls `ExternalProject_Get_Property` on each. So the only OSPRay path that
configures is `PREBUILT_OSPRAY=ON` (system `find_package(ospray 2.10.0)`).

Nothing caught it because both ospray CI jobs (`ccpp.yml` `linux-gui-ospray`,
`mac.yml` `mac-gui-ospray`) are `if: false`.

## Options

1. Restore the three externals from a844a8e63^ and re-enable the includes.
   Reinstates the ~pre-May behaviour; cost is three more externals for anyone
   who turns OSPRay on.
2. Drop source building entirely: make `WITH_OSPRAY` mean prebuilt-only,
   delete `OsprayExternal.cmake`, and retire `PREBUILT_OSPRAY`. Simplest, and
   matches what actually works today, but `docs/start/build.md` currently
   documents both paths.

Either way, un-`if: false` one ospray CI job so the chosen path stays
configured. Whether it also has to *build* on CI is a separate cost question
(#1487).

[...]

PR #2752, opened 2026-09-14
Rename BUILD_WITH_PYTHON → WITH_PYTHON and BUILD_OSPRAY → WITH_OSPRAY

### LINKED ISSUE #2670: Flag convention: rename BUILD_WITH_PYTHON to WITH_PYTHON

### LINKED ISSUE #2749: Flag convention: BUILD_OSPRAY is the real option; the documented WITH_OSPRAY is silently ignored

dcwhite:
Closes #2670, closes #2749. Second PR of #2668.

**Stacked on #2750** (needs `DeprecatedFlags.cmake` from it). Base is that branch; retarget to `master` once #2750 merges.

## What

Two pure renames under the `WITH_` = dependency rule. Sense and defaults unchanged. Old names shimmed for one release via `scirun_renamed_option`: value copied forward, `DEPRECATION` warning, old cache entry dropped.

**`BUILD_WITH_PYTHON` → `WITH_PYTHON`**
- 63 C++ `#ifdef` guards, 16 CMake files, `BoostExternal.cmake`'s `-DBUILD_PYTHON` pass-through, `BoostConfig.cmake.in`'s `@…@` substitution, the two `-D` strings in `reusable-build.yml`.
- The compile definition lives in `Superbuild/UsePython.cmake`, which is `COPYONLY`'d into the Python install dir at *Superbuild configure* time — so an existing build dir picks up the new define on its next outer `cmake` run. CI is always fresh.
- The workflow input keeps its `build-with-python` name: it's a wrapper, same as `build.sh --headless` will keep its spelling under #2671.
- `OPTION(WITH_PYTHON)` moves above the "force superbuild Python" block that reads it. That block previously ran ~90 lines before the option was declared, so a first configure with no `-D` skipped it silently.

**`BUILD_OSPRAY` → `WITH_OSPRAY`**
- The Superbuild option had been `BUILD_OSPRAY` since a470ba6e9 while every doc, script and CI input passed `WITH_OSPRAY`, which a plain `SET(WITH_OSPRAY OFF)` then shadowed. That `SET` is gone; the cache variable is forwarded directly.
- `PREBUILT_OSPRAY` stays as the "find it, don't build it" modifier and still implies `WITH_OSPRAY` on its own, as before.

Docs (`CLAUDE.md`, `docs/start/build.md`) updated to the new names.

## Verification

Superbuild configure (scratch dir, headless) under each of: default, `-DBUILD_WITH_PYTHON=OFF`, `-DWITH_PYTHON=OFF`, `-DBUILD_OSPRAY=ON`, `-DWITH_OSPRAY=ON`. Each gives the expected warning (or none) and the expected `WITH_PYTHON` / `WITH_OSPRAY` value in both the outer cache and the generated `SCIRun_external-cache-Release.cmake`.

Inner build configured against already-built externals (initial-cache trick from #2695):
- **ON**: all 104 `flags.make` carry `-DWITH_PYTHON`, zero carry the old name; `PythonInterpreter.cc.o` compiles with its Python symbols present.
- **OFF**: configures clean with no Python targets; `Application.cc` and `NetworkEditorController.cc` (always-built, guarded) compile down their non-Python branches. This is the path the issue asked to exercise deliberately.

## Found along the way

`-DWITH_OSPRAY=ON` without `PREBUILT_OSPRAY` fails at configure **on master too**: a844a8e63 deleted the TBB/RKCommon/Embree externals that `OsprayExternal.cmake` still depends on. Not touched here — filed as #2751.


dcwhite:

51 files, but 47 of them are one-token `BUILD_WITH_PYTHON` → `WITH_PYTHON` substitutions (63 `#ifdef`s, 16 CMake `IF`s). Those are safe to skim: `git diff -w --word-diff` shows nothing but the rename. The four hunks that are *not* mechanical, all in **`Superbuild/Superbuild.cmake`**:

1. **`OPTION(WITH_PYTHON)` moved up ~90 lines** (from the "Configure python" block to just above "Force superbuild Python"). Before this, `if(BUILD_WITH_PYTHON)` at line ~46 ran before the option was declared, so a first configure with no `-D` skipped that block. Behavior change on first configure only, and it's the intended one; the option's default is unchanged.
2. **OSPRay option block** (line ~140): `BUILD_OSPRAY` → `WITH_OSPRAY`. One deliberate semantic change: the old `IF(BUILD_OSPRAY AND PREBUILT_OSPRAY) SEND_ERROR` is gone, because under the new reading `WITH_OSPRAY=ON PREBUILT_OSPRAY=ON` is the valid way to say "use an installed one". `PREBUILT_OSPRAY=ON` alone still implies `WITH_OSPRAY` (the `FORCE` at line ~148) so nobody's existing invocation changes meaning.
3. **OSPRay external block** (line ~284): the `IF(BUILD_OSPRAY OR PREBUILT_OSPRAY) SET(WITH_OSPRAY ON) ELSE() SET(WITH_OSPRAY OFF)` is deleted. That plain `SET()` is the bug from #2749 — it shadowed any `-DWITH_OSPRAY` the user passed. Now the cache variable flows through untouched to `SCIRUN_CACHE_ARGS`.
4. **Two `scirun_renamed_option` calls** at the top, next to #2750's.

Also worth a glance, outside `Superbuild.cmake`:
- `Superbuild/UsePython.cmake:4` — the one `add_definitions` that feeds all 63 `#ifdef`s. Note it's `COPYONLY`'d into the Python install dir at Superbuild configure time, so an existing build dir gets the new define on its next outer `cmake`, not just a rebuild.
- `Superbuild/BoostExternal.cmake:195` — `-DBUILD_PYTHON:BOOL=${WITH_PYTHON}`. The LHS is Boost's own option name and is unchanged; only the RHS renamed.
- `.github/workflows/reusable-build.yml` — the input is still called `build-with-python`; only the `-D` string it emits changed.

Not touched here, tracked in #2751: `WITH_OSPRAY=ON` without `PREBUILT_OSPRAY` fails at configure on master too (deleted externals).

jessdtate (dismissed):

jessdtate (approved):

PR #2753, opened 2026-09-14
Replace BUILD_HEADLESS with WITH_GUI (inverted sense)

### LINKED ISSUE #2671: Flag convention: replace BUILD_HEADLESS with WITH_GUI (inverted sense)

dcwhite:
Closes #2671. Third and last rename of #2668.

**Stacked on #2752** (which is stacked on #2750). Base is B's branch; retarget as the ones below merge.

## What

`BUILD_HEADLESS` (default OFF) → `WITH_GUI` (default ON). The only rename in #2668 that changes **meaning**, so it lands alone.

- 13 `IF(NOT BUILD_HEADLESS)` → `IF(WITH_GUI)`; the positive uses invert correspondingly. No `ELSE()` branches move.
- **Hard error on the old name.** `scirun_removed_option(BUILD_HEADLESS ... FATAL)` fires on any `DEFINED BUILD_HEADLESS`, ON or OFF. A stale `-DBUILD_HEADLESS=ON` would otherwise be silently ignored while `WITH_GUI` defaults ON — a GUI build where a headless one was requested. That's the polarity trap from the issue, and the reason this isn't a `DEPRECATION` like #2752's renames.
- **One compile definition instead of three.** It was pushed by `ADD_DEFINITIONS` in `Superbuild.cmake` (outer project, compiles nothing — dead), `-DBUILD_HEADLESS=1` in `src/CMakeLists.txt` (the live one), and a redundant copy in `src/Main/CMakeLists.txt`. Now: `ADD_DEFINITIONS(-DWITH_GUI=1)` at the same point in `src/CMakeLists.txt`, same directory coverage as before.
- 8 C++ guards flip sense: `#ifdef BUILD_HEADLESS` → `#ifndef WITH_GUI`, `#ifndef` → `#ifdef`, and the `defined(BUILD_HEADLESS) && defined(__APPLE__)` in `PythonInterpreter.cc` → `!defined(WITH_GUI) && ...`.
- Wrapper flags keep their user-facing spelling — `build.sh --headless`, `build.ps1 -Headless`, `scripts/headless_build.sh`, the CI `variant: headless` — and emit `-DWITH_GUI:BOOL=OFF` underneath.
- Docs: `CLAUDE.md`, `docs/start/build.md`, the `mac.yml` comment explaining the macOS-14 GUI job.

## Verification

**Superbuild** (scratch dir):
| args | result |
|---|---|
| `-DWITH_GUI=OFF` | configures with no Qt; `WITH_GUI "OFF"` forwarded; no `Qwt_external` in the Makefile |
| `-DQt_PATH=… -DSCIRUN_QT_MIN_VERSION=6.3.1` | configures; `WITH_GUI "ON"` + `Qt_PATH` + `Qt6Core_DIR` forwarded; Qwt external present |
| `-DBUILD_HEADLESS=ON` | `CMake Error: BUILD_HEADLESS is no longer a build option. It was replaced by WITH_GUI with the opposite sense. Use -DWITH_GUI=OFF …` |
| `-DBUILD_HEADLESS=OFF` | same error — any value is refused |

**Inner build** against already-built externals (initial-cache trick from #2695):
| | `WITH_GUI=ON` | `WITH_GUI=OFF` |
|---|---|---|
| Interface targets | 17 | 0 |
| `flags.make` with `-DWITH_GUI=1` | 102 of 104 (the 2 without are `Externals/spire` and `Externals/pugixml`, added before the definition — same as `BUILD_HEADLESS` before) | 0 of 87 |
| any `BUILD_HEADLESS` left | 0 | 0 |
| `SCIRun` link line mentions Qt | — | no |
| the 6 guarded TUs compile | ✔ | ✔ |
| `Module.cc.o` has `logHeadlessExecutionMessage` | no | yes |

The issue's last acceptance criterion — check the headless **CI artifact**, not the exit code — is the one thing this can't do locally. Once the tier runs, the `linux-headless` job's build log should show no `Interface_*` targets and the binary should have no Qt in `ldd`.


dcwhite:
Headless verification, per the last acceptance criterion on #2671 (check the artifact, not the exit code).

The PR tier has no headless job, so I dispatched the full Linux matrix on this branch: [run 34921431568](https://github.com/SCIInstitute/SCIRun/actions/runs/34921431568). All 7 jobs green (`linux-gui-ospray` skipped as always).

For each of the three headless jobs I pulled the raw build log and checked what was actually built:

| job | `Built target Interface_*` | `Qwt_external` | Qt package lookups | `BUILD_HEADLESS` anywhere | `SCIRun` linked |
|---|---|---|---|---|---|
| linux-headless (x86_64) | 0 | 0 | 0 | 0 | yes |
| linux-headless-arm | 0 | 0 | 0 | 0 | yes |
| linux-2604-headless | 0 | 0 | 0 | 0 | yes |

The configure line shows the expected shape: `build.sh` emits its own `-DWITH_GUI:BOOL=ON`, then CI's `--cmake-args` appends `-DWITH_GUI:BOOL=OFF`, last `-D` wins — same ordering as the old `-DBUILD_HEADLESS=OFF … =ON`, just inverted.

The three GUI jobs on the same run (`linux-gui`, `linux-gui-nonpython`, `linux-gui-arm`, `linux-gui-nonpython-arm`) cover `WITH_GUI=ON` with and without Python on both architectures.

dcwhite:

18 files. The mechanical part is `IF(NOT BUILD_HEADLESS)` → `IF(WITH_GUI)` (13 sites) and `IF(BUILD_HEADLESS)` → `IF(NOT WITH_GUI)` (the rest), plus 8 C++ guards flipping `#ifdef`↔`#ifndef`. No `ELSE()` branch moved. Those hunks are checkable by eye: every `NOT` that vanished on the left reappears as its absence on the right and vice versa.

The hunks that carry actual risk:

1. **The FATAL shim** — `Superbuild/Superbuild.cmake:34-35`. Fires on any `DEFINED BUILD_HEADLESS`, ON *or* OFF. This is the polarity trap from #2671: a stale `-DBUILD_HEADLESS=ON` in a cache, CI line or `CMakeUserPresets.json` would otherwise be an ignored variable while `WITH_GUI` defaults ON — a GUI build where a headless one was asked for, which fails late on a box without Qt or ships the wrong artifact on one with it. Decide whether you're happy with a hard error over a warning; the argument for it is in the issue.
2. **Compile definition collapsed from three sites to one** — `src/CMakeLists.txt:698-701`. Previously: `ADD_DEFINITIONS(-DBUILD_HEADLESS)` in `Superbuild/Superbuild.cmake` (dead — the outer project compiles nothing), `-DBUILD_HEADLESS=1` in `src/CMakeLists.txt` (the live one), and a redundant copy in `src/Main/CMakeLists.txt`. Now just `ADD_DEFINITIONS(-DWITH_GUI=1)` at the same point in `src/CMakeLists.txt`. Same directory coverage as before — everything added after that line, which excludes `Externals/` (spire, pugixml) exactly as it did before. The sense inverts: the define is now present in *GUI* builds, absent in headless.
3. **`src/Core/Python/PythonInterpreter.cc:243`** — `#if defined(BUILD_HEADLESS) && defined(__APPLE__)` → `#if !defined(WITH_GUI) && defined(__APPLE__)`. Only C++ site with a compound condition; worth a second look that the negation landed on the right term.
4. **`Superbuild/PythonExternal.cmake:209`** — the one `IF(BUILD_HEADLESS)` outside `Superbuild.cmake`/`src/`. Chooses between `Python.framework/...` and `Frameworks/Python.framework/...` for the mac module search path; inverted to `IF(NOT WITH_GUI)`.

Wrappers keep their spelling: `build.sh --headless`, `build.ps1 -Headless`, `scripts/headless_build.sh`, CI `variant: headless`. Each now emits `-DWITH_GUI:BOOL=OFF`. In `build.sh` the variable `headless="OFF"` became `gui="ON"` — check that the `--headless` case sets `gui="OFF"` (it does, line 248).

Evidence the flip works end-to-end is in the comment above: three Linux headless CI jobs built with 0 Interface targets, 0 Qwt, 0 Qt lookups. mac/Windows headless were not run — they only exist in the nightly matrices.

jessdtate (approved):

PR #2761, opened 2026-09-16
ViewScene: don't stall the network when the window can't paint (#2760)

### LINKED ISSUE #2760: ViewScene that is open but not exposed (behind other windows, minimized, other Space) stalls network execution
Reported: a network stops executing when a ViewScene window is open but in the background. Raising the ViewScene (or closing it) lets the network continue.

## Mechanism

`ViewScene::execute()` on the module thread must take `screenShotMutex` (`src/Modules/Render/ViewScene.cc:381`). The dialog holds that mutex **permanently** — it locks it in its constructor (`src/Interface/Modules/Render/ViewScene.cc:458`) — and releases it for 1 ms in `unblockExecution()`, which is called from `frameFinished()`, i.e. after `GLWidget::paintGL` has completed a *requested* frame.

When new geometry arrives (`updateModifiedGeometriesAndSendScreenShot`, `ViewScene.cc:1423`):

```cpp
if (impl_->mGLWidget->isVisible() && impl_->mGLWidget->isValid())
  impl_->mGLWidget->requestFrame();   // sets a flag; only a paint can complete it
else
  unblockExecution();
```

`QWidget::isVisible()` is true for a window that is behind other windows, minimized, or on another Space. But Qt does not paint an unexposed `QOpenGLWidget` (on macOS occlusion sets `QWindow::isExposed()` false). So the frame is requested, `paintGL` never runs, `finishedFrame` never fires, `unblockExecution` is never called, and the module thread blocks on the mutex indefinitely. Raising the window exposes it → one paint → `frameFinished` → the mutex is released and execution continues. Closing the dialog destroys the lock owner with the same effect. Exactly the reported symptom.

A second variant of the same hole: a frame requested just before the window is hidden (`hideEvent` doesn't unblock).

## Fix (two layers)

1. Ask the right question: replace `isVisible()` with "will actually paint" — visible, exposed (`window()->windowHandle()->isExposed()`), not minimized. Fall through to `unblockExecution()` otherwise, and unblock from `hideEvent` when a frame is pending.
2. Backstop: the module's wait is unbounded. Make it a timed wait (a few seconds) that warns and proceeds without the screenshot. No dialog state can then stall the network, whatever the next edge case turns out to be.

Headless builds are unaffected (`BUILD_HEADLESS` path in `ViewScene::execute` never takes the mutex).

dcwhite:
Closes #2760.

## Mechanism (see the issue for the full trace)

The dialog holds `screenShotMutex` for its whole life and releases it for 1 ms after each *requested* frame paints. `ViewScene::execute()` blocks on that mutex. A frame requested while the window is visible-but-unexposed is never painted by Qt, so the release never happens and the module thread waits forever. Raising the window paints one frame and everything continues — the reported symptom.

## Fix, two layers

**Dialog** (`Interface/Modules/Render/ViewScene.cc`)
- `updateModifiedGeometriesAndSendScreenShot` now asks `glWidgetCanPaint()` — visible, valid, top-level not minimized, and `QWindow::isExposed()` — instead of bare `isVisible()`. If it can't paint, the module is released immediately (the pre-existing else branch).
- `hideEvent` releases the module if a frame is still pending, so hiding a window mid-request can't strand it either. (`GLWidget::frameRequested()` accessor added.)

**Module** (`Modules/Render/ViewScene.{h,cc}`)
- `screenShotMutex_` is now a `std::timed_mutex`; `execute()` waits at most 10 s, then warns `ViewScene window did not render within 10 s (hidden or minimized?); continuing without screenshot data.` and carries on. Whatever the next unpainted-window edge case turns out to be, it costs 10 s, not a hung network.

Headless builds are untouched (`BUILD_HEADLESS` path never takes the mutex).

## Verification

GUI regression path on macOS, all exit 0 with no timeout warnings — the normal render → release → screenshot flow is intact, including the network that consumes the screenshot ports:

```
Renderer/rgbScreenshotMatrices   3.7 s
Renderer/basicViz                1.8 s
Renderer/showColorMap            1.8 s
Modules/calcFieldData1           2.1 s
Renderer/asyncViewScene          2.3 s
```

**Not reproduced automatically:** the occluded-window case itself needs a real window manager to obscure the ViewScene mid-execute. The reporter's scenario (open ViewScene, put it behind the main window, execute) is the test — with this branch it should either just work (layer 1) or, worst case, pause 10 s and log the warning (layer 2). Would appreciate the reporter trying it.


jessdtate (approved):
seems to be hitting the time out limit (10s) when I encounter the problem.  This is noticeably better than previous, which would never change until the window is reopened (docked windows had to be closed and reopened in some tests).  The module view in the network editor still doesn't return to the a normal state, but still looks busy.  However, the execution is not blocked anymore.  Closing and reopening the offending viewscene window is still the quickest way to resolve the hanging.  

One thing I noticed during testing is that the original bug seems to be related to changes in the VS state during execution. For example, if I open and close vs windows while the network is executing, I see the bug more often.  This particular test will probably be easier with larger networks.

PR #2764, opened 2026-09-18
ViewScene: widen the widget hit-test window so handles are easier to grab

dcwhite:
## Summary
- `SRInterface::select` read a single pixel from the widget-selection FBO, so a click had to land exactly on a thin widget handle to grab it.
- Now reads a 13×13 window (`widgetSelectRadius = 6` logical px, clamped to the viewport) and picks the widget hit closest to the cursor.
- The drag origin stays at the actual click position; only the depth used for the initial-W calculation comes from the hit pixel, so the drag math is unchanged.

## Test plan
- [x] Compiles clean on macOS (arm64, Release flags)
- [x] Manually tested on `regression/Renderer/pointWidgets.srn5`: clicking near (not on) sphere/bounding-box handles now selects and drags them


jessdtate:
This does make it easier to grab the widgets, but I don't think it solves the underlying problem of the point widgets.  I think there is still a misalignment of the selction mask and the objects, and it seems to be worse when the objects are farther away in the visualization (z-buffer maybe?).  I see if I can make a video or some stills to demonstrate this.

